PR #1, opened 2022-06-05
Open simh master

markpizz:
This pull request contains the commits made by others (Trevor Warwick @trevorwarwick in particular) since the addition of the LICENSE.txt in the simh/simh master branch


pkoning2:
Thanks Mark.  It looks like https://github.com/simh/simh/commit/ac73e7b4c7bd17ab6c501b81bf6dc7aeea309520 is also needed for all this to work.  Is that right?

markpizz:
The commits could have been digested with no ill effects to any build, but only the PDP11 (and PDP10) would have included @trevorwarwick modified DUP code since only those simulators included the DUP device.  I have explicitly released the changes in the commit you reference from any of the previously discussed LICENSE.txt constraints on my future commits.

PR #8, opened 2022-06-14
QBus VAX and PDP11: Add support for DPV11 in DDCMP mode

trevorwarwick:
This commit extends the existing DUP11 support to add simulation for a DPV11 in DDCMP mode.

It has been tested with DECnet-VAX Phase V, which has the only host driver I know of. Note
that you may see a few CRC errors logged with high traffic levels, these are due to a couple of
bugs in the VMS driver, for which I'm looking at developing a patch.

Regression testing has been done on the DUP11 on VMS, RSX and TOPS20, which all seemed to work
as well as previously.


markpizz:
In your testing, did you test DPV11 on the PDP11?  Possibly Phase IV on PDP11 talking to a Phase IV DMC on a Unibus VAX?

trevorwarwick:
No, I don't know which PDP11 OS even supports the DPV in DDCMP mode, let alone have a boot disk I can use to try it out. If someone has such a thing + instructions, I can try it.

Issue #9, opened 2022-06-14
Incorrect disk size reporting (and patch)


al20878:
On platforms that have `t_addr` and `t_offset` of different bitness (e.g. Cygwin), disk capacity for large disks may be reported buggy due to the integer overflow:
```
 $ ./pdp11

PDP-11 simulator V4.0-0 Current        git commit id: 94a10aba
sim>  set cpu 11/70 4m
Disabling XQ
sim> set rq0 rauser=9000
sim> show rq0
RQ0     414MB, not attached, write enabled
        RA8U, UNIT=0, autosize
        AUTO detect format
```
Note the 414MB for a 9G disk above.

The code responsible for the output resides in `scp.c`:  `sprint_capac()`.  It uses `t_addr` in computations, which can be a 32-bit integer (whereas `t_offset` is 64-bit at the same time).  Note that `sim_disk.c` calculates disk capacity (with approximately the same flow control) using `t_offset`.  The following patch (attached) fixes the issue:

```
From 59fb2f19a662ccdf45150c4f7d94661ec5f4b432 Mon Sep 17 00:00:00 2001
From: Tony Lawrence <al20878@gmail.com>
Date: Tue, 14 Jun 2022 18:15:40 -0400
Subject: [PATCH] Use 64-bit values to avoid overflow in disk capacity output

---
 scp.c | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/scp.c b/scp.c
index 4a747cb8..2a3f4ce1 100644
--- a/scp.c
+++ b/scp.c
@@ -6473,10 +6473,10 @@ return SCPE_OK;
 const char *sprint_capac (DEVICE *dptr, UNIT *uptr)
 {
 static char capac_buf[MAX_WIDTH + 12];
-t_addr kval = (uptr->flags & UNIT_BINK)? 1024: 1000;
-t_addr mval;
+t_offset kval = (uptr->flags & UNIT_BINK)? 1024: 1000;
+t_offset mval;
 double remfrac;
-t_addr psize = uptr->capac;
+t_offset psize = (t_offset) uptr->capac;
 const char *scale, *width;

 if (sim_switches & SWMASK ('B'))
--
2.33.0
```
After it, the disk capacity gets reported correctly:
```
$ ./pdp11

PDP-11 simulator V4.0-0 Current        git commit id: 59fb2f19
sim> set cpu 11/70 4m
Disabling XQ
sim> set rq0 rauser=9000
[...]

PR #10, opened 2022-06-15
SCP: Fix incorrect disk size reporting when t_addr is 32 bits

markpizz:
As reported and fixed by @al20878 in #9


pkoning2:
Thanks Mark and Anthony!

PR #14, opened 2022-06-16
Supnik current updates from v3.12

markpizz:
These commits are part Bob's of simh v3.12.  Authored by Bob, pulled apart from his zip file into separate commits that make sense.  The rest of v3,12 to follow.


markpizz:
All of Bob's simh v3.12-2 code is now here.  He probably didn't intend the version to be 3.12-2, but that's what's defined in sim_ref.h, so that is all of his code.

PR #15, opened 2022-06-16
V4 Merge of some of Bob's simh v3.12

markpizz:
These commits to the master branch reflect the v3.12 commits moved into the Supnik-Current branch applied to the v4 codebase.


markpizz:
Another v3.12 update authored by Bob is part of this.  More to come.  These commits can be merged now with no negative consequences.

PR #18, opened 2022-06-21
Regular updates to PDP10 simulators.

rcornwell:
Regular updates to PDP10 simulators. These can be squashed if desired.


PDP10/kx10_imp.c
```diff
@@ -900,7 +901,6 @@ t_stat imp_devio(uint32 dev, uint64 *data)
              check_interrupts(uptr);
              break;
         case TYPE_BBN:
-             break;
```
pkoning2, line 903: You might put a `/* fall through */` comment here (and similarly below) to show that falling through to the next case is intentional.

rcornwell: I am not doing a fall through. I am removing a break so that the two cases are handled together. Two cases's following each other do not need a fall through:

             check_interrupts(uptr);  
             break;  
        case TYPE_BBN:  
        case TYPE_WAITS:  
             if (*data & IMP_ODPIEN) {

PR #21, opened 2022-06-24
SCP: Updated REG macros and manipulation APIs - J. David Bryan v3.12 merge

markpizz:
- Rewrote get_rval, put_rval to support fields within arrays of structures
- REG "size" field now determines access size
- REG "maxval" field now determines maximum allowed value

Merge from v3.12

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -24,6 +24,9 @@
    in this Software without prior written authorization from Robert M Supnik.
 
    21-Oct-21    RMS     Fixed bug in byte deposits if aincr > 1
+   16-Feb-21    JDB     Rewrote get_rval, put_rval to support arrays of structures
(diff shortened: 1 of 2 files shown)


markpizz:
Note: This change completes the simh v3.12 merge that it makes any sense to to.

What has not been merged is the changes to sim_tape that J. David Bryan implemented specifically for the various HP simulators he wrote/maintains and exist in current form only in his separate simh fork.  Since those changes were only useful for a simulator not available here AND the current v4 sim_tape.c is about 5100 lines of code and the v3 sim_tape.c is about 2300 lines of code.  The merge would be complicated and have no use to any existing simulators.  If someone wants to take that on, have fun.

Issue #7, opened 2022-06-14
Modernize and simplify the building system


XVilka:
There were multiple suggestions to switch to the higher-level buildsystem for SIMH to simplify the readability, portability, and structure of the project. Two main options were offered - CMake and Meson.

# CMake

@bscottm ported the SIMH to the CMake here: https://github.com/simh/simh/pull/775

> Add CMake build support as a parallel alternative to the makefile-based build system. This is not intended to be a replacement for the makefile-based build system (or the collection of Visual Studio solutions, for that matter.)
> 
> CMake is a meta-build system: It generates the build environment for a multiplicity of build systems (Ninja, VS 2019, VS 2015, MinGW Makefiles, to name a few). Documentation for how to use is in README-cmake.md.

# Meson

To simplify compilation and cross-platform support, and inspired by various C projects that switched from Autotools to Meson, including QEMU, I did an initial work of migrating from the Makefile to Meson.

There are two implementations of this language:
- [Meson](https://mesonbuild.com)
- [Muon](https://sr.ht/~lattis/muon/) - mostly work-in-progress

For now, I did the simplest compilation without running tests and optional features. Please let me know if there is an interest in this, before I proceed.

I split into two commits - adding the buildsystem and removing the unnecessary files.
The second commit is not really necessary, since these can coexist without any conflict, making the gradual transition possible, like X.org, Wayland, Mesa, QEMU did, slowly deprecating autotools on the course of multiple releases.

My PR was https://github.com/simh/simh/pull/1099 - less complete than the CMake one, I think.

LegalizeAdulthood:
CMake is what the entire C/C++ community has been moving towards for years.  CMake is mature, battle tested and has a huge community of users that can help you if you need it.

As a long-time C/C++ developer I have to say that I've never even heard of Meson/Muon.  Lately (last ~5 yrs or so) a bunch of new build systems have been popping up in the C/C++ community and none of them are sufficiently better than CMake to warrant the community's interest.  Looking up Meson, wikipedia says it's been around for 9 years.  That I've never heard of it (and I'm a bit of a C/C++ geek) implies it doesn't have any significant mindshare in the C/C++ community.

eli-schwartz:
> none of them are sufficiently better than CMake to warrant the community's interest. Looking up Meson, wikipedia says it's been around for 9 years. That I've never heard of it (and I'm a bit of a C/C++ geek) implies it doesn't have any significant mindshare in the C/C++ community.

[...]

PR #33, opened 2022-07-08
Basic CMake build for SIMH

LegalizeAdulthood:
This includes the work done in the repository retroprom/simh for a CMake based build on Linux and my additions for a CMake based build on Windows.

I'm opening this pull request on open-simh/simh in order to provide a basis for discussion around issue #7

I had initially made a pull request on retroprom/simh but it has sat there for a week with no comment or activity, so I've closed that pull request (which inadvertently attributed the commit to the wrong email address) and I'm resubmitting directly to open-simh/simh.

I haven't audited the original makefile(s) and Visual Studio projects to compare all the compile options, etc.  I would not propose deleting the original build scripts until everyone is satisfied that all the features and flags of the original scripts are supported in the CMake version.


pkoning2:
Any documentation on how to use this?  Known limitations?

LegalizeAdulthood:
Would you like me to add some sort of a `Building.md` file?  I didn't see an existing file where build instructions were located.

LegalizeAdulthood:
Regarding known limitations:
- This is a "work in progress"
- I have tested basic build scenarios on Linux and Windows
- I have not run-time tested the simulators; is there any sort of automated run-time testing of the simulators?
- Compile flags haven't been audited to verify the same options are used as the makefile/VS projects
- I picked up where retroprom left off; I haven't audited their CMake work for correctness

LegalizeAdulthood:
Additional limitations:
- The following simulators are not being built yet:
  - 3b2
  - 3b2-600
  - Various vax flavors are just copies of the existing vax simulator, but the CMake build isn't making the copies only the unique builds
  - ibmpc
  - ibmpcxt
  - scelbi
- The BootROMS code is not being built yet

There appears to be CMakeLists.txt for the missing simulator builds, but they were commented out in retroprom's repo for some reason.

For simulators that need libraries beyond pthreads (e.g. PCRE and SDL2), I haven't tried that out yet on Windows and more work is needed to enable those features.

LegalizeAdulthood:
> Regarding known limitations:
> 
> * I have not run-time tested the simulators; is there any sort of automated run-time testing of the simulators?

I see that some simulators have tests that can be invoked by the makefile.  I'll look at adding [CTest](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html) support for those

LegalizeAdulthood:
Missing SDL support:
- [ ] besm6 SDL TTF font support
- [ ] NOVIDEO option to disable SDL even if found

LegalizeAdulthood:
Missing network support:
- [ ] Locate pcap library
- [ ] pcap vcpkg dependency
- [ ] vdeplug library
- [ ] tap library
- [ ] slirp library

LegalizeAdulthood:
I've audited the targets in the `makefile` and ensured that all the simulators listed in `all` are built by default in the CMake build with the same target names.

Instead of relying on associated CMake variables and a call to `add_simulator` with only the target name, I've moved everything into arguments to `add_simulator` using CMake's built-in argument handling ability.  There is a big block comment explaining the use of `add_simulator` in the top-level `CMakeLists.txt`

I've dropped `sim` and `sim_scsi` as static libraries.  (There are still static libraries where it makes sense for code shared between simulators, e.g. nova and eclipse.)  I didn't realize at first that there are macro-dependent definitions used throughout the `sim_*.[hc]` files, e.g. the types for `t_value` and `t_addr`.

All CMake configuration options now begin with `SIMH` so that they can be easily found in the interactive CMake configuration programs `cmake-gui` and `ccmake`.  I will add documentation for these variables and how they function once I finish off the remaining missing build functionality.

Current To Do list:
- [ ] Proper `DISPLAY` handling for simulators with displays, e.g. pdp1, pdp7 type 340, etc.
- [ ] Proper `NETWORK` handling for simulators with network interfaces
- [ ] Squish all the development commits from Ingo Albrecht into a single commit as the intermediate commit messages don't yield much insight and represent incremental development more than they do bug fixes or features.

LegalizeAdulthood:
The windows-build repository doesn't seem to work correctly for PCRE.  I get missing symbols like `pcre_create`.  It works with the pcre from vcpkg, although that is a DLL version and not a static library version.  (vcpkg adds rules to copy the DLLs into your build directory, so debugging and executing the simulators from there works as expected.)

LegalizeAdulthood:
Still needed:
- [ ] pcre dependency
- [ ] SDL dependency

markpizz:
The original Visual Studio Projects dynamically assessed which Visual Studio version was being used during a build and if needed, dynamically changed the libraries referenced in the link commands to the appropriate compatible libraries.  You may need to do the same (or use the script that does this already Pre-Build-Event.cmd).

LegalizeAdulthood:
Interesting, I did an experiment on 3B2 after upgrading the solution/projects to VS2022 and it linked without error, so I'll compare and see what it is doing.

markpizz:
"Upgrading" would happen on opening the solution.  
After that, building any simulator would attempt to setup the windows-build library and include directories.

I say "attempt" since Microsoft has been known to change to object format and/or internal C libraries during the lifetime of a particular Visual Studio version when fixes are released via updates without having a new formal VS version.  The windows-build stuff has stable functionality for older VS versions, but the current VS version has historically been a moving target.

I've recommended folks to install an older VS version to get reliable builds.

The simulators that give the most library coverage are the PDP11, the VAX780, and the BESM6.  When checking library functionality, you need to build both the Debug and Release builds.

markpizz:
Indeed, the Microsoft Visual Studio internal version has changed since back in January when the initial VS2022 support was added.

I've updated the windows-build repo with rebuilt libraries using the latest VS2022 update.

markpizz:
> The windows-build repository doesn't seem to work correctly for PCRE. I get missing symbols like pcre_create. It works with the pcre from vcpkg, although that is a DLL version and not a static library version. (vcpkg adds rules to copy the DLLs into your build directory, so debugging and executing the simulators from there works as expected.)

The existing Visual Studio Projects all build with HAVE_PCRE_H and PCRE_STATIC defined.

XVilka:
I recommend also adding "Closes #7" in the PR description so that the corresponding issue would be closed automatically when this is merged. Amazing job; looking forward to it!

LegalizeAdulthood:
> I recommend also adding "Closes #7" in the PR description

Yes, I'm familiar with github workflow, but this pull request doesn't satisfy all the build requirements yet, it's a work-in-progress.

markpizz:
For the uninitiated to completely review how this PR should work, please include a complete description of how someone would engage this build on the major platforms (Linux, MacOS and Windows).

Meanwhile, looking through the CMakeLists.txt in the root directory, I see several things which don't map back to what's happening in the makefile now:
- The simh v4 codebase uses the SIM_HAVE_DLOPEN and not HAVE_DLOPEN definition
- The simh v4 codebase DOES NOT care for HAVE_PCREPOSIX_H and also DOES NOT want PCREPOSIX_LIBRARIES, it also does not want HAVE_REGEX_H
- ONLY the frontpanel test program cares about HAVE_NCURSES and that program is NOT a simulator so it doesn't care about any of the other simulator build definitions.
- Support for SDL prior to SDL2 was removed long ago, so the logic detecting things relating to SDL1 is not valid.
- Pthreads seems to be at best weakly referenced
- Like you've already admitted, the various network capabilities aren't reasonably detected or configured during build.
- There is nothing which gathers and passes in the git commit-id and time into the simulator being built.  Knowing precisely what users who report issues is running, is absolutely valuable.

You should build a simulator with the makefile and compare the output of sim> SHOW VERSION to what you get with the same simulator built with cmake differing only in "Build Tool:".

Some of the above issues may be related to when the original poking around with cmake was done in what you started from retroprom/simh which was done long ago.

Apart from these things, as I've said previously, building with Cmake isn't going to remove the need for the makefile since cmake isn't available on all currently supported platforms.  When someone once came along and previously proposed cmake, I made a good effort to provide a way to avoid the concept of maintaining yet another set of build components, defines and libraries for each simulator.  I marked up the makefile (it is still marked up) to give hints to a program I wrote to dynamically extract the components for each build target from the makefile during the initial phase of the cmake build.
Another goal was to avoid scattering 50+ CMakeLists.txt files throughout the source code tree. All CMake build directions living in a single CMake directory,

Whatever I knew previously to build with CMake was forgotten long ago when the that project never got merged.  So, what I said in my initial sentence above, is still useful for me and everyone going forward who might want to use this.

markpizz:
> > Regarding known limitations:

> > 
> > o I have not run-time tested the simulators; is there any sort of automated run-time testing of the simulators?
>
>  I see that some simulators have tests that can be invoked by the makefile. I'll look at adding [CTest](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html) support for those

The optional simulator specific extended tests are reasonable to possibly be invoked by CTest, but the current build procedures specifically invoke each simulator immediately after a successful compile with the single argument "RegisterSanityCheck".  

This invocation causes validation code within the simulator to perform checks which are logically equivalent to compile time checks but are beyond the actual compile time language processing.  These checks validate that critical data structures within the simulator code are all reasonably described to avoid inappropriate memory access or overwrite during simulator execution.   The failure of these RegisteSanityChecks is equivalent to a failed compile.

LegalizeAdulthood:
You keep adding comments, but you're not telling me anything I don't already know, so it's just noise.

markpizz:
I will continue to comment on things that your existing comments suggest you might not realize, like the comment I made today.  

You've taken on this project with some goal of doing everything better than is already being done.  Before you do it better you've got to do it as well as it is already being done.

Get things to a testable state that produces some working binaries and describe precisely how to actually test it and I'll give useful feedback or congratulations about how fantastic the result is.

I won't be commenting on anything else until I see other comments from you or others that raise questions or until you push more current state to your personal repo/branch.  The most recent activity there is some 17 days back.

pkoning2:
Richard, should we close this in light of the work in PR #53, or leave it open?  If open, how do the two relate?

rcornwell:
Not sure, this is a good place to discuss non-pr specific issues.

markpizz:
The only reason I brought up the issue(s) is to discuss out of the problem scope (cmake building) issues that are in his current changes set.

XVilka:
Shouldn't this be closed in favor of https://github.com/open-simh/simh/pull/53?

XVilka:
@pkoning2 @LegalizeAdulthood since https://github.com/open-simh/simh/pull/53 is complete, I believe there is no point of keeping this open.

XVilka:
Could be closed now

PR #45, opened 2022-08-03
QBus VAX and PDP11: Add support for HDLC framing to DUP11/DPV11

trevorwarwick:
This commit extends the existing DUP11/DPV11 support to add support for HDLC framing. There
is no requirement to support bit/byte stuffing, so the aim is simply to encapsulate/decapsulate
the datalink frame provided by the host driver.

DDCMP has been regression tested on VMS, RSX and TOPS20, and the HDLC mode was developed
and tested with VAX/VMS DECnet Phase V, only using the HDLC datalink implementation.

In theory this support is sufficient to allow Phase IV/V P.S.I and even VMS/SNA to be used,
but these remain to be demonstrated.

### DIFF
--- a/PDP11/pdp11_dup.c
+++ b/PDP11/pdp11_dup.c
@@ -125,6 +125,7 @@ static t_stat dup_set_W5 (UNIT* uptr, int32 val, CONST char* cptr, void* desc);
 static t_stat dup_show_W5 (FILE* st, UNIT* uptr, int32 val, CONST void* desc);
 static t_stat dup_set_W6 (UNIT* uptr, int32 val, CONST char* cptr, void* desc);
 static t_stat dup_show_W6 (FILE* st, UNIT* uptr, int32 val, CONST void* desc);
+static uint16 dup_crc_ccitt (uint8 const *bytes, uint32 len);
 static t_stat dup_help (FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const char *cptr);
 static t_stat dup_help_attach (FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const char *cptr);
 static const char *dup_description (DEVICE *dptr);
@@ -414,7 +415,7 @@ static BITFIELD dpv_parcsr_bits[] = {
 #define PARCSR_V_DPV_ADSYNC   0
 #define PARCSR_S_DPV_ADSYNC   8
 #define PARCSR_M_DPV_ADSYNC   (((1<<PARCSR_S_DPV_ADSYNC)-1)<<PARCSR_V_DPV_ADSYNC)
-    BITF(DPV_ERRDET,8),                         /* CRC Type */
+    BITF(DPV_ERRDET,3),                         /* CRC Type */
 #define PARCSR_V_DPV_ERRDET   8
 #define PARCSR_S_DPV_ERRDET   3
 #define PARCSR_M_DPV_ERRDET   (((1<<PARCSR_S_DPV_ERRDET)-1)<<PARCSR_V_DPV_ERRDET)
@@ -935,7 +936,7 @@ switch ((PA >> 1) & 03) {                               /* case on PA<2:1> */
             if (dup_txcsr[dup] & TXCSR_M_DPV_SEND) {
                 dup_txcsr[dup] |= TXCSR_M_DPV_TXACT;
                 sim_activate (dup_units+dup, dup_wait[dup]);
-                /* Go ahead, Abort not supported */
+                /* Go ahead not supported */
             }
         }
         break;
@@ -1205,9 +1206,12 @@ return SCPE_OK;
 t_bool dup_put_msg_bytes (int32 dup, uint8 *bytes, size_t len, t_bool start, t_bool end)
 {
 t_bool breturn = FALSE;
+int32 charmode;
 
 if ((dup < 0) || (dup >= dup_desc.lines) || (DUPDPTR->flags & DEV_DIS))
     return FALSE;
+
+charmode = ((UNIBUS) ? (dup_parcsr[dup] & PARCSR_M_DECMODE) : (dup_parcsr[dup] & PARCSR_M_DPV_PROTSEL));
 
 if (!tmxr_tpbusyln(&dup_ldsc[dup])) {  /* Not Busy sending? */
     if (start) {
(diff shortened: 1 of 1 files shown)


markpizz:
One rule I always required when folks submit changes to existing modules is that they must consistently follow the indentation style already in use in the module being changed.  If they're creating their own source module, they get to use their personally preferred style.

The pdp11_dup.c in this zip:
[pdp11_dup.zip](https://github.com/open-simh/simh/files/9254374/pdp11_dup.zip)
is functionally unmodified, but follows the existing indentation style previously used in the module.  One change removed an unused variable that produces a compiler warning.

pkoning2:
I still haven't figured out how to describe the indentation style that (most of) simh uses.  My attempts at creating an emacs style definition have been rough approximations at best.    So given that I'm somewhat hesitant to be picky about this.  It's hard to follow a non-standard.

pkoning2:
Meanwhile, there seems to be a problem with the checker; it doesn't seem to want to do the Mac builds even after several attempts.

markpizz:
In general:
- open braces go on the line which created their need (if, else, do, while, etc.).
- Code within the brace section is indented 4 relative to the line which created their need .
- The matching closing brace is indented the same as the code section it is closing.
- else is at the same indent level as the "if" it relates to.

Exceptions to "In general":
- Opening braces for functions is in column 1.  
- Closing brace for that opening function brace is also in column 1.

A blank line between local variable declarations and the code that references them.

Usually complex condition expressions usually avoid depending on the formal operator precedence and are best written with additional parenthesis to clarify the intended effect.

No tab characters in source files since folks have different tab stop settings.
For legacy reasons, source/text file line endings are CRLF.

markpizz:
> Meanwhile, there seems to be a problem with the checker; it doesn't seem to want to do the Mac builds even after several attempts.

Like on my earlier case, this also is saying that the mac builds are being canceled.  Maybe he and I did something to cause the cancelation.  I don't recall doing anything, but who knows.

After Trevor digests the changed module in the zip, if you pull in hist forced push, locally, build the PDP11 simulator and it builds cleanly, you can just ignore the "canceled" checks.

pkoning2:
Yes, they build cleanly on my Mac.
As for indent rules, there is the problem that the first level code inside a function isn't indented at all, at least in part of the files.  That is just too strange for editors to cope with.  I also don't like it at all.

friesga:
The rules you follow for submissions to simh v4 and the indentation style used in that project do not necessarily have to be the rules and style for OpenSimh. As you show in your next reply this indentation style is inconsistent and it is no wonder then that Paul couldn’t create an emacs style definition. I once tried to define a clang-format style definition and failed too.

 

As  1. the current code style makes the code hard to read and 2. the code style isn’t consistent, I would advise the steering group to adapt a consistent definition and record that style in a definition file of a tool such as clang-format. Any decent format will do as from that moment on everyone can use his favourite style. Just when files are committed, the defined format is applied to the file so only actual changes are included.

 

Van: Mark Pizzolato ***@***.***> 
Verzonden: woensdag 3 augustus 2022 21:25
Aan: open-simh/simh ***@***.***>
CC: Subscribed ***@***.***>
Onderwerp: Re: [open-simh/simh] QBus VAX and PDP11: Add support for HDLC framing to DUP11/DPV11 (PR #45)

 

One rule I always required when folks submit changes to existing modules is that they must consistently follow the indentation style already in use in the module being changed. If they're creating their own source module, they get to use their personally preferred style.

The pdp11_dup.c in this zip:
pdp11_dup.zip <https://github.com/open-simh/simh/files/9254374/pdp11_dup.zip> 
is functionally unmodified, but follows the existing indentation style previously used in the module. One change removed an unused variable that produces a compiler warning.

ethandicks:
On Wed, Aug 3, 2022 at 5:43 AM Trevor Warwick ***@***.***> wrote:
> This commit extends the existing DUP11/DPV11 support to add support for HDLC framing. There
> is no requirement to support bit/byte stuffing, so the aim is simply to encapsulate/decapsulate
> the datalink frame provided by the host driver.
>
> DDCMP has been regression tested on VMS, RSX and TOPS20, and the HDLC mode was developed
> and tested with VAX/VMS DECnet Phase V, only using the HDLC datalink implementation.
>
> In theory this support is sufficient to allow Phase IV/V P.S.I and even VMS/SNA to be used,
> but these remain to be demonstrated.

Very cool.  I used to make DEC-to-IBM products (COMBOARD/HASP,
COMBOARD/3780 and COMBOARD/SNA) and our earlier product, the HASPBOX
was a PDP-11 (/04 or /03) with a DUP11/DPV11 and an earlier version of
the COMBOARD/HASP front-end code.  It's nice to see some sync comms
make it into simh.

At some point in the next couple of years I hope to get my machines to
the point where I can run COMBOARDs again so I could (eventually) do
some box-to-box testing between a 1980s product and simh.  In the
meantime, Any HASP and 3870 implementation can be connected to another
(with one end in Central Mode and the other in Remote Mode) and if
it's a good implementation, it should "just work".  SNA requires a lot
of smarts in the middle typically.  COMBOARD/SNA is a PU Type 2
implementation only (printers and interactive terminals).  We never
got an LU2.1 product rolled out.

I'll be interested to see any attempts to set up 2780/3780 products on
simh (there are a couple candidates)

pkoning2:
PSI exists on RSX as well, doesn't it?
As for 2780, I thought that's BISYNC.

trevorwarwick:
Yes, there is an RSX P.S.I.

IIRC 2780/3780 is Bisync.  The VMS driver supports Bisync as well, and so
it would be easy enough to add it to the simh dup/dpv code in future. At
the moment it assumes that character mode will be DDCMP only.

markpizz:
I had the code conventions so that folks looking at the code in whatever module would be able to read, regardless of their personal formatting style, the code and see something consistent instead of having to mentally switch gears as they moved from one place to another in the module.

Like I said, when you write a new module you get to pick the style which you should then use consistently.

Maybe the steering committee should allow a modified convention whereby anyone changing any module in the code should reformat that whole module to their preferred style.  The party making the change could then be completely comfortable with what they are looking at as they make their change.  This would simplify things for the person making the change and then make the merge process massive and the reviewer's work possibly overwhelming when looking at actual differences.

@trevorwarwick, when making any changes during a PR review that polish the details of the PR, it is best to not pass along a series of commits that achieved this polish, but to back out the earlier commits and then commit the updated file(s) and then force push your working branch again.  The PR will nicely be updated with the current state.

In your latest commit you've chosen to pick and choose which of the formatting changes I provided.  Your thinking?

pkoning2:
Mark, I suggested a second commit for those trivial tweaks.  The two were my suggestion, they were (from reading the diffs in the PR) the only two that stood out as inconsistent.

trevorwarwick:
If you follow the link with the failure message it says it's a scheduled
24hr brownout, to encourage people to migrate to a later/different version.

It should work tomorrow.

Issue #29, opened 2022-07-06
Strange behavior of the VH device on PDP-11


louis-chretien:
While setting up a config file for my PDP-11/70 and RSTS/E system, i noticed something strange:

The VH device (DHQ-11) seems to have a weird behavior.

Since my system is a Unibus machine (pdp-11/70), it's supposed to be a DHU-11 with 16 lines per unit. If i set it up this way:

set VH enable
set VH lines=32
attach VH 6000

the "show config" returns:

VH      address=17760440-17760477*, vector=500-514*, BR4, lines=32, 2 units
  VH0         attached to 6000, DHU mode, 0 current connections
  VH1         DHU mode

which is most correct.

But if i try to set it up as 3 devices (like in my system):

set VH enable
set VH lines=48
attach VH 6000

the "show config" returns:

VH      address=17760440-17760517*, vector=500-524*, BR4, lines=48, 3 units
  VH0         attached to 6000, DHU mode, 0 current connections
  VH1         DHU mode
  VH-RCV-CON  DHU mode

When i looked at the pdp11_vh.c code, i find a "VH-RCV-CON" reference in the initialization code, as a name when a unit is reset.

It seems that the DHQ code has trouble handling more than 32 lines.

Or am i overlooking something?

pkoning2:
The code says that the max controller count is 4 and it looks like the line count is figured from that and 8 lines per controller.  In a comment it says the line count can be 8, 16, 24, or 32.
I don't know why it should be limited like that, though.  But it looks like it needs code changes, and in any case the error handling should be strengthened.

markpizz:
The only problem is the display of the unusual unit name for the 3rd controller in this case.  The number of working lines and is correct and will behave as normal.  

[...]

Issue #28, opened 2022-07-04
Tags for releases in branch Supnik-Current


wfjm:
The branch `Supnik-Current` used to track Bob's releases of SimH-Classic.  
And up to 2020 Bob's releases were properly tagged, last tag is [v3.11-1](https://github.com/open-simh/simh/releases/tag/v3.11-1).  

With pull requests [14](https://github.com/open-simh/simh/pull/14) and [23](https://github.com/open-simh/simh/pull/23) the releases `v3.12` and `v3.12-2` were somehow added to branch `Supnik-Current`.  
The pull 14 is labeled _These commits are part Bob's of simh v3.12_.
The pull 23 in fact has two commits, the first apparently reflects `v3.12-2`, the second an additional change.

What is the purpose of the `Supnik-Current` branch in `open-simh` ?  
Pull's are including _are part Bob's ..._, or mix Bob's releases with other changes.
There are no tags for v3.12 and v3.12-2.  
And maybe that's not even possible now.

I'd highly appreciate if Bob's releases are properly reflected in the `Supnik-Current` branch.  
As it used to be in `simh/simh`, where all Simh_Classic releases from `v2.3a` to `v3.11-1` are easily accessible.

In case of SimH-Classic we have proper releases, and that should be imho visible as tags in open-simh.

pkoning2:
As far as I know that branch is supposed to track Bob's work.  If it doesn't that should be fixed.  Mark Pizzolato has contributed the commits for this; I don't know the details (other than that things seem to build and run).

wfjm:
@pkoning2 : I checked the `Supnik-Current` branch. 
I checked out commit 14dccaa which is labeled `simh v3.12-2`.
And downloaded `simhv312-2.zip` from http://simh.trailing-edge.com.  
And diff'ed the two source trees and got
```
  Only in os-simh/simh: .git
  Only in os-simh/simh: .gitattributes
  Only in os-simh/simh: .gitignore
  Only in os-simh/simh: 0readme_311.txt
  Only in te-simh/sim/VAX: vmb.exe
  Only in os-simh/simh: Visual Studio Projects
```
Both file trees are identical at the source files level.
A bit astonishing is that `vmb.exe` is in Bob's zip fille, but but in the open-simh branch.  
The open-simh branch contains `Visual Studio Projects` which is not in Bob's zip file.

So the `Supnik-Current` branch somehow tracks Bob's work.

What is missing are proper tags, as we had them before.  
[...]

PR #47, opened 2022-08-08
CMake branch

bscottm:
Definitely works with MSVC and Linux-GCC. MinGW and Clang on Windows are a bit dodgy at the moment for a variety of reasons that need to be worked out with the upstream maintainers.

- Documentation in `README-CMake.md`; might be slightly out of date, but pretty workable.
- Simulator `CMakeLists.txt` files are automagically generated via the `cmake/generator.py` script. If you change `Makefile`, you will trip the regeneration process, if you have `python3` installed.

Quickstart:
````
$ cd cmake
$ ./cmake-builder.sh
````

For PowerShell:
````
PS> cd cmake
PS> .\cmake-builder.ps1
````

### CODE REVIEW

**Comment by bscottm** [comment]
Address a few of @markpizz's comments (which should be in this merge request):
  - PCRE is now the default RE library, not PCRE2. The CMake default was PCRE2, but that requires patches that aren't in the repo. Yet.
  - PCAP: Don't build 'rpcapd'. YAGNI. Unbreaks the Windows VS2022 build.
  - sel32 simulator: `makefile` used "$()" on variables, which confused the `generator.py` script, which expects variables to be surrounded by "${}".

CMake with MinGW note: Working with CMake's developers to figure out why they thought it would be a great idea to put the cross-compile chain prefix on `windres`, when Cygwin and MinGW don't even package `windres` that way.

**Comment by bscottm** [comment]
Also, remove the "all-in-one" CMake simulator source (commented the code out, to be removed at a later time.)

**Comment by bscottm** [comment]
1. HAVE_DLOPEN=xxx should be changed to SIM_HAVE_DLOPEN=xxx.
2. simh only depends on pcre.h and it seems that although this file exists on the system, the compiles are not then invoked with HAVE_PCRE_H defined.
3. You aren't building all the simulators. Specifically, I notice that for some reason you're not building the SEL32 simulator. Maybe you should build all directories found with an exclusion list of directories to avoid (i.e. display, .github, .travis, BIN, doc, .etc)
4. On Ubuntu Linux, I happened to not have installed the OS provided libsdl2-ttf-dev package installed, your procedure tried to fetch and build sdl2-ttf, but that invocation of cmake_builder failed. I installed the OS provided package and started over and that worked. Your build of the library failed. You should really prefer the OS provided packages since, when they are available, they are validated by the OS packager.
5. Your information in README-cmake.md about WinPcap and Npcap is at least out of date. WinPcap is obsolete and Npcap is fully supported. Building simh with Pcap ethernet support on essentially all platforms only requires pcap.h to be found. The interface to the kernel level packet capture and injection capabilities are dynamically loaded at runtime. When pcap.h is available, compiles should be invoked with HAVE_PCAP_NETWORK, USE_SHARED and USE_READER_THREAD. In your README-cmake.md, you specifically say that flex and bison (winflexbison) are needed to compile and build the pcap library. Since this never needs to happen for perfectly fine networking support, the extra tool chain dependencies are not needed either.
5. On Linux you don't seem to detect the needed include for tun/tap support, and thus compile with HAVE_TAP_NETWORK defined.

Things might get further on Windows once the effort to build pcap is removed...

On MacOS, it initially needs gnu-getopt to be installed which isn't mentioned in your README-cmake.md. Once that's installed, the cmake-builder.sh script blows up:

````
marks-mac-mini:cmake mark$ ./cmake-builder.sh 
./cmake-builder.sh: line 173: realpath: command not found
usage: dirname string [...]
./cmake-builder.sh: Can't determine SIMH top-level source directory.
Did this really happen?
````

**Comment by bscottm** [comment]
@markpizz:

1. `SIM_HAVE_DLOPEN` should be set by `HAVE_DLOPEN`.
2. My repo had patches for PCRE2 and PCRE2 is the default in my repo. Temporarily reverted the default option in `CMakelists.txt` to grab PCRE.
3. See "$()" should be "${}" in notes above.
4. Reread your comment -- if you install the package, CMake prefers it (although it might be that I'm expect something called `SDL2_ttf` and not `SDL_ttf`.) Unclear why the build failed when `SDL2_ttf` was built as a dependency.
5.  Partially updated notes.
6. Something to look into.

**Comment by markpizz** [comment]
> 1. SIM_HAVE_DLOPEN should be set by HAVE_DLOPEN

Nope!  The compiles of the simh modules DO NOT have this properly defined.  

To test, build any simulator with the makefile, save the result and then build the same with cmake.  Run each, and type anything at the sim> prompt.  hit up-arrow on the keyboard.  The makefile build will show the first command you typed.  The cmake build will show ^[[A.  Most folks will like the makefile results better.

> 2. My repo had patches for PCRE2 and PCRE2 is the default in my repo. Temporarily reverted the default option in CMakelists.txt to grab PCRE.

Just because PCRE2 is a later "version" of PCRE doesn't mean that the code in simh can leverage any part of the functionality of the extended PCRE2.  The functionality provided by the PCRE APIs are all that is needed AND that required functionality and APIs are present when PCRE2 is installed.  Originally simh tried to use pcreposix APIs hoping that the C runtime (usually Posix) reex support could be used if pcreposix wasn't available.  This turned out not to be reliable on some modern platforms, so I migrated to direct use of the PCRE APIs,

> 4. Reread your comment -- if you install the package, CMake prefers it (although it might be that I'm expect something called SDL2_ttf and not SDL_ttf.) Unclear why the build failed when SDL2_ttf was built as a dependency.

You are right, you do prefer the installed package.  My point wasn't intended suggest that you didn't.  What I meant was that, if a package is known to have useful vendor provided package, then the cmake default should NOT be to build from source, but to either suggest the user install the vendor one, or live without it and let the user know to install the vendor one.  There are absolute answers to whether vendor support exists for each of the dependent packages for the most common target platforms.

> 5. Partially updated notes.

What is required IS NOT about updating the notes.  It is completely about NOT trying to build pcap.  The earliest simh Ethernet support used locally built libpcap in the simh build.  That was created BEFORE a sufficiently capable libpcap was present in *nix operating systems some 14 years ago.  Current operating systems have had built-in support for sufficiently capable libpcap for more or less 12 years.  Dynamically loading pcap (on all potential platforms) is the right thing to do and what you'll get when compiling with HAVE_PCAP_NETWORK, USE_SHARED and USE_READER_THREAD.  The ONLY requirement to build when these defines exist is to have a pcap.h defined which provides the signatures for the needed pcap APIs.

Removing this libpcap build activity simplifies the tool-chain needed to build using the whole package.  That isn't a concern for you since you've already jumped through those hoops, but minimizing the needed tool-chain components for the casual walk up party has value for the desired target audience.

**Comment by LegalizeAdulthood** [comment]
@bscottm I'm busy for a couple weeks, but at first glance it looks like your effort is farther along than what I had worked on in that other pull request.  I think it might make more sense to contribute to your effort, but I won't be able to give it a serious inspection for a couple weeks.

**Comment by bscottm** [comment]
Latest updates to address @markpizz original merge comments, tested on Windows with VS 2022 and ARM64 Linux. Other platforms, YMMV.

- "HAVE_DLOPEN" -> "SIM_HAVE_DLOPEN"
- Updated docs to note `coreutils` and `util-linux` dependencies. Not a problem for most base Linux distributions, but MacOS developers will likely have to install those equivalents if missing.
- Expand and fix `pkg-config`-detected packages when CMakes `Find*` functions don't detect required headers and libraries.

N.B.: `pkg-config` for SDL TrueType support is `SDL2_ttf`, whereas SDL2 is `sdl2`, sometimes `SDL2`.

**Comment by bscottm** [comment]
@LegalizeAdulthood: This should be a bit more mature. I've been using it for 3 years. :-)

**Comment by markpizz** [comment]
> - Updated docs to note coreutils and util-linux dependencies. Not a problem for most base Linux distributions, but MacOS developers will likely have to install those equivalents if missing.

realpath is not generally available on MacOS and seems to only be available for MacPorts and not HomeBrew.  From what I've seen the most significant population of MacOS users are HomeBrew oriented.

It would seem useful to support that community of users.  Ask Paul Koning and Clem Cole.

> - TUN networking is enabled by default (PCAP, VDE, TUN and SLIRP are now all enabled.)

TUN is now available, but on my Ubuntu test with both libvde-dev and libvde0 installed, no VDE support is mentioned or included when invoking cmake-builder.sh.

----------

Since you have not acted or commented on my input about the need to compile and link against libpcap, Please explain how/why you see it is valuable or important to compile and link against libpcap?

Given the active effort to build libpcap, why isn't libvde being built with the same effort?

**Comment by bscottm** [comment]
> > 1. SIM_HAVE_DLOPEN should be set by HAVE_DLOPEN
> 
> Nope! The compiles of the simh modules DO NOT have this properly defined.

Mark, it turned out to have been a simple rename. I could easily done a substitution by setting SIM_HAVE_DLOPEN to the value of HAVE_DLOPEN when I add that particular compiler define. Easier to just rename (add "SIM_".)

> > 2. My repo had patches for PCRE2 and PCRE2 is the default in my repo. Temporarily reverted the default option in CMakelists.txt to grab PCRE.

I wrote the patch for PCRE2 around two years ago. It's a different API, so it's not a drop-in replacement. And, yes, I do remember when you implemented that patchset. The upside of PCRE2 is that it is actively maintained. PCRE is obsolete. However, I can't contribute the PCRE2 patchset easily with the `makefile`-based build system.

> You are right, you do prefer the installed package. My point wasn't intended suggest that you didn't. What I meant was that, if a package is known to have useful vendor provided package, then the cmake default should NOT be to build from source, but to either suggest the user install the vendor one, or live without it and let the user know to install the vendor one. There are absolute answers to whether vendor support exists for each of the dependent packages for the most common target platforms.

While I would agree with you in principle on *nix systems, having a reliable place where dependencies install is a hard requirement on Windows. CMake could support something that looks like `autoconf`'s `--with-` constructs, which is suboptimal from a maintainer's point of view. If it's possible to build something from source, then build it and put it into a consistent place and let the build system proceed normally.

Now, if your statement is, "On Linux and MacOS, just yell at the user to install the dependency," I respectfully disagree. The point of open source is access to code, and access to code means that it can be built.

> What is required IS NOT about updating the notes. It is completely about NOT trying to build pcap.

I haven't found a good alternative or mechanism for locating and installing a pre-built `libpcap` DLL on Windows. That's why it's built as a dependency on Windows.

**Comment by markpizz** [comment]
> > What is required IS NOT about updating the notes. It is completely about NOT trying to build pcap.

I haven't found a good alternative or mechanism for locating and installing a pre-built libpcap DLL on Windows. That's why it's built as a dependency on Windows.

There is NO NEED for you to provide a DLL.  Build with a pcap.h in the include paths and the defines I mention and the binaries will work just fine when Npcap (or Winpcap) is installed, AND they will also work just fine if Npcap (or Winpcap) is NOT installed, without pcap networking.  Similar results on *nix systems except all likely have pcap available, but most will need root access to leverage pcap.

**Comment by bscottm** [comment]
It's a bit more complicated than just `pcap/pcap.h`.

**Comment by markpizz** [comment]
It may be more than simply the pcap.h file, but it certainly DOES NOT require compiling and linking any pcap components except  sim_ether.c.

**Comment by bscottm** [comment]
@markpizz: Ready for a new round of acceptance testing. PCAP suggestion implemented.

Idea for new option:  Linux default - on, Windows default - off -- terminate if dependencies not met (Linux would terminate, Windows wouldn't.)

**Comment by XVilka** [comment]
It would also require a section in README, especially on what dependencies are required and options to enable/disable them.

**Comment by bscottm** [comment]
@XVilka: In the `README.md` or `README-CMake.md`? Here's the current list in the `CMakeLists.txt`:

````
##-- Options --##
option(WITH_ASYNC
       "Enable (=1)/disable (=0) Asynchronous I/O and threading."
       TRUE)
option(WITH_REGEX
       "Enable (=1)/disable (=0) PCRE/PCRE2 regular expression support (def: enabled)"
       TRUE)
option(PREFER_PCRE
       "Prefer (=1)/ignore (=0) Prefer PCRE over PCRE2 (def: ignore)"
       TRUE)
option(WITH_NETWORK
       "Enable (=1)/disable (=0) simulator networking support. (def: enabled)"
       TRUE)
option(WITH_PCAP
       "Enable (=1)/disable (=0) libpcap (packet capture) support. (def: enabled)"
       TRUE)
option(WITH_SLIRP
       "Enable (=1)/disable (=0) SLIRP network support. (def: enabled)"
       TRUE)
option(WITH_VDE
       "Enable (=1)/disable (=0) VDE2/VDE4 network support. (def: enabled)"
       TRUE)
option(WITH_TAP
       "Enable (=1)/disable (=0) TAP/TUN device network support. (def: enabled)"
       TRUE)
option(WITH_VIDEO
       "Enable (=1)/disable (=0) simulator display and graphics support (def: enabled)"
       TRUE)
option(PANDA_LIGHTS
       "Enable (=1)/disable (=0) KA-10/KI-11 simulator's Panda display. (def: disabled)"
       FALSE)
option(DONT_USE_ROMS
       "Enable (=1)/disable (=0) building support ROMs. (def: disabled)"
       FALSE)
option(ENABLE_CPPCHECK
       "Enable (=1)/disable (=0) 'cppcheck' static code analysis. (def: disabled.)"
       FALSE)
````

I haven't made a pass through the `makefile` to turn off options that conflict, e.g., `WITH_NETWORK` is `0`, disable all other network options.

**Comment by markpizz** [comment]
> It's a bit more complicated than just pcap/pcap.h

It looks like you worked out the details for this. Great.

The cmake-builder.ps1 and any related documentation should now remove the requirement for Bison and Flex.  Also there should be no mention of Npcap's path in cmake-builder.ps1.  Mention of Winpcap in README-cmake.md should be replaced with Npcap.  README-CMake.md should also not mention Npcap's path.

-----------
Earlier I said and you said:

> > You are right, you do prefer the installed package. My point wasn't intended suggest that you didn't. What I meant was that, if a package is known to have useful vendor provided package, then the cmake default should NOT be to build from source, but to either suggest the user install the vendor one, or live without it and let the user know to install the vendor one. There are absolute answers to whether vendor support exists for each of the dependent packages for the most common target platforms.
>
> While I would agree with you in principle on *nix systems, having a reliable place where dependencies install is a hard requirement on Windows. CMake could support something that looks like autoconf's --with- constructs, which is suboptimal from a maintainer's point of view. If it's possible to build something from source, then build it and put it into a consistent place and let the build system proceed normally.
> 
> Now, if your statement is, "On Linux and MacOS, just yell at the user to install the dependency," I respectfully disagree. The point of open source is access to code, and access to code means that it can be built.

Absolutely, in theory it can be built.  However, the default details to build something reliably useful on a particular platform MAY NOT be what you get by default when you just build it.  The folks who prepare the official packages worry about such details and deliver what is known to work on the platform they support.  You prove this point yourself with what's in your cmake/patches.  If a Linux, MacOS or other "common" system type has dependent packages which are known to work as delivered by the respective packaging system, then those should absolutely be preferred and thus you don't have to work out what :works or doesn't:.  Sometimes what "works or doesn't" is more complicated than merely that a build succeeds.  There can be subtleties that are only visible when the package is in use.  The package maintainers will handle these things, you needn't.

If a user has, for their own reasons, previously built some dependent package and the cmake system finds it, that should absolutely be used.

You said: 
> Idea for new option: Linux default - on, Windows default - off -- terminate if dependencies not met (Linux would terminate, Windows wouldn't.)

Is this more or less what I've just described above?

------------
Also, the basic RegisterSanityCheck test performed by the makefile for all simulator builds is logically the equivalent of extended compile time checks which couldn't really be done by the compiler.  To see a contrived failure case, temporarily make the following change to pdp1_clk.c:
```
  REG clk_reg[] = {
+    { STRDATA (REGSHOULDFAIL, clk_cntr, 16, 32, 24, 32, 128, 0) },
+    { DRDATA (REGSHOULDFAIL2, clk_cntr, 64), REG_HRO },
      { ORDATAD (CNTR, clk_cntr, 16, "clock counter, 0-59999(base 10)") },
      { DRDATA (SBS32LVL, clk32ms_sbs, 4), REG_HRO },
      { DRDATA (SBS1MLVL, clk1min_sbs, 4), REG_HRO },
      { NULL }
      };
```
The compile step succeeds, and the RegisterSanityCheck should fail. 
This should be done immediately after linking and not really later as part of the test stage.

This test should take less than a second for each simulator.  Somehow when run in the cmake test phase, it takes 3 minutes...

-------------
cmake-builder.sh behavior on MacOS is unchanged (doesn't work).

**Comment by bscottm** [comment]
> You said:
> 
> > Idea for new option: Linux default - on, Windows default - off -- terminate if dependencies not met (Linux would terminate, Windows wouldn't.)
> 
> Is this more or less what I've just described above?

Likely, but it shouldn't take several comments for me to intuit the course of action.

> Also, the basic RegisterSanityCheck test performed by the makefile for all simulator builds is logically the equivalent of extended compile time checks which couldn't really be done by the compiler. 

> The compile step succeeds, and the RegisterSanityCheck should fail. This should be done immediately after linking and not really later as part of the test stage.

Unfortunately, that's not how the CMake process works. `ctest` is not intended to be called during the CMake build phase. Adding  steps to the build process can be done, although adding build steps is intended to be special-purpose, e.g., regenerating the simulators due to an updated `makefile`.

> This test should take less than a second for each simulator. Somehow when run in the cmake test phase, it takes 3 minutes...

Log files, PATH/environment setup, ... `ctest` does a fair bit of work to ensure a consistent environment in which tests execute. Does take longer on Windows.

> cmake-builder.sh behavior on MacOS is unchanged (doesn't work).

Don't have access to a MacOS machine -- are there build logs you can truncate/post that give me some insight? Or is it that `realname`/`dirname`/`basename` aren't available (i.e., `coreutils`-equivalent isn't installed)?

**Comment by markpizz** [comment]
> > Also, the basic RegisterSanityCheck test performed by the makefile for all simulator builds is logically the equivalent of extended compile time checks which couldn't really be done by the compiler.
> >
> > The compile step succeeds, and the RegisterSanityCheck should fail. This should be done immediately after linking and not really later as part of the test stage.
>
> Unfortunately, that's not how the CMake process works. ctest is not intended to be called during the CMake build phase. Adding steps to the build process can be done, although adding build steps is intended to be special-purpose, e.g., regenerating the simulators due to an updated makefile.

What I really pointing out is that the RegisterSanityCheck belongs in the link step (on all simulators and platforms) and NOT as a separate ctest activity.  It is logically equivalent to more robust compiler checks but the executable is needed to perform the test so it should happen after link.

The things that are checked would usually never be noticed by normal user use of simulators, since users only rarely use EXAMINE and DEPOSIT and then only on a few common registers and essentially never on REGisters in obscure devices.  What is broken is SAVE/RESTORE which broken registers will often cause SEGFAULTs.

------------
> > This test should take less than a second for each simulator. Somehow when run in the cmake test phase, it takes 3 minutes...
>
> Log files, PATH/environment setup, ... ctest does a fair bit of work to ensure a consistent environment in which tests execute. Does take longer on Windows.

It wasn't on Windows it was Linux., and the excess delay has nothing to do with ctest setup:

```
24/70 Test #24: simh-id32 ........................   Passed    0.91 sec
      Start 25: simh-lgp
25/70 Test #25: simh-lgp .........................   Passed    0.88 sec
      Start 26: simh-eclipse
26/70 Test #26: simh-eclipse .....................   Passed    0.90 sec
      Start 27: simh-nova
27/70 Test #27: simh-nova ........................   Passed    0.96 sec
      Start 28: simh-pdp1
28/70 Test #28: simh-pdp1 ........................***Timeout 180.10 sec
 Running internal register sanity checks on PDP-1 simulator.
PDP-1 CLK:REGSHOULDFAIL used the STRDATA macro to describe a 32 bits wide and 32 elements deep array
PDP-1 CLK:REGSHOULDFAIL2 used the DRDATA macro to describe a 64 bit wide field
Simulator device register sanity check error

      Start 29: simh-pdp10
29/70 Test #29: simh-pdp10 .......................   Passed    0.89 sec
      Start 30: simh-pdp10-ka
```

When there is/was no error in this PDP1 simulator, the RegisterSanityCheck took less than a second like the others.

------------------
> > cmake-builder.sh behavior on MacOS is unchanged (doesn't work).
>
> Don't have access to a MacOS machine -- are there build logs you can truncate/post that give me some insight? Or is it that realname/dirname/basename aren't available (i.e., coreutils-equivalent isn't installed)?

Same realpath command not found.  Very early in the execution of the script.

A little google searching for realpath on MacOS tells a couple of ways a GUI user could dig that up, but nothing from a script.

No such thing as coreutils (or any *utils*) in HomeBrew.

I am not a MacOS desktop user.  I've got a Mac-Mini that I use to test and fix simh builds.

**Comment by bscottm** [comment]
@markpizz: Latest round of updates:

- NO_DEP_BUILD option in the CMakeLists.txt. Default is OFF/0 for Windows (missing dependencies will build), ON/1 for everything else (will wail about missing deps.)
- README-CMake.md updates with suggested edits.
- `cmake/cmake-builder.sh` will look for `realpath` and `dirname` and complain if missing. (Seriously, though, don't you have MacPorts or Brew?)

Why are `realpath` and `dirname` required in the `cmake/cmake-builder.sh` script? Because the script tries to look for `CMakeLists.txt` and `..` doesn't quite mean what you think it means if you arrived at a directory via a symlink.

**Comment by bscottm** [comment]
@markpizz: Will look into the register sanity check issue, although that doesn't really smell like a `CMake` issue, per se. I'm not really detecting the subtlety between running the sanity check immediately after the link vice during the test running phase conducted by `ctest`.

All I'm divining is, "You'd catch a problem earlier." If that's the point, it seems to me to be a distinction without a substantial difference. Functionally, it's not a major difference between the `makefile` and CMake.

**Comment by markpizz** [comment]
> - cmake/cmake-builder.sh will look for realpath and dirname and complain if missing. (Seriously, though, don't you have MacPorts or Brew?)

My test environment has BOTH HomeBrew and MacPorts since they are required for the makefile build.

Like I said, what you want is not provided directly.

-----------------
> @markpizz: Will look into the register sanity check issue, although that doesn't really smell like a CMake issue, per se. I'm not really detecting the subtlety between running the sanity check immediately after the link vice during the test running phase conducted by ctest.

If the ctest stuff is always run, then fine, but actively running ctest seems to be an option and not the default.

If that stays in the ctest, then the 3 minutes vs < 1 second issue should be solved.

**Comment by markpizz** [comment]
> - NO_DEP_BUILD option in the CMakeLists.txt. Default is OFF/0 for Windows (missing dependencies will build), ON/1 for everything else (will wail about missing deps.)

OK.  You could even go further on MacOS if HomeBrew is installed.  HomeBrew does not require running as root, so you could just invoke **'brew install missing-deps'**.

**Comment by bscottm** [comment]
> My test environment has BOTH HomeBrew and MacPorts since they are required for the makefile build.
> 
> Like I said, what you want is not provided directly.

Insert visual here of me scratching my head that `coreutils` can't be installed by the user. An emphatic "Nuh-huh" for automagically installing software via a package manager. As the old maps say, "There lie dragons."

> If the ctest stuff is always run, then fine, but actively running ctest seems to be an option and not the default.

In the automated CI/CD builds, they should be, which is one of the reasons why `cmake-builder.ps1` and `cmake-builder.sh` exist (besides making it easier for newbies.)

> If that stays in the ctest, then the 3 minutes vs < 1 second issue should be solved.

Agreed.

**Comment by markpizz** [comment]
> > cmake/cmake-builder.sh will look for realpath and dirname and complain if missing. (Seriously, though, don't you have MacPorts or Brew?)
> My test environment has BOTH HomeBrew and MacPorts since they are required for the makefile build.
>
> Like I said, what you want is not provided directly.

Hmmm... I was thinking that 'brew list coreutils' would list the package of that name, to install or better yet 'brew list | grep coreutils'.  However , it seems that brew list only shows installed packages.  You must use 'brew search coreutils' to find available packages.  In any case, I installed the coreutils package and realpath was found, so cmake-builder.sh could run.

I ran cmake-builder.sh with all the HomeBrew supplied packages needed to build with the makefile already installed and it didn't find the HomeBrew packages and thus dives into building stuff and then ultimately fails:

[cmakerun.log](https://github.com/open-simh/simh/files/9330709/cmakerun.log)

--------------
BTW, my test system is x86, and I don't have access to an ARM architecture box.

**Comment by bscottm** [comment]
@markpizz: I do some check builds on my RPi 4, which is ARM64, and WSL x64. My dev area is `autofs` mounted NFS, which loves symlinks. Having been bitten by the ".." issue with symlinks, `realpath` is a defensive code move. Apologies for the inconvenience.

**Comment by markpizz** [comment]
When I mentioned ARM, I was really pointing out that MacOS (in particular package support) on Intel vs ARM is somewhat different, so testing on that platform is needed.  Paul Koning has one of those.

**Comment by markpizz** [comment]
You haven't removed the Bison, Flex and Npcap path stuff from cmake-builder.ps1 and Npcap path from README-CMake.md yet,

**Comment by markpizz** [comment]
You  still haven't removed the bison & flex dependency check in cmake-builder.ps1.  I will test windows when that is done.

------------------
On Linux (and MacOS), cmake-builder.sh needs to have chmod +x before being run.,  Doing this is a change noticed by git and thus the resulting commit-id ends up having +uncommitted-changes.

------------------
On Linux, when the various targets resulting from cmake-builder.sh are installed in the BIN directory, all targets BUT microvax3900 have execute mode enabled.

---------------
in CMakeLists.txt, there appears to be a typo where "message" is mistyped as "messge" 2 times.

**Comment by pkoning2** [comment]
Homebrew and Macport both required?  That doesn't sound right.  I (grudgingly) installed Homebrew, I haven't used Macport in ages, never liked it.
Actually, I don't want either of them.  The machinery should pick up dependencies if they are simply installed from source, GNU style (for those that are GNU type projects, as most are).  For example, that means finding them in /usr/local rather than in oddball places like /opt.

**Comment by bscottm** [comment]
> Homebrew and Macport both required?

No, both are not required. I think @markpizz has both installed. Pick one, have fun. Just so long as you install GNU `coreutils` :-)

**Comment by markpizz** [comment]
I install and remove each respectively to reflect the possible normal user setups to test against.  I hadn't read the fine print about what 'brew list' produces and thus didn't realize I really wanted 'brew search'.  I 'assumed' that brew behaved line apt and port...  :-(

**Review by markpizz — CHANGES_REQUESTED** [review]
Notice that the same problem exists 2 lines above this change.  I reported this to you yesterday morning....

**Comment by XVilka** [comment]
> @XVilka: In the `README.md` or `README-CMake.md`?

The short version right in README, or at least a paragraph about the basic build steps and a link to the `README-CMake.md`, then users will know exactly what to do. Also, I suggest renaming `README-CMake.md` into `BUILDING.md` or something like that to reflect the purpose better.

**Comment by bscottm** [comment]
@markpizz Your remaining notes have been addressed. I tried to add the `cmake/cmake-builder.sh` script to the Github Action, ran into the error message (because `getopt -T` doesn't exist with a status value of 4.) Need to diagnose this on a separate branch.

**Comment by bscottm** [comment]
@markpizz: Yes, I'm aware that the CMake build bails on macos-10.15. Am investigating the logs for something obvious. Interestingly, the ubuntu builds work. Might shuffle the OS matrix to see if macos-latest works.

**Comment by pkoning2** [comment]
FWIW, I build SIMH all the time on my Mac -- my current laptop is 11.6.8 and I have others that run various older versions.  That's with the Makefile.  Obviously, if that works, the CMake version should work as well.

**Comment by bscottm** [comment]
@pkoning2: You think?

**Comment by markpizz** [comment]
I have several issues/observations with the current state of the cmake stuff.  
These comments are specifically about things on Windows.
I will separate each issue with separate horizontal lines like this:

-----------------------------
 Issue # 1
One of the goals when describing how/what is necessary to perform builds (on any platform) is to consistently describe the minimal set of components of the tool chain and other dependencies that are needed to get started.

Recent versions of Visual Studio include cmake built-in (at least starting at VS2017 and including VS2019 and VS2022).  The build-in cmake should be preferred over externally installing cmake.  (fewer tool-chain installs).

The Windows installed compiler and co-installed build tools are not in user account PATHs.  Each of these Visual Studio versions has a menu selection which starts with the appropriate path(s) to the relevant tools for that activity.  The cmake windows instructions should describe everything needed to have a working environment that they can then invoke the cmake-builder.ps1 script.

Note also, that users will have to change their account setup to allow execution of unsigned powershell scripts.

Let say you get far enough along to have cmake in your path in a running powershell which allows execution of arbitrary scripts and you happen to have Visual Studio 2019 installed.

You enter the cmake-build.ps1 command and the following results:

```
PS C:\Users\Mark\Documents\open-simh> .\cmake\cmake-builder.ps1
** cmake-builder.ps1: cmake is 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe'
** cmake-builder.ps1: SIMH top-level source directory is C:\Users\Mark\Documents\open-simh
** cmake-builder.ps1: Creating C:\Users\Mark\Documents\open-simh\cmake\build-vs2022 subdirectory
** cmake-builder.ps1: Configuring and generating
cmake.exe : CMake Error: Could not create named generator Visual Studio 17 2022
At C:\Users\Mark\Documents\open-simh\cmake\cmake-builder.ps1:367 char:9
+         & ${cmakeCmd} ${generateArgs} 2>&1
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (CMake Error: Co... Studio 17 2022:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

Not searching for unused variables given on the command line.
Generators
* Visual Studio 16 2019        = Generates Visual Studio 2019 project files.
                                 Use -A option to specify architecture.
  Visual Studio 15 2017 [arch] = Generates Visual Studio 2017 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 14 2015 [arch] = Generates Visual Studio 2015 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 12 2013 [arch] = Generates Visual Studio 2013 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 11 2012 [arch] = Generates Visual Studio 2012 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 10 2010 [arch] = Generates Visual Studio 2010 project files.
                                 Optional [arch] can be "Win64" or "IA64".
  Visual Studio 9 2008 [arch]  = Generates Visual Studio 2008 project files.
                                 Optional [arch] can be "Win64" or "IA64".
  Borland Makefiles            = Generates Borland makefiles.
  NMake Makefiles              = Generates NMake makefiles.
  NMake Makefiles JOM          = Generates JOM makefiles.
  MSYS Makefiles               = Generates MSYS makefiles.
  MinGW Makefiles              = Generates a make file for use with
                                 mingw32-make.
  Green Hills MULTI            = Generates Green Hills MULTI files
                                 (experimental, work-in-progress).
  Unix Makefiles               = Generates standard UNIX makefiles.
  Ninja                        = Generates build.ninja files.
  Ninja Multi-Config           = Generates build-<Config>.ninja files.
  Watcom WMake                 = Generates Watcom WMake makefiles.
  CodeBlocks - MinGW Makefiles = Generates CodeBlocks project files.
  CodeBlocks - NMake Makefiles = Generates CodeBlocks project files.
  CodeBlocks - NMake Makefiles JOM
                               = Generates CodeBlocks project files.
  CodeBlocks - Ninja           = Generates CodeBlocks project files.
  CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
  CodeLite - MinGW Makefiles   = Generates CodeLite project files.
  CodeLite - NMake Makefiles   = Generates CodeLite project files.
  CodeLite - Ninja             = Generates CodeLite project files.
  CodeLite - Unix Makefiles    = Generates CodeLite project files.
  Eclipse CDT4 - NMake Makefiles
                               = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - MinGW Makefiles
                               = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.
  Kate - MinGW Makefiles       = Generates Kate project files.
  Kate - NMake Makefiles       = Generates Kate project files.
  Kate - Ninja                 = Generates Kate project files.
  Kate - Unix Makefiles        = Generates Kate project files.
  Sublime Text 2 - MinGW Makefiles
                               = Generates Sublime Text 2 project files.
  Sublime Text 2 - NMake Makefiles
                               = Generates Sublime Text 2 project files.
  Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
  Sublime Text 2 - Unix Makefiles
                               = Generates Sublime Text 2 project files.
==== Last exit code 1
** cmake-builder.ps1: Configuration errors. Exiting.
PS C:\Users\Mark\Documents\open-simh>
```

AND you've got 2 new new directories:

       C:\Users\Mark\Documents\open-simh\cmake\dependencies
AND
       C:\Users\Mark\Documents\open-simh\cmake\build-vs2022

------------
Notice the build directory says build-vs2022 and not build-vs2019.  The cmake-builder.ps1 script should detect the Visual Studio version and make that directory.

Lets say I remove the above-directories and this time I invoke this command in PowerShell "cmake-builder.ps1 --flavor vs2019"  this actually works and runs to completion with simulator binaries.

----------

Additional feedback later.  I'm traveling with family for the next several days so I may not be able respond very completely until late Sunday or Monday.

**Comment by bscottm** [comment]
@markpizz: What's the "tl;dr" version?

**Comment by bscottm** [comment]
@LegalizeAdulthood: I've addressed everything with the exception of updated documentation (README-CMake.md lags a bit) and @markpizz's musings from a few days ago. There should not be any remaining blockers from starting a code review for a merge.

It seems to me that Mark's musings better discussed over on the Wiki than in the middle of a merge request. I didn't see anything in there that is a show-stopper. Two of his issues are easily solved via documentation updates, i.e., the default is VS 2022, if you need VS 2019, do <foo>. Not much that I can do about unsigned PowerShell scripts; I prefer the additional scripting expressiveness, so recoding in DCL<^H><^H><^H>CMD is not much of an option for me.

@pkoning2: MacOSX builds work again. Also made a consistency pass through package detection and compiler flags so that `libpng` version mismatch messages at runtime don't happen. There's a lot of extra stuff in the MacOSX runners that can cause runtime dylib issues. Basically, if `pkg-config` is installed, use `pkg-config` on MacOSX and Linux. The (somewhat cruftier) CMake FindPackage approach is only used if `pkg-config` isn't installed or the platform is Windows.

**Comment by bscottm** [comment]
@markpizz: Seems like I'm pissing into the wind by contributing to `open-simh`. Have a nice day.

**Comment by bscottm** [comment]
I've squashed branch in my repo. Feel free to use it if it meets the project's needs. I'm still willing to respond to issues. I just cannot work with Mark.

**Comment by pkoning2** [comment]
Trying this out on a Mac gave me a message during the "cmake" operation:

```
pkoning:cmake-build pkoning$ cmake ..
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- CMAKE_BUILD_TYPE defaulted to Release
-- Adding HomeBrew paths to library and include search
-- Setting NO_DEP_BUILD to TRUE, will NOT BUILD missing dependencies
-- Replacing '-O3' with '-O2'
-- Adding Clang-specific optimizations to CMAKE_C_FLAGS_RELEASE
--     -fno-strict-overflow
-- Found PkgConfig: /opt/homebrew/bin/pkg-config (found version "0.29.2") 
-- SIM_GIT_COMMIT_ID:   e03a7451a0a2248c89733f3d3cbf3d91ac8be24f
-- SIM_GIT_COMMIT_TIME: 2022-08-22T13:38:00-07:00
-- Git detected uncommitted changes.
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Looking for semaphore.h
-- Looking for semaphore.h - found
-- Looking for sem_timedwait
-- Looking for sem_timedwait - not found
-- Looking for sem_timedwait
-- Looking for sem_timedwait - not found
-- Looking for sys/ioctl.h
-- Looking for sys/ioctl.h - found
-- Looking for linux/cdrom.h
-- Looking for linux/cdrom.h - not found
-- Looking for utime.h
-- Looking for utime.h - found
-- Looking for glob.h
-- Looking for glob.h - found
-- Looking for fnmatch.h
-- Looking for fnmatch.h - found
-- Looking for sys/mman.h
-- Looking for sys/mman.h - found
-- Looking for shm_open
-- Looking for shm_open - found
-- Looking for dlfcn.h
-- Looking for dlfcn.h - found
-- Looking for dlopen
-- Looking for dlopen - found
-- Looking for linux/if_tun.h
-- Looking for linux/if_tun.h - not found
-- Looking for net/if_tun.h
-- Looking for net/if_tun.h - not found
-- Checking for module 'ncurses'
--   Found ncurses, version 5.7.20081102
-- Checking for module 'zlib'
--   Found zlib, version 1.2.11
-- Checking for module 'libpcre'
--   Found libpcre, version 8.45
-- Checking for module 'libpng16'
--   Found libpng16, version 1.6.37
-- Checking for module 'harfbuzz'
--   No package 'harfbuzz' found
-- Checking for module 'freetype2'
--   Found freetype2, version 24.1.18
-- Checking for module 'sdl2'
--   Found sdl2, version 2.0.22
-- Checking for module 'SDL2_ttf'
--   Found SDL2_ttf, version 
-- Checking for module 'vdeplug'
--   No package 'vdeplug' found
-- Found PCAP: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/local/include  
-- pcap_compile requires BPF_CONST_STRING
-- 
-- 
-- Libraries and features:
--   Thread support ................... Platform-detected threading support
--   zlib ............................. pkg-config ZLIB
--   Perl-Compatible RegExps........... pkg-config pcre
--   PNG, Freetype, SDL2, SDL2_ttf .... pkg-config SDL2_ttf, pkg-config SDL2, pkg-config Freetype, pkg-config PNG
--   Network support .................. PCAP dynamic, NAT(SLiRP)
--   ncurses .......................... pkg-config ncurses
--   OS Feature definitions ...........
--     HAVE_SYS_IOCTL
--     HAVE_UTIME
--     HAVE_GLOB
--     HAVE_FNMATCH
--     HAVE_SHM_OPEN
--     SIM_HAVE_DLOPEN=dylib
--   OS Feature libraries ............. m
-- 
-- Build options:
--   Build with video/graphics support. TRUE
--   Build with networking support .... TRUE
--   Build internal ROMS .............. Yes
--   'cppcheck' ....................... 'cppcheck' not installed.
-- 
-- Found PythonInterp: /Library/Frameworks/Python.framework/Versions/3.11/bin/python3 (found version "3.11") 
-- Added makefile update rule.
-- Updating cmake/simh-simulators.cmake
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/pkoning/Documents/svn/open-simh/simh/cmake/generate.py", line 141, in <module>
    sims = process_makefile(makefile_dir, debug=debug_level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pkoning/Documents/svn/open-simh/simh/cmake/generate.py", line 40, in process_makefile
    for all_targ in SPM.shallow_expand_vars(all_rule, defs).split():
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pkoning/Documents/svn/open-simh/simh/cmake/simgen/parse_makefile.py", line 139, in shallow_expand_vars
    m = _var_rx2.search(s)
generate.py: Processing /Users/pkoning/makefile
generate.py: "all" rule not found. Cannot process.
        ^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
-- Adding simulators
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/pkoning/Documents/svn/open-simh/simh/cmake-build
```

That "generate.py" issue didn't seem to break things, but it's disconcerting.  Any idea?

BTW, I tried it with the Ninja generator and that built fine, and fast.  Tried just pdp11 so far.

bscottm:
@pkoning2: I'll look into it. From the traceback, the script seems to have found the wrong makefile.

pkoning2:
Curious.  It's running as part of the generation of the build files, so that suggests there might not be a makefile at all.  Especially if I'm using a builder other than Make.

pkoning2:
I just did some more experiments.  I noticed the message refers to /users/pkoning/makefile.  That certainly doesn't exist.  And it means the script is looking in the wrong place, since that's my home directory.  The build directory is someplace entirely different.

pkoning2:
Looking at PDP11/CMakeLists.txt -- I see that's a generated file.  Interesting.  But it seems to say that "pdp11" is built only if BUILD_WITH_VIDEO is true.  That doesn't seem right; the intent is that pdp11 is built either way, but perhaps with certain files omitted or conditional compiles turned off.

LegalizeAdulthood:
@bscottm I wouldn't let mark "pizz" you off.  He's not the owner of this repository and isn't the one making decisions on what would make this pull request a go/no-go.  Paul Koning @pkoning2 is the owner.  After VCF midwest, I'll take a look at creating a new pull request from the work you've done, which is much farther along than what I had in my fork.

bscottm:
@pkoning2: Yes, the PDP11/CMakeLists.txt is a generated file. It is generated by reading the simh top-level makefile (curiously, though, CMake finds your Python interpreter in a strange place, but it's Python 3, so...) The `cmake/generate.py` script starts searching for the top-level makefile starting in the directory in which the `generate.py` script is located, if the `__file__` variable is correctly set by the Python interpreter.

BTW: `generate.py` does not rewrite files until it absolutely needs to rewrite files. OTOH, there shouldn't be a need to regenerate, so another thing to push on the stack.

Re: Building with video (BUILD_WITH_VIDEO), I'll take that as another thing to look into. I interpreted the `makefile` at one point to mean "don't build those targets if video isn't available."

PR #49, opened 2022-08-16
PCRE2 patches

bscottm:
PCRE2 support migrated over from my `simh` repo, tested with WSL and The `makefile`.

Note to self: Change default in CMake to use pcre2 when both are merged.


bscottm:
Retracted. Need to figure out what's going on in my WSL repo clone.

PR #50, opened 2022-08-17
PCRE2 patches

bscottm:
WSL repo issue under control. Working PCRE2 support.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [pull_request]
+on: [pull_request, push]
 
 defaults:
   run:
--- a/.travis/deps.sh
+++ b/.travis/deps.sh
@@ -5,6 +5,7 @@ install_osx() {
     brew install sdl2
     brew install sdl2_ttf
     brew install vde
+    brew install pcre2
 }
 
 install_linux() {
@@ -12,6 +13,8 @@ install_linux() {
     sudo apt-get install -ym libegl1-mesa-dev libgles2-mesa-dev
     sudo apt-get install -ym libsdl2-dev libpcap-dev libvdeplug-dev
     sudo apt-get install -ym libsdl2-ttf-dev
+    ## Remove when accepted: prospective
+    sudo apt-get install -ym libpcre2-dev
 }
 
 install_"$1"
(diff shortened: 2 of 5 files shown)


markpizz:
This is change for the sake of change.

Changing to PCRE2 does not solve any problem with the existing functionality needed by simh., nor does it add anything to the functionality provided there.

You can say that PCRE is old and mature.  It certainly is.  You can say the same thing for much of the C runtime library.  

Both PCRE and PCRE2 are delivered to the world via sourceforge, and if you look at the weekly download statistics for both, we see that PCRE2 has about 2K downloads/wk and PCRE has about 42K downloads/wk.  Clearly a factor of 21 times more downloads indicates that PCRE isn't dead and no longer in use, but is in fact vastly more popular than PCRE2.

The package managers on MacOS and Linux systems all provide PCRE as well as PCRE2, in fact Ubuntu/Debian provides libpcre3-dev which provides both PCRE and PCRE2.

Meanwhile, you're throwing this PCRE migration into the mix in the middle of the cmake assessment ss out of turn.  Finish one project before tangling in other stuff.  There are at least 2 very simple points of cmake feedback I've made that you've not acted or commented on.

bscottm:
It'll be another week or so until @LegalizeAdulthood can take a look at the CMake pull request. If the maintainers decide to pass on this, fine, I can accept that. I intentionally left the PCRE support in the code as a fallback.

PCRE was originally coded for the Exim MTA, per the PCRE web page. Exim changed over to PCRE2 in 2021. Hence, even the original use case for which PCRE was intended is now deprecated.

Re: CMake. Remind me over on that pull request what remains unaddressed _as succinctly as possible_. I thought I'd addressed everything except a documentation issue.

markpizz:
Look back 4 days (3 messages prior in #47).  I identified 4 separate simple problems.  You ignored my specific feedback about misspelling message as messge and then stumbled through 2 separate self discoveries of the same problem.  2 of my inputs remain unanswered.

Meanwhile, PCRE may have indeed been coded for Exim, meanwhile the 21X consumption of PCRE vs PCRE2 suggests that it is VERY useful in many other places and this large majority of uses isn't rushing to convert for the sake of converting.  Maybe the places that are converting have a reason which PCRE2 compels the conversion process for.

PR #53, opened 2022-08-26
CMake build infrastructure II

bscottm:
@pkoning2, @LegalizeAdulthood:

- A new flag, "--novideo" (cmake-builder.sh)/"-novideo" (cmake-builder.ps1), added to disable video support.
- Explicitly tell `generate.py` the `makefile`'s location. Should fix the issue you had with automagic regeneration.
- "--novideo" uncovered a few new issues that I've fixed.

I rebased `master` onto my `cmake` branch. Hence the extra three commits.

PR #62, opened 2022-09-21
README: add note about the restricted simh/simh repository.

pkoning2:
This hopefully will help people avoid the license issue flagged by Mark Pizzolato in the discussion of PR #53.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,7 @@ This is the codebase of SIMH, a framework and collection of computer system simu
 SIMH was created by Bob Supnik, originally at Digital Equipment Corporation, and extended by contributions of many other people.  It is now an open source project, licensed under an MIT open source license (see [LICENSE.txt](LICENSE.txt) for the specific wording).  The project gatekeepers are the members of the [SIMH Steering Group](SIMH-SG.md).  We welcome and encourage contributions from all.  Contributions will be covered by the project license.
 
 The Open SIMH code base was taken from a code base maintained by Mark Pizzolato as of 12 May 2022.  From that point onward there is no connection between that source and the Open SIMH code base.  A detailed listing of features as of that point may be found in [SIMH-V4-status](SIMH-V4-status.md).
+
+## PLEASE NOTE
+
+**Do not** contribute material taken from `github.com/simh/simh` unless you are the author of the material in question.


markpizz:
Hmmm...  I didn't read the words you wrote before merging the PR.  

If I make changes to content originally authored by other folks, the changes are mine and thus those changes are explicitly not licensed elsewhere.

I'd say that the correct wording should be:

**Do not** contribute material taken from `github.com/simh/simh` unless the author of any content changes is other than Mark Pizzolato.

pkoning2:
I don't believe that is completely accurate.

The key point is that the license for all content in the Open SIMH project is the MIT-style open source license stated on the home page.  When you contribute something, you're licensing it under that license.

So the question is: can you, the contributor, in fact make that decision?  If you are the author, then yes.  If you're not (you're bringing in work originated by someone else) then no.  In that second case, the contribution is allowed if and only if the original author licensed it under a compatible license.  So, for example, importing something that was licensed under GPL is not allowed because GPL is not a compatible license.  Similarly, for someone else to bring over something recently authored by you is not allowed because your license for such recent content isn't compatible either.

However, if you submit something from your simh repository of which you are the author, you have the power to license that contribution, as it appears here, under the license Open SIMH uses.  The fact that it was originally licensed differently is not a concern, because as the author you can license a work under any number of different licenses, that is entirely your decision.   If you are the coauthor of a contribution that first appeared on your repository, and you contribute that here, I think the same applies, because your private licenses seems to be compatible with the MIT license as to the work of anyone other than yourself.  So if you waive the restriction on your parr of the work, which you can as author, no other restriction would remain and the entire contribution is thus validly licensed under the MIT license we use.

The one thing no one can do is contribute something under the precondition that it is licensed under any terms other than the license chosen by the Open SIMH project.

rcornwell:
Mark I also believe you can't take contributions from open-simh and apply them to your repository, since doing this you are changing the copyright notice to something the author might not agree with.

markpizz:
I'm not sure what you're saying here:

> So the question is: can you, the contributor, in fact make that decision? If you are the author, then yes. If you're not (you're bringing in work originated by someone else) then no. In that second case, the contribution is allowed if and only if the original author licensed it under a compatible license. 

I was specifically trying to say something that identified the fact that changes, by themselves, are distinctly copyrightable things without regard to who originally wrote the thing being now changed.  When I make commits/changes on behalf of others, I identify them as the author and thus copyrights are theirs and not mine, even when I'm the committer.  My point is that if I personally change something that was provided by the original author under a BSD license, that change is mine and bound by the constrained non FOSS license in github.com/simh/simh repo.

@rcornwell said:
> Mark I also believe you can't take contributions from open-simh and apply them to your repository, since doing this you are changing the copyright notice to something the author might not agree with.

When I pick up things from the open-simh repo, I am making no change to the authorship of that content/change, so the original author's BSD licensed on the content exists.  The open-simh BSD license explicitly allows anyone to make what they pick up closed source if they want...  I'm only making license claims on the changes I make to things.

So getting back to your revised language, it seems to say that someone would be free to pick up something they wrote originally which now includes changes made by me.  This is not so.  My changes have my constrained license.

Meanwhile, I was also trying to say, that anyone (not merely the original author of a piece) can freely pick up any changes in the repo that I did not author,

rcornwell:
Your repository is under a different license. Without the authors permission, you really can't change it. Didn't you complain when someone moved some stuff from simh to open-simh?

pkoning2:
A changed work has joint authorship: the two authors are (a) the author of the original work and (b) the author of the change.  The key point is that a changed work is not copyrighted by the author of the change alone.  That was the point of my analysis of how to treat contributions that involve both original work and changes.  Perhaps the simpler answer is to ask for contributions that come from Mark's repository to have single authorship and be submitted by that sole author.  But I suspect that isn't  technically necessary, though it makes it easier to understand that everything is clean.

rcornwell:
I think to avoid conflict and confusion it is best to have the original author submit to which repositories he wants to. If the original author only contributes to one repo, it stays there until the author says it can be applied to the other.

markpizz:
That would be a different situation than we have.  Anyone contributing to open-simh is explicitly licensing those contributions under the MIT license which allows anyone (me other others) to do what they want with the content including changing the license.

I can therefore pickup stuff from open-simh and do anything with it.  The language I'm trying to get the README to say is that I've made no license constraints on specific contributions by others.  In fact the vast majority of the PR requests in the open-simh repo since it's creation were by me for contributions made by others.  I'm trying to say that anyone can pickup pieces like that from github.com/simh/simh and do what they want with it, including putting it into the open-simh/simh repo.

rcornwell:
>  I'm trying to say that anyone can pickup pieces like that from github.com/simh/simh and do what they want with it, including putting it into the open-simh/simh repo. 

This is not true. Your license says otherwise. So I think it is only fair that you don't pull stuff that authors do not commit to your repository.

markpizz:
> > The language I'm trying to get the README to say is that I've made no license constraints on specific contributions by others. In fact the vast majority of the PR requests in the open-simh repo since it's creation were by me for contributions made by others.   I'm trying to say that anyone can pickup pieces like that from github.com/simh/simh and do what they want with it, including putting it into the open-simh/simh repo.
>
> This is not true. Your license says otherwise. So I think it is only fair that you don't pull stuff that authors do not commit to your repository.

My license very specifically applies to changes BY ME after a certain date in the master branch of the github.com/simh/simh repo.  Please point at language where you think "Your license says otherwise" (i.e. put constraints on changes by others)?

Where does anything say that all of the PR's I've submitted on behalf of others violate the license constraints in github.com/simh/simh?

rcornwell:
Because there is no way of knowing if they might effect the things you consider unchangeable. Note changes you submit to open-simh are fine. I would recommend you not submit changes commit by other authors to open-simh. When you grab changes from open-simh you are changing the license terms under which the authors released them on. While this does not violate the terms of the license it violates the spirit of the license. Now if you were to place simh/simh under the original license there would be no problem.

Rhialto:
On Fri 23 Sep 2022 at 18:11:13 -0700, Mark Pizzolato wrote:
> [...] the MIT license which allows anyone (me other others) to do what
> they want with the content including changing the license.

I don't think the MIT license allows anyone to randomly change the
license of code they take from a MIT-licensed project.

This part makes that impossible:

"""
   The above copyright notice and this permission notice shall be included in
   all copies or substantial portions of the Software.
"""

where the referenced permission notice reads

"""
   Permission is hereby granted, free of charge, to any person obtaining a
   copy of this software and associated documentation files (the "Software"),
   to deal in the Software without restriction, including without limitation
   the rights to use, copy, modify, merge, publish, distribute, sublicense,
   and/or sell copies of the Software, and to permit persons to whom the
   Software is furnished to do so, subject to the following conditions:
"""

Together I read that to mean that nobody can take away the rights that
are given to me here.

Now if it is just about a patch, one can argue about if this constitues
a "substantial portion" or not; but for the Software as a whole it
certainly does.

Now Mark could argue that he isn't taking away my rights on pre-existing
code, only on code he writes new. But those different kinds of code are
commingled in one repository, and from any random line of code it's not
plainly obvious in which category it falls. It would be better if it is
clearly indicated in each source line, for example. Otherwise authors of
the pre-existing MIT code could probably argue that the license they
wished to place on it is being misrepresented.

-Olaf.
-- 
___ "Buying carbon credits is a bit like a serial killer paying someone else to
\X/  have kids to make his activity cost neutral." -The BOFH    ***@***.***

pkoning2:
I'm not going to change the text.  While it might be technically valid to be less restrictive, the current wording is easy to understand and not much of a burden as far as I can see.

PR #68, opened 2022-09-28
simh ui: replace "readline" by "editline".

pkoning2:
This avoids infecting SIMH with the GPL license, since readline, very surprisingly, is GPL rather than LGPL.

Instead we use "editline" if available at build time.

### DIFF
--- a/.travis/deps.sh
+++ b/.travis/deps.sh
@@ -12,6 +12,7 @@ install_linux() {
     sudo apt-get install -ym libegl1-mesa-dev libgles2-mesa-dev
     sudo apt-get install -ym libsdl2-dev libpcap-dev libvdeplug-dev
     sudo apt-get install -ym libsdl2-ttf-dev
+    sudo apt-get install -ym libedit-dev
 }
 
 install_"$1"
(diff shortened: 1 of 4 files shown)


XVilka:
There is also https://github.com/jart/bestline

markpizz:
Wouldn't you want to squash these commits into 1?

markpizz:
Also, I've seen man pages for suggesting that using libedit requires libtermcap.  

It would be hard to imagine that libtermcap isn't installed on essentially every system, but it might be needed at link time when using libedit.  I don't see a harm in adding -ltermcap to the link line when libtermcap is available...

pkoning2:
Neither Linux nor Mac OS seem to require termcap.  And yes, I'll squish it.

markpizz:
> Neither Linux nor Mac OS seem to require termcap

Agreed, but it doesn't hurt.  Someone wrote the man page info probably because it must mean something.  Maybe libedit is now dynamically loading (like we were with readline) and thus no longer needs explicit linking with libtermcap????

pkoning2:
It's weird.  The documentation on Linux says termcap is needed, the one on Mac OS does not.  The homepage for editline says curses is needed but termcap isn't.  And reality says that nothing extra is needed (linking with -ledit and without -ltermcap works just fine).
Anyway, I'll put in termcap if it's not a Mac OS build.

pkoning2:
Now tested on  MacOS arm64,  MacOS x86_64, Linux FC32 x86_64, Linux FC10 i386, Debian arm, NetBSD 9.3 i386.

PR #70, opened 2022-10-06
Harmonize calibration code amongst VAXen.

Rhialto:
Take the version of the code from VAX/vax4nn_stddev.c, vax4xx_stddev.c and vax_stddev.c, and apply it to 730, 750, 780, 820 and 860.
If we can assume that clk_tps always remains 100, this can be simplified further.

### DIFF
--- a/VAX/vax730_stddev.c
+++ b/VAX/vax730_stddev.c
@@ -782,9 +782,11 @@ return SCPE_OK;
 
 t_stat clk_svc (UNIT *uptr)
 {
-sim_activate_after (uptr, 10000);
-tmr_poll = sim_rtcn_calb (100, TMR_CLK);
-tmxr_poll = tmr_poll * TMXR_MULT;                       /* set mux poll */
+int32_t t;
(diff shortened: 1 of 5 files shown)


markpizz:
It's interesting that this PR was proposed and rushed out and several of the changed simulators fail to build....

markpizz:
There were changes to several vax simulators.  Most if not all of the changed simulators fail to build on at least windows.  Lar's CI build reports the build failures on Appveyor.

pkoning2:
I still don't see any specifics.  What fails, on what system, with what error messages?

Rhialto:
I had to find an actual appveyor build. It would have been faster with a concrete reference. Then I stared at it for a bit, and noticed I had corrected the wrong-looking `int32` type to `int32_t`, as is usual standard.
I created #74 to fix that.
Maybe we should start using standard types for these things everywhere.

pkoning2:
I think it's a bit late to use standard types.

PR #72, opened 2022-10-09
AltairZ80: SS1: Fix disable after reset.

hharte:
The CompuPro System Support 1 could not be disabled after enabled and used due to timers causing it to be busy.  Reset properly so that "set ss1 disabled" works after reset.

FYI, @psco


psco:
Thank you @hharte. @pkoning2, this is fine and should be merged.

hharte:
Rebase on top of latest master.

hharte:
@pkoning2 this is ready to go.

PR #74, opened 2022-10-13
Replace standard in32_t by local int32 type.

Rhialto:


Rhialto:
Maybe we should start using standard types for these things everywhere.

markpizz:
@Rhialto, You missed fixing vax750_stddev.c

@pkoning2 This would have been caught by the Appveyor CI build @larsbrinkhoff referred to in his simh mailing list on Wednesday with the subject: "[simh] AppVeyor continuous integration"

PR #75, opened 2022-10-14
PDP11: Add option for 216K memory.

larsbrinkhoff:
The Small ITS timesharing system prefers this amount, leaving 32K free for mapping framebuffer data.


larsbrinkhoff:
That's consistent with the usage I see in historical PDP-11 software; it's usually in words.  In this case the code says the maximum amount of core is `108.` blocks of Kwords.  But I suppose maybe SIMH is geared towards modern users who might be more comfortable with memory sizes in bytes.

pkoning2:
Modern users might be less used to words of memory, but then again one might argue that learning the historic terminology is a part of learning the historic machine.  It's more obvious for machines that don't have 8-bit byte access to memory; it's hard to imagine speaking of kB of memory on a PDP-10 or B5500.

It might be worth at some later point adding sizes in words to the help text for each of those settings.

PR #77, opened 2022-10-15
PDP11: Add DH11 device.

larsbrinkhoff:


pkoning2:
I see an anomaly on RSTS.  With DH configured for 16 lines, and VH (DHU) also, the VH config is as expected but DH shows a spurious second controller:
```
sim> show dh
DH      address=17760020-17760037*, vector=300-304*, BR5, lines=16, 2 units
  DH0   attached to 9958,Line=0,9957;nomessage, 0 current connections
  DH1   
sim> show vh
VH      address=17760540-17760557*, vector=340-344*, BR4, lines=16
        attached to 9956,Line=0,9955;nomessage, DHU mode, 0 current connections
sim> 
```
Is the "modem" setting (on "attach") supposed to enable DM11BB emulation?  RSTS doesn't show it.  Also, while the "modem" option is accepted with an attach that gives a specific line, it isn't accepted for a default attach (as in `attach dh0 9958;modem`)

The output pacing default is different than what other devices do.  I tried both VH and DH, via Telnet.  VH still looks like it's pacing (it's not going flat out -- unless what I'm seeing is RSTS driver limitations) but it's going a whole lot faster than the DH emulation.  It would seem like a good idea to have the new emulation match existing practice.

Apart from these minor points it seems to be working correctly. 

Do old Unix version use DH?  I have a vague memory that they do, but I have no experience with that.  If so, that would be another good case to test.

On the silo thing, that seems like a fine limitation.  The only case I can imagine where that would matter is the DH driver in Typeset-11 (RSX/IAS based) for the VT71 terminal since that is a bulk data transfer terminal rather than a character at a time device.  I don't actually remember, but it seems plausible that it would use a different fifo setting.  As far as I know that software no longer exists so we can't test the theory.

markpizz:
> - The DH device is implemented using two units. That's not a second controller. Is there a way to hide the second unit?
```
UNIT dh_unit[] = {
  { UDATA (&dh_input_svc, UNIT_ATTABLE, 0) },
  { UDATA (&dh_output_svc, UNIT_DIS, 0) }
};
```

> - It's strange that VH is faster than DH. I have not added any kind of throttling to DH.
> - The DH11 has speed settings, but I don't pass them on to tmxr. I can add that.

To get optimal pacing 2 things are required for best operation: 
1) describe the roles that the different units play relating to which units do input vs which do output processing in your dh_reset routine by calling tmxr_set_line_unit and tmxr_set_line_output_unit,
2) pass along OS programmed speed settings into tmxr by setting tmxr_set_port_speed_control also in the reset routine, and then pass OS programmed changes into tmxr with tmxr_set_config_line.

@pkoning2 said:
> Do old Unix version use DH? I have a vague memory that they do, but I have no experience with that. If so, that would be another good case to test.

Clem is the man to look for this.  He's wanted DH implementation for a long time.

markpizz:
FYI, UNIT_DIS just marks a unit as invisible to the output of SHOW DH.  The purpose you are using the second unit for (scheduling some internal activity) is completely unaffected.

Passing the other info in your dh_reset routine about the relationship between lines and the units that do input and output lets tmxr precisely schedule the activation activity of your various units completely considering the input and output states of each line.  You then need to do essentially nothing in your sim_activate() arguments except to provide an upper bound on scheduling on a polling activity time for the unit in question.

Right now you're supporting a SET DH LINES=n which is at least unusual.  I'm pretty sure that the DH device always had 16 lines per controller.  The MTAB entry for this device has 16 as a upper limit and allows setting of any lesser value (including 0).
The VH and DZ devices use this mechanism to enable multiple controllers which would then involve separate register sets for each controller.  I'd suggest removing removing this MTAB entry unless or until you want to add support for multiple controllers.

Add this revised Visual Studio Project for the PDP11 to this PR
[PDP11.zip](https://github.com/open-simh/simh/files/9803072/PDP11.zip)

pkoning2:
It certainly would be logical to support multiple DH11 controllers.

markpizz:
Your reset routine briefly had 2 calls to tmxr_set_line_output_unit specifying both unit 0 and unit 1.
From a quick glance at the code, unit 1 is the output unit and unit 0 does input processing and connection polling.  You need to tell tmxr about the input role for each line.  Look at the pdp11_vh and pdp11_dz for examples.

larsbrinkhoff:
Thanks for the feedback, Mark.  Made some updates to the pull request per your comments.

Yes, I experimented with introducing speed control.  But something broke in testing, and getting that detail right it outside the scope of what I intended to contribute this time.

markpizz:
Your REG declarations had some issues:

Corrected is:
```
REG dh_reg[] = {
  { ORDATAD(DHSCR,  dh_scr,  16, "System Control Register") },
  { ORDATAD(DHNRCR, dh_nrcr, 16, "Next Received Character Register") },
  { BRDATAD(DHLPR,  dh_lpr,  8, 16, DH_LINES, "Line Parameter Regiser") },
  { BRDATAD(DHCAR,  dh_car,  8, 18, DH_LINES, "Current Address Register") },
  { BRDATAD(DHBCR,  dh_bcr,  8, 16, DH_LINES, "Byte Count Register") },
  { ORDATAD(DHBAR,  dh_bar,  16, "Buffer Active Register") },
  { ORDATAD(DHBRCR, dh_brcr, 16, "Break Control Register") },
  { ORDATAD(DHSSR,  dh_ssr,  16, "Silo Status Register") },
  { BRDATAD(DHSILO, dh_silo, 8, 16, 64, "Silo") },
  { NULL }
};
```

larsbrinkhoff:
Thanks!  Updated.

pkoning2:
Thanks Lars!

For a later addition you might do multiple controller support, to handle configs with more than one DH.

PR #80, opened 2022-10-17
Bug fixes for various video code.

larsbrinkhoff:
Fix some bugs that were (probably) introduced by me.

- `vid_ready` in the SIM_DISPLAY struct can be used uninitialized.  This patch sets it to FALSE right away.
- SIMH would complain about an unrecognized event, mostly when multiple windows are in use.  That's because some events come without a valid windowID field, such as OPEN_WINDOW.
- The PDP11 "NG" device had the MATCH_CMD sense reversed.
- On closer reading of code using NG, it turns out that part of the state register is separate for each display.

### DIFF
--- a/PDP11/pdp11_ng.c
+++ b/PDP11/pdp11_ng.c
@@ -1,7 +1,7 @@
 #ifdef USE_DISPLAY
 /* pdp11_ng.c: NG, Knight vector display
 
-   Copyright (c) 2018, Lars Brinkhoff
+   Copyright (c) 2018, 2022, Lars Brinkhoff
 
    Permission is hereby granted, free of charge, to any person obtaining a
    copy of this software and associated documentation files (the "Software"),
@@ -232,11 +232,9 @@ ng_boot(int32 unit, DEVICE *dptr)
 t_stat
 ng_set_type(UNIT *uptr, int32 val, CONST char *cptr, void *desc)
 {
-  //if ((uptr->flags & DEV_DIS) == 0)
-  //return SCPE_ALATT;
-  if (MATCH_CMD (cptr, "DAZZLE"))
+  if (MATCH_CMD (cptr, "DAZZLE") == 0)
     ng_type = TYPE_DAZZLE;
-  else if (MATCH_CMD (cptr, "LOGO"))
+  else if (MATCH_CMD (cptr, "LOGO") == 0)
     ng_type = TYPE_LOGO;
   else
     return SCPE_ARG;
--- a/display/ng.c
+++ b/display/ng.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Lars Brinkhoff
+ * Copyright (c) 2018, 2022 Lars Brinkhoff
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -24,6 +24,7 @@
  * from the authors.
  */
 
+#include <string.h>
 #include "display.h"                    /* XY plot interface */
 #include "ng.h"
 
@@ -51,7 +52,7 @@ extern void _sim_debug_device (unsigned int dbits, DEVICE* dptr, const char* fmt
 int ng_type = 0;
 int ng_scale = PIX_SCALE;
 
-static uint16 status = 0;
+static uint16 status[8];
 static int reloc = 0;
 static int console = 0;
 static int dpc[8];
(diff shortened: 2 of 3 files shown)


tlhackque (commented):
Why is '8' a magic number rather than a #defined constant?

Constants have a way of turning into variables, which is easier and safer when they are initially coded as names rather than magic numbers... when it happens later, instances are missed and bugs happen.

Otherwise, nothing jumps out at me.

larsbrinkhoff:
Good point.  I added a commit to replace the 8 with a symbolic constant.

tlhackque (commented):
See below.

larsbrinkhoff:
Thanks, it was a good thing you wrote back.  It turns out I overreached a bit with the introduction of a symbol.  The decoding of instructions has nothing to do with the number of displays, so those constants 7 and 8 shouldn't be converted to use DISPLAYS.  So I reverted those changes around "inst".

> is the "eight" really $DISPLAYS?

Yes, but on the other hand DISPLAYS should never be anything but 8.  I took your suggestion to enforce this with an asssert.

> And is the "one" a reference to NG devices, or the number of displays?

The number of displays.  The display library only supports one opening window, so only display 0 is shown to the user.

The only documentation for the device is in Appendix A of [AI memo 240A](https://dspace.mit.edu/bitstream/handle/1721.1/6195/AIM-240a.pdf) (page 29).  However, this describes the 11SIM implementation which is slightly different from the hardware implementations we can infer from software using it.

tlhackque:
That sounds good.  I don't see the assert in the commit, however.

larsbrinkhoff:
Forgot to push it.  There.

markpizz:
The merged commit message first line (Bug fixes for various video code. (#80)) does not follow the convention @tlhackque suggested long ago.

Apart from that, for all of the v4 timeframe, I explicitly had folks avoid explicitly references to the assert macro.  Specifically this is because the macro is only defined to do something in debug compiled code and not in release compiled code.  The simh solution for this is the ASSURE macro which is always defined.  In general such assurances will normally appear only in initialization code paths or certainly in device specific code and thus any overhead by having these in released code will not have any impact on performance.

tlhackque:
Rather than educate by exception, the project could use a document that spells out the conventions. 

The need for one has been on my RADAR, and has been discussed by the steering group.  There is a home for it (http://opensimh.org/contributions/ or a link from it), but I have limited bandwidth and more urgent tasks.

Perhaps as the long-time arbiter for the previous project, @markpizz, could start a draft for consideration and review.  Any contribution is welcome, but plaintext, ideally Markdown, would be ideal.  I expect some discussion on what conventions are required, desirable, and what's forbidden.  

I've opened #86 for any further discussions.

With respect to `assert()` - in this particular case it seems appropriate, as it's a static compilation issue, not a runtime logic check.

markpizz:
One of the things which will be listed in the simh conventions is to explicitly avoid non simh include files and local system calls in simulator code.  Interfacing to OS/dependency APIs is the domain of the SCP libraries.  They are positioned to deal with such interactions in the optimal way according to the proper needs and timing of the all the other SCP libraries.

> With respect to assert() - in this particular case it seems appropriate, as it's a static compilation issue, not a runtime logic check.

The specific "#include <assert.h>" falls into this domain.  Surely in this very specific case, the arguments happen to potentially evaluated by compile time code, but the actual assertion failure action is NOT detected at compile time.

This program always compiles and will only produce an assertion failure if compiled with debug:

```
#include <assert.h>
int main (int argc, char **argv)
{
assert (1==2);
}
```

pkoning2:
If you want a compile time assert, there are macros for that; assert is specifically for run-time checks.  I don't know which compilers provide some sort of standard definition of "static_assert" or "compile_time_assert", but it isn't all that hard to construct.

markpizz:
> If you want a compile time assert, there are macros for that; assert is specifically for run-time checks. I don't know which compilers provide some sort of standard definition of "static_assert" or "compile_time_assert", but it isn't all that hard to construct.

That may indeed be true, but that detail seems like extra complexity and having such a thing directly in simulator code would violate the goal of having SCP deal with platform details.  Additionally, I fail to see the advantage of that over the ASSURE macro which has been part or the codebase since March of 2014.

pkoning2:
A possible answer re ASSURE is that it's a non-standard thing that people don't know about, not unless they spend a lot of time digging through source files.

As for "assert only works in programs compiled with DEBUG" -- not true, not if the compiler obeys the 1989 C standard.  According to both Linux and Mac OS manpages, assert is defined by that standard, and is enabled by default but can be turned off by defining preprocessor symbol "NDEBUG".

tlhackque:
Please move this useful discussion from this closed PR to #86.

As a parting comment for the documents, asserts of whatever kind deserve a comment that explains why.  "do not change" isn't helpful; "must fit in bits <x:y> of instruction" is.

I didn't make an issue of it in this case; Lars had already made a number of changes and it didn't seem worth another round.

PR #82, opened 2022-10-20
Changed formating of SIM_GIT_COMMIT display

louis-chretien:
Changed formating of SIM_GIT_COMMIT display to match the format on GitHub.


markpizz (commented):
This commit contains absolutely no code changes.

Meanwhile, in a previous discussion with you back on July 30, 2022, I explained:

Git commit-ids are, by design, a unique identifier for a change set.  It is actually 40 byte string, which can be abbreviated to uniqueness when using any tool that understands anything about git.  How/where do you think you have an advantage with a 7 character possibly unique string vs an 8 character unique string?  Looking at github’s web interface, you may indeed find 7 character unique strings for commit-ids, but if you actually look at the URL those web links point at, you’ll see the full 40 character commit-id.  Not only, that, but github (like all other git tools) really only cares about the unique amount to actually find the commit in question.  Try these various links:

```
https://github.com/open-simh/simh/commit/58809e3
https://github.com/open-simh/simh/commit/58809e35
https://github.com/open-simh/simh/commit/58809e35b
https://github.com/open-simh/simh/commit/58809e35b4
https://github.com/open-simh/simh/commit/58809e35b4b8445cc3f67ff6d8663bc8f2e58154
```
Which are displayed in the issue/pr system as:
https://github.com/open-simh/simh/commit/58809e3
https://github.com/open-simh/simh/commit/58809e35
https://github.com/open-simh/simh/commit/58809e35b
https://github.com/open-simh/simh/commit/58809e35b4
https://github.com/open-simh/simh/commit/58809e35b4b8445cc3f67ff6d8663bc8f2e58154

pkoning2:
The commit ID is a SHA-1 checksum.  That, by definition, is 20 bytes, so 40 hex digits.  Each GIT commit has its checksum attached, that's the ID.
So, given a particular quantity of commits to consider, it's fine to abbreviate the checksum because the probability of two IDs matching when truncated to, say, 8 digits is still extremely low.  But that does depend on the total number of commits.  For example, the Linux repository has so many commits in it that the default number of digits in the abbreviated GIT ID is not sufficient.
With 7 hex digits, you have a roughly 50% probability of a duplicate value in a population of 16k items; with 8 hex digits it takes 64k items to get that.  This is known as the "birthday rule".  So for a moderately active repository like SIMH, 8 hex digits is reasonable; 7 is marginal.  But it's fine to use longer strings; the only reason for abbreviating them is convenience (if you have to type them) or for pretty-printing.

PR #90, opened 2022-10-26
Visual Studio Projects: Report warnings as errors so CI builds fail on warnings

markpizz:
Make visual studio builds fail on warnings like is done with the makefile for gcc and clang


pkoning2:
It's not ideal to have two unrelated items in one PR, but ok...

PR #91, opened 2022-10-26
makefile: don't use LTO

pkoning2:
Don't enable LTO when building with CLang, matching the existing practice with GCC.  This makes everything build cleanly; with LTO enabled (with either compiler) errors appear in a number of simulators.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1276,7 +1276,8 @@ else
   else
     NO_LTO = 1
     ifeq (Darwin,$(OSTYPE))
-      CFLAGS_O += -O4 -flto -fwhole-program
+      #CFLAGS_O += -O4 -flto -fwhole-program
+      CFLAGS_O := -O2
     else
       CFLAGS_O := -O2
     endif


markpizz:
Hmmm.  I did not see any of these errors when building on macOS 12.6 Intel.

Some of these warnings are real and should be fixed.  I was just working with Bill Beach for the swtp6800 bug.  PR coming.

Others may indeed be compiler bugs.  Minimal reproducible cases should be found and passed to the compiler folks...

pkoning2:
I'm thinking they may well be real, including the weird pdp4 messages I originally complained about.  The reason is that GCC, when building with LTO enabled, also complains about pdp4 (and indeed about pdp1 also, and with more errors for both of them).  
I'm not sure how mature LTO is, and I'm very much skeptical about its value for SIMH, which is why I'm proposing to turn it off.  If at some point we have LTO-clean code and believe that it actually is useful, we can always turn it back on later.

pkoning2:
FWIW, I'm building on MacOS 11.7, Apple silicon (ARM64).

markpizz:
Hmmm....  More digging and I have no idea why or how the current makefile lto stuff is supposed to be working or how/why it was originally done the way it has been.  

Testing with both gcc and clang on Fedora 36 (current) Fedora 37 beta - both Intel and there are no problems reported with any of the pdp1, pdp4, pdp11 simulators when "-flto -fwhole-program" are specified.  -fwhole-program not applicable on clang on these systems.  I can't find a Fedora 32 iso...

Was your Linux testing on Intel or ARM architecture compilers?

I'll rework the testing logic to disable LTO based on a make command argument if that doesn't work now.

markpizz:
The goal of CI activities would seem to be to usefully report the most potential problems.  That would suggest that LTO should be enabled during CI builds and maybe be optional during direct user make commands.  This could balance the desire for the highest level of error/problem detection vs reasonably minimal build times during user interactive builds.  Of course, building only the changed files (instead of everything) would seem to be best/fastest during development.

markpizz:
I stand corrected about Fedora 36 gcc -flto detecting issues with intel-mds and i7080.

An updates from Bill Beech for the intel-mds are part of PR #90 which changes warnings to errors for the Visual Studio Projects.

I created #92 for @rcornwell to address the i7080 bug(s).

pkoning2:
On "assuming that an array of 8 byte entities is a string" -- that is a clear error in the compiler.  We should not humor such a defect.

pkoning2:
The Linux test where I saw LTO errors in SIMH builds was on Intel.

markpizz:
I presume your test with ```uint8 ttix_buf[TTX_MAXL + 1]``` avoided the error report, and my guess about the logic in the warning report was right.

> On "assuming that an array of 8 byte entities is a string" -- that is a clear error in the compiler. We should not humor such a defect.

Agreed, and it seems to be fixed in more recent versions.

Must run errands now.  Fix LTO engagement in makefile later.

pkoning2:
Closing, I think on reconsideration this isn't the best answer.

PR #94, opened 2022-10-28
PDP11: Devices for SITS.

larsbrinkhoff:
Includes the raster bitmap display controller and keyboards, and a stub for the bus-snooping memory debug device.  Plus some updates to reduce host CPU usage when SITS is running.


markpizz:
The pdp11_tv.c module has a REGister called TVSEL which lives in 1 byte of storage, but the register description says the storage is 2 bytes.  It should probably be:
```
  { ORDATAD(TVSEL,  TVSEL,      8 "Console select") },
```

larsbrinkhoff:
Thanks @markpizz, I opened this: #100

PR #95, opened 2022-10-29
makefile: Add explicit LTO=1 make option to force gcc and clang LTO

markpizz:
When building compiler optimized binaries with the gcc or clang compilers,
invoking GNU make with LTO=1 on the command line will cause the build
to use Link Time Optimization to maximally optimize the results.
Link Time Optimization can report errors which aren't otherwise detected
and will also take significantly longer to complete.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [pull_request]
+on: [pull_request, push]
 
 defaults:
   run:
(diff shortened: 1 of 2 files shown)


markpizz:
We once had some confusion about how come CI build steps seemed to have gotten canceled.

It seems that if one step fails, then the as yet uncompleted steps are canceled.

pkoning2:
It's my impression that CI builds are canceled (a) when another one fails, (b) if the PR is merged, and (c) probably also if a new commit arrives for the PR.

PR #97, opened 2022-10-29
Fix another int32_t -> int32

Rhialto:
This was missed in 3a95ba4b / #74


markpizz:
In the future or now, will you please follow the long established commit message format, which in this case should be:


VAX750: Fix another int32_t -> int32

This was missed in #74

PR #98, opened 2022-10-30
AltairZ80: HDC-1001: Fix format.

hharte:
Also a minor improvement in the debugging for wd179x.

FYI, @psco

### DIFF
--- a/AltairZ80/wd179x.c
+++ b/AltairZ80/wd179x.c
@@ -913,11 +913,10 @@ static uint8 Do1793Command(uint8 cCommand)
                       " Error: READ_TRACK not implemented.\n", wd179x_info->sel_drive, PCX);
             break;
         case WD179X_WRITE_TRACK:
-            sim_debug(WR_DATA_MSG, &wd179x_dev, "WD179X[%d]: " ADDRESS_FORMAT
-                      " CMD=WRITE_TRACK\n", wd179x_info->sel_drive, PCX);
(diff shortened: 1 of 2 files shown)


AltairZ80/wd179x.c
```diff
@@ -913,11 +913,10 @@ static uint8 Do1793Command(uint8 cCommand)
...
             break;
         case WD179X_WRITE_TRACK:
-            sim_debug(WR_DATA_MSG, &wd179x_dev, "WD179X[%d]: " ADDRESS_FORMAT
-                      " CMD=WRITE_TRACK\n", wd179x_info->sel_drive, PCX);
             sim_debug(FMT_MSG, &wd179x_dev, "WD179X[%d]: " ADDRESS_FORMAT
-                      " CMD=WRITE_TRACK, T:%2d/S:%d.\n", wd179x_info->sel_drive,
-                      PCX, pDrive->track, wd179x_info->fdc_head);
+                      " CMD=WRITE_TRACK, T:%2d/S:%d/N:%d.\n", wd179x_info->sel_drive,
+                      PCX, pDrive->track, wd179x_info->fdc_head,
+                      wd179x_info->fdc_datacount = 128 << wd179x_info->fdc_sec_len);
```
pkoning2, line 919: I don't understand what this is doing in the argument list.  Did you really mean there to be an assignment in the list?  And it doesn't seem to match any % marker in the format.

hharte: You are absolutely correct, @pkoning2 .  It was a copy/paste error on my part.  While it "worked" it definitely was not correct.

FYI, @psco

PR #99, opened 2022-10-30
makefile: Make non debug builds with -O3 and LTO under gcc with -fwhole-program

markpizz:
- Add missing BUILD_ROMS dependency to pdp11 simulator.  ROMs don't change often so this usually didn't matter, but it should be a dependency.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1270,10 +1270,10 @@ ifneq ($(DEBUG),)
   LTO =
 else
   ifneq (,$(findstring clang,$(COMPILER_NAME))$(findstring LLVM,$(COMPILER_NAME)))
-    CFLAGS_O = -O2 -fno-strict-overflow
+    CFLAGS_O = -O3 -fno-strict-overflow
     GCC_OPTIMIZERS_CMD = ${GCC} --help 2>&1
   else
-    CFLAGS_O := -O2
+    CFLAGS_O := -O3
   endif
   LDFLAGS_O = 
   GCC_MAJOR_VERSION = $(firstword $(subst  ., ,$(GCC_VERSION)))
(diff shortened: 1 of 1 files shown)


pkoning2:
That still fails with my Clang. 

It would be helpful to have a way to override the default optimization, without having to edit the makefile.  I did this in a local test:
```
OPTIMIZE ?= -O3
...
CFLAGS_O = $(OPTIMIZE)
```
that produces the result you have in this PR by default, but I could do make OPTIMIZE=-O2 to work around the compiler bugs.

markpizz:
OPTIMIZE= is a good suggestion.  I just pushed an update to include it.

Meanwhile, I'm seeing some very strange behavior.  Compiling the pdp11 as a test case from the open-simh/simh repo as the simh/simh repo I'm seeing the crazy errors on open-simh/simh and not on simh/simh.  Both repos have the same logic relating to LTO and OPTIMIZE.  There are slight differences between these cases:

- sim_tape.c has 2 instances where it calls a new ABORT() macro instead of abort().
- sim_fio.c has various new functionality, but NONE of the API's used by sim_tape (and mentioned in the crazy errors) have any changes at all.

The systems which demonstrate this are Fedora 36 & 37 with gcc 12.2.1 and clang 15.0.0.  Only gcc seems to produce the crazy errors compiling open-simh/simh pdp11 with my latest makefile PR built with LTO.  To try and isolate what's going on, I pulled out all the routines mentioned in the crazy messages into a separate sim_fio2.c added mention of that to the SIM variable in the makefile and compiled and the problem persists.  The only other file which could influence this is sim_defs.h which has changes but none potentially related to the crazy errors.

pkoning2:
It seems to me that LTO should be ignored for gcc, until gcc can actually do it reliably.

markpizz:
gcc -flto -O3 is 100% reliable on github.com/simh/simh..  :-(

LTO has the potential to detect cross module declaration inconsistencies even when compiling with -O2 as it did for the I7080 and intel-MDS.  I just confirmed this with "make intel-mds LTO=1 OPTIMIZE=-O2" after modify the original intel source file to have the incorrect declaration.

I'm thinking best to change default to -O2 for gcc and leave it at -O3 for clang.

Since the CI builds currently run successfully with LTO=1, it looks like we don't need to change anything now.  Maybe rethink in the future.

Please merge @larsbrinkhoff's #96 and #100 and @Rhialto's #97.

pkoning2:
Interesting, since sim_tape has not meaningfully changed between the two forks, and sim_fio makes no difference, and in the pdp11 simulator (which is one of those that fail) no sim_tape calls have changed.  Also, while some simulators that use sim_tape break, others (like b5500) now build successfully, though they didn't earlier.

I'm going to give up on LTO.  I don't know of any evidence that it adds anything useful in the first place, and clearly it is not yet ready for production use.  If and when it does become real, we can reconsider using it.

PR #101, opened 2022-11-01
makefile: don't use -fwhole-program

pkoning2:
The GCC documentation explicitly says not to use that option when -flto is used, and since that is the only place where the makefile was using it, remove it to conform to the documented rules.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1317,9 +1317,6 @@ else
   ifneq (,$(LTO))
     ifneq (,$(findstring -flto,$(GCC_OPTIMIZERS)))
       CFLAGS_O += -flto
-      ifneq (,$(and $(or $(findstring gcc,$(COMPILER_NAME)),$(findstring GCC,$(COMPILER_NAME))),$(findstring -fwhole-program,$(GCC_OPTIMIZERS))))
-        CFLAGS_O += -fwhole-program
-      endif
       LTO_FEATURE = , with Link Time Optimization,
     endif
   endif


markpizz:
From what I thought I read, the -fwhole-program was somehow only relevant if -flto was specified.

From what I just read, the above is true, BUT what it is saying is that when the link is happening, specifying -fwhole-program tells it that the whole program is present for the linker to deal with.  I general this would seem to be true if the simh binaries were linked statically, since there is no other dynamic loading and possible routine or variable occlusion going on.  The current link isn't static, so I agree that -fwhole-program isn't appropriate for production builds (nor is -flto probably), however since it is only directly enabled in the CI builds and that is when we're looking for the most feedback about coding errors so PRs with these get caught before merging, unless the CI builds encounter the crazy errors case, I don't see the harm.  The current CI builds do not encounter the crazy errors...  Surely, we see the crazy errors with some compiler combinations when the LTO builds are engaged, but I think the value of the error reporting
outweighs these corner cases which users have to explicitly ask for.  If/when crazy errors crop up in the CI builds, we will then need to adjust the CI instructions and/or the makefile to avoid them.

pkoning2:
I thought so too but the documentation is quite clear.  Quoting the info page:
```
'-fwhole-program'
     Assume that the current compilation unit represents the whole
     program being compiled.  All public functions and variables with
     the exception of 'main' and those merged by attribute
     'externally_visible' become static functions and in effect are
     optimized more aggressively by interprocedural optimizers.

     This option should not be used in combination with '-flto'.
     Instead relying on a linker plugin should provide safer and more
     precise information.
```

Rhialto:
The way it mentions "the current compilation unit", and assuming that a
"compilation unit" corresponds to a source file (which is I think the
usual interpretation, for systems that have files), the option is only
even suitable for programs that consist only of one source file. That
sounds a bit weird to me...

(Unless gcc suddenly defines "compilation unit" as the whole collection
of files given to the compiler at once, but surely that will lead to
contradictions with the Standard when you also compile them separately
or in arbitrary other subsets)

markpizz:
Go ahead and merge the removal of -fwhole-program.  The -flto seemed to produce the useful real error reports.

pkoning2:
@Rhialto -- "compilation unit" indeed means a source file, and everywhere in GCC I have seen that term used it is used in that standard way.  The description that says all functions other than "main" are made static matches that meaning.

PR #103, opened 2022-11-02
VAX: silence compiler warning

jtcave:
I was unable to build the vax simulator on Ubuntu 22.04 (GCC 11.3) because of a compiler warning in vax_sysdev.c. By adding a test for NULL, I was able to get the simulator to build again.

### DIFF
--- a/VAX/vax_sysdev.c
+++ b/VAX/vax_sysdev.c
@@ -1181,9 +1181,12 @@ while (memsize > 1) {
         }
     for (i=0; boards[i].capacity > memsize; ++i)
         ;
-    fprintf(st, "Memory (@0x%08x): %3d Mbytes (%s)\n", baseaddr, boards[i].capacity, boards[i].option);
-    memsize -= boards[i].capacity;
-    baseaddr += boards[i].capacity<<20;
+    if (boards[i].option != NULL)
+    {
+        fprintf(st, "Memory (@0x%08x): %3d Mbytes (%s)\n", baseaddr, boards[i].capacity, boards[i].option);
+        memsize -= boards[i].capacity;
+        baseaddr += boards[i].capacity << 20;
+        }
     }
 return SCPE_OK;
 }


pkoning2:
Did you get that warning with the latest code?  What options?  Could you show what the warning message was?

jtcave:
Yes, this happens on the latest code. I'm not using any special options, just running `make`. Here's the output of `make vax`:

```
james@SUNRISE:~/simh$ make vax
lib paths are: /lib/ /lib/x86_64-linux-gnu/ /usr/lib/wsl/lib/ /usr/lib/x86_64-linux-gnu/libfakeroot/ /usr/lib/          include paths are:  /usr/lib/gcc/x86_64-linux-gnu/11/include /usr/local/include /usr/include/x86_64-linux-gnu /usr/include                                                                                                                      using libm: /lib/x86_64-linux-gnu/libm.so                                                                               using librt: /lib/x86_64-linux-gnu/librt.a                                                                              using libpthread: /lib/x86_64-linux-gnu/libpthread.a /usr/include/pthread.h                                             using libpcre: /lib/x86_64-linux-gnu/libpcre.so /usr/include/pcre.h                                                     using semaphore: /usr/include/semaphore.h                                                                               using libdl: /lib/x86_64-linux-gnu/libdl.a /usr/include/dlfcn.h                                                         using libpng: /lib/x86_64-linux-gnu/libpng.so /usr/include/png.h                                                        using zlib: /lib/x86_64-linux-gnu/libz.so /usr/include/zlib.h                                                           using mman: /usr/include/x86_64-linux-gnu/sys/mman.h                                                                    using libSDL2: /usr/include/SDL2/SDL.h                                                                                  *** No SDL ttf support available.  BESM-6 video panel disabled.                                                         ***
*** Info *** Install the development components of libSDL2-ttf
*** Info *** packaged for your Linux operating system distribution:
*** Info ***        $ sudo apt-get install libsdl2-ttf-dev
using libpcap: /usr/include/pcap.h
using libvdeplug: /lib/x86_64-linux-gnu/libvdeplug.so /usr/include/libvdeplug.h
***
*** vax Simulator being built with:                                                                                     *** - compiler optimizations and no debugging support. GCC Version: 11.3.0.
*** - dynamic networking support using Linux provided libpcap components.
*** - Local LAN packet transports: PCAP VDE TAP NAT(SLiRP)
*** - video capabilities provided by libSDL2 (Simple Directmedia Layer).
*** - Per simulator tests will be run.
***
*** git commit id is 009d748ad69701c10f27af530f0b52b3a43f44ec.
*** git commit time is 2022-11-02T14:05:31+0100.
***
gcc -std=gnu99 -U__STRICT_ANSI__  -O3 -finline-functions -fgcse-after-reload -fpredictive-commoning -fipa-cp-clone -fno-unsafe-loop-optimizations -fno-strict-overflow -DSIM_GIT_COMMIT_ID=009d748ad69701c10f27af530f0b52b3a43f44ec -DSIM_GIT_COMMIT_TIME=2022-11-02T14:05:31+0100  -DSIM_COMPILER="GCC Version: 11.3.0" -DSIM_BUILD_TOOL=simh-makefile -I . -Werror -D_GNU_SOURCE -DUSE_READER_THREAD -DSIM_ASYNCH_IO  -DHAVE_PCRE_H -DHAVE_SEMAPHORE -DHAVE_SYS_IOCTL -DHAVE_LINUX_CDROM -DSIM_HAVE_DLOPEN=so -DHAVE_UTIME -DHAVE_LIBPNG -DHAVE_ZLIB -DHAVE_GLOB -DHAVE_SHM_OPEN  ./VAX/vax_cpu.c ./VAX/vax_cpu1.c ./VAX/vax_fpa.c ./VAX/vax_io.c ./VAX/vax_cis.c ./VAX/vax_octa.c  ./VAX/vax_cmode.c ./VAX/vax_mmu.c ./VAX/vax_stddev.c ./VAX/vax_sysdev.c ./VAX/vax_sys.c  ./VAX/vax_syscm.c ./VAX/vax_syslist.c ./VAX/vax_vc.c ./VAX/vax_lk.c ./VAX/vax_vs.c ./VAX/vax_2681.c ./PDP11/pdp11_rl.c ./PDP11/pdp11_rq.c ./PDP11/pdp11_ts.c ./PDP11/pdp11_dz.c ./PDP11/pdp11_lp.c ./PDP11/pdp11_tq.c ./PDP11/pdp11_xq.c ./PDP11/pdp11_vh.c ./PDP11/pdp11_cr.c ./PDP11/pdp11_td.c ./PDP11/pdp11_io_lib.c ./PDP11/pdp11_dup.c ./scp.c ./sim_console.c ./sim_fio.c ./sim_timer.c ./sim_sock.c ./sim_tmxr.c ./sim_ether.c ./sim_tape.c ./sim_disk.c ./sim_serial.c ./sim_video.c ./sim_imd.c ./sim_card.c -DVM_VAX -DUSE_INT64 -DUSE_ADDR64 -DUSE_SIM_VIDEO -I ./VAX -I ./PDP11 -DHAVE_PCAP_NETWORK -I/usr/include/ -DBPF_CONST_STRING -DUSE_SHARED -DHAVE_VDE_NETWORK -DHAVE_TAP_NETWORK -Islirp -Islirp_glue -Islirp_glue/qemu -DHAVE_SLIRP_NETWORK -DUSE_SIMH_SLIRP_DEBUG slirp/*.c slirp_glue/*.c -DHAVE_LIBSDL -DUSE_SIM_VIDEO `sdl2-config --cflags` `sdl2-config --libs` -o BIN/vax -lm -lrt -lpthread -lpcre -ldl -lpng -lz -lvdeplug -Wl,-R,/lib/x86_64-linux-gnu/ -L/lib/x86_64-linux-gnu/
In file included from ./sim_defs.h:1119,
                 from ./VAX/vax_defs.h:51,
                 from ./VAX/vax_sysdev.c:58:
./VAX/vax_sysdev.c: In function ‘cpu_show_memory’:
./scp.h:193:17: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  193 | #define fprintf Fprintf
./VAX/vax_sysdev.c:1184:5: note: in expansion of macro ‘fprintf’
 1184 |     fprintf(st, "Memory (@0x%08x): %3d Mbytes (%s)\n", baseaddr, boards[i].capacity, boards[i].option);
      |     ^~~~~~~
cc1: all warnings being treated as errors
make: *** [makefile:2282: BIN/vax] Error 1
```

markpizz:
Hmmm..  The compiler is completely wrong on it's analysis.

Try the following and step through the cpu_show_memory() routine:

```
sim> SET CPU 512
sim> SHOW MEMORY
```

The for loop:
```
    for (i=0; boards[i].capacity > memsize; ++i)
        ;
```
is never executed when memsize is > 64M.

Try:
```
make vax OPTIMIZE=-O2
```

Rhialto:
Would the compiler be smart enough to shut up if you change the `if` condition to an `assert(boards[i].option)`? (If I were a compiler builder I'd put that in my compiler, but of course I'm not...)

pkoning2:
It works on GCC 10 and GCC 12, also Clang.  Seems like GCC 11 turned on an extra warning (-Werror=format-overflow) which isn't on by default in those other versions.  If I turn it on explicitly I see that same message.

It's a spurious error because the case of a null pointer doesn't actually happen, but indeed the check will suppress it.

Yes, an assert suppresses it also.  Mark, the loop is never executed, period -- because the first entry in "boards" has size 16M and 16M is also the smallest memory size in that system.

pkoning2:
I assume the assert fixes it because it expands into a check for null which then does an abort, so by the time the printf is reached the compiler knows the value is not null.

Assert is actually not the best answer because if you compile with -DNDEBUG, the assert goes away and the message comes back.  So I think the workaround in this PR is the best answer for the compiler bug.

markpizz:
The vax simulator is not the only one that has new issues when building with gcc and -O3 on this version of ubuntu.

We should go back to -O2 and avoid the wack-a-mole problems -O3 introduces.  Compiler static analysis seems to be shaky at best.

gcc 12.2.1 (as reported in this issue) DOES not produce a problem on fedora 36.

pkoning2:
Mark, I agree with that suggestion.  The question then becomes whether to accept this PR anyway. 

It may be a good idea for the check to have a comment on it saying that it's a redundant check that's there to avoid spurious compiler warnings for some compiler versions and optimization levels.  That will help code readers understand it's not actually testing for a condition that can occur in reality.  @jtcave could you add that?

markpizz:
> It may be a good idea for the check to have a comment on it saying that it's a redundant check that's there to avoid spurious compiler warnings for some compiler versions and optimization levels. That will help code readers understand it's not actually testing for a condition that can occur in reality. @jtcave could you add that?

This is the wack-a-mole that we should strive to avoid.  There is absolutely no need for an assert in that code and the other places which display screwy errors can't be addressed by asserts.  The pdp4 problem is among the other simulators which don't compile on that system.

pkoning2:
Another option would be to disable the defective warnings: `-Wno-format-overflow -Wno-stringop-overflow`

markpizz:
If there were only this one failing case producing a warning, maybe comments might be needed.  Since there isn't I don't think so.

Maybe we add a WARNINGS=ALLOWED as a non default makefile option and then all the warnings will be issued and probably working simulators will be produced.  This series of problems started with the addition of -Werror

You said:
> Another option would be to disable the defective warnings: -Wno-format-overflow -Wno-stringop-overflow

There has absolutely been very useful format warnings issued by the compilers over time, so I wouldn't remove any of them.  I think that going back to default -O2 and adding the WARNINGS=ALLOWED will allow Joe User to get builds when they want, but we've added maximal real warnings to fail in the CI builds so real problems don't creep into the codebase.

markpizz:
The code which inspired this PR has been 100% stable for 9+ years.  There is nothing suddenly there which warrants any changes, so accepting this PR isn't the right approach.  

Addressing the underlying cause for the problems while allowing users to get varied behavior depending on their desires and needs to stretch beyond the defaults.  I've created #104 to change the makefile and the CI run parameters

jtcave:
If this is specifically due to a GCC 11 bug, then I agree with @markpizz that this isn't the right approach. I've also seen a similar issue building the S3 simulator on my system due to `format-truncation` warnings, and playing "satisfy the compiler" with those would absolutely not be worth the effort. Having -O3 and -Werror be opt-in is a better workaround than this PR. As such, I'll go ahead and close this.

PR #104, opened 2022-11-02
makefile: Add WARNINGS=ALLOWED make option and change default optimize to -O2

markpizz:
The CI build specifies OPTIMIZE=-O3 and runs with the default where warnings are converted to errors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,4 +32,4 @@ jobs:
       - name: Build
         env:
           SIM: ${{matrix.simulators}}
-        run: make LTO=1 $SIM
+        run: make LTO=1 OPTIMIZE=-O3 $SIM
--- a/S3/s3_sys.c
+++ b/S3/s3_sys.c
@@ -272,7 +272,7 @@ int32 printf_sym (FILE *of, char *strg, t_addr addr, uint32 *val,
 {
 int32 c1, c2, group, len1, len2, inst, aaddr, baddr;
 int32 oplen, groupno, i, j, vpos, qbyte, da, m, n;
-char bld[128], bldaddr[96], boperand[32], aoperand[32];
+char bld[128], bldaddr[160], boperand[32], aoperand[32];
 int32 blk[16], blt[16];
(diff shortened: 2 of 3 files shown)


markpizz:
As a follow-up to #103 

Recent changes to the makefile build logic which added support to abort builds on warnings along with messing around with optimization levels and link time optimization (LTO) has generated a ton of tooth gnashing which, in general, we haven't had to deal with essentially forever.   Getting the CI builds to provide as much relevant feedback before changes are added to the codebase is most useful and important.  Minimizing tooth gnashing is also a goal, so setting defaults and allowing options should work well.

Meanwhile, building the s3 simulator with -O3 on the Ubuntu 22-04 system reports a couple of warnings, one which is truly a coding error.  The error is the walk through the regname array references outside the bounds of the array when it references regname[15] with the regname array having only 15 elements.  While fixing this real problem, it is worth a look at the claim of string truncation.  The string truncation claim is completely incorrect.  The target buffer is 128 bytes long and the format string has one constant character plus 2 string inserts.  The first string insert references an 8 byte array element which as a proper string could contain 7 characters + a \0, the second string is coming from a 32 byte buffer which can contain 31 characters plus a \0.  This should total 40 plus a \0, which is clearly less than the 128 buffer size.  The wording of the warning message suggests that the first string is 120 bytes rather than 7 (or 8).  That would be the entire size of the array that is referenced rather than one of it's elements.  This seems like a gcc bug.  clang 14.0.0 does not complain about this.  This complaint comes from gcc with both -O2 and -O3.

I'm pretty sure that some string overflow buffer size checking has appeared in Coverity scans over time and just to shut up those Coverity issues, changes have been made to silence the Coverity messages.  Given the makefile stop on warning AND the more aggressive gcc incorrect checking we're now experiencing, since this s3_sys.c buffer problem is the only warning that's reported when compiling all the simulators with -O2, I'm changing both the above mentioned bug and increasing the target buffer size since a bug fix is needed in that source module anyway.

Given I've got this open PR, the s3_sys.c changes are an additional commit here.

PR #105, opened 2022-11-04
Fix "non-void function does not return a value" warning

jbglaw:
Found with Debian clang version 15.0.3-2 and LTO=1.


markpizz:
Please revise this commit to have the project standard commit message format which is:

<affected-component-or-simulator>: 1-line summary of the commit's purpose limited to a total of 80 characters
<blank-line-if-extended-commit-messages-is-appropriate>
<extended-commit-messages wrapped < 80 characters>

Please force push the revised commit replacing the original commit.

PR #106, opened 2022-11-06
AltairZ80: i86: Fix bugs in REPE/REPNE as well as AAM instructions.

hharte:
REPE_5170
Fixed a couple of bugs in the i86 CPU core.

Verified the REPE/REPNE segment override behavior against a real 80286 CPU, and also assembled MS-DOS 2.11 from source, as well as ran CP/M-86 and rebuilt the BIOS with success.

For the AAD instruction, the next byte was not fetched, causing the instruction sequence to be corrupted.  The fix is to align the behavior with AAM.

FYI @psco as discussed.


psco:
@hharte Thank you Howard for this thorough analysis and bug fix!

PR #111, opened 2022-11-12
BUILDROMS: Fix potential compiler warnings

markpizz:
If the Appveyor Windows CI build were working we'd have problems with these warnings.

The makefile build of BuildROMs doesn't use it as a normal make target but instead compiles and runs it under the covers without the CFLAGS values.  This is OK since it really isn't a target, but a pre-build sanity check to confirm the ROM inputs haven't changed.


markpizz:
Change logic in extended simulator tests (pdp8, pdp9, vax, sel32, i650) to limit runtime by instructions rather than wall clock time.   This will then allow the tests to work on arbitrarily slow or busy host system which may be the case for CI builds.

PR #113, opened 2022-11-17
RK source consistency

al20878:
Please consider the following fix for source code typos


pkoning2:
I noticed another typo: in the help text, 128 words/sector (correct) and 256 words/drive (should be "sector").

markpizz:
> I noticed another typo: in the help text, 128 words/sector (correct) and 256 words/drive (should be "sector").

Good catch.

Meanwhile, once again, the commit comment should follow the project standard.  I suggest:

PDP11: Correct wording for RK MTAB entries and help text

al20878:
I updated the commit message, and also included the fix for words/sector that Paul found.
(The cyl count is a copy-n-paste blunder from the RL code.)

pkoning2:
I was wondering about that 2* thing.

PR #114, opened 2022-11-17
1ms timer resolution in Cygwin (independent of Windows version)

al20878:
This patch was suggested a long time ago for SimH on Cygwin to have a consistent 1ms timer (rather than 10ms / 1ms depending on *other* applications running on the Windows host -- if they happen to use the performance setting).  This fix does not affect any other platform (including the native Windows builds).


markpizz:
This looks OK, but the commit message format needs to conform to the project standard.  Suggest it should be:

TIMER: Support 1ms timer resolution in Cygwin

al20878:
I fixed the commit message per the suggestion.  I'll try to keep the format in mind in any future requests.

markpizz:
The word consistent in the commit message adds confusion not clarity which is why I suggested the simpler form.

If you want to include consistent, then explain in detail what it means in the additional commit message content (skip a line, and then add additional commit message content wrapped < 80 characters).

al20878:
IMO "consistent" provides additional information esp. for Cygwin users who might have experienced the timer accuracy to spontaneously change between 1ms and 10ms, depending on which other Windows applications were currently running (or starting / stopping) in the system.  But fine, I dropped the "consistent".

PR #122, opened 2022-11-20
AltairZ80: Add CP/M DDT-style stopped display option

deltecent:
SCP supports custom stopped message generation by setting `sim_vm_fprint_stopped`.  This PR first changes how `sim_vm_fprint_stopped` is processed. Prior to this PR, `sim_vm_fprint_stopped` was only used if the stop status was an SCP error (>= SCPE_BASE).  This would ignore `sim_vm_fprint_stopped` when, for example, single-stepping. This PR changes `fprint_stopped_gen()` to always invoke `sim_vm_fprint_stopped` if it is set.  If the `sim_vm_fprint_stopped` handler returns FALSE, `fprint_stop_gen()` does not continue with the standard output. This puts more control in the hands of the VM rather than SCP.

This PR also adds a "DDT" option to the CPU device that implements `sim_vm_fprint_stopped` with CP/M DDT-style stop output.

```
"SET CPU DDT" output:

Simulation stopped
C0Z1M0E1I1 A=02 B=007F D=DF06 H=EA0E S=EA37 P=FA6B ANI 01h

"SET CPU NODDT" output:

Simulation stopped, PC: 0FA6B (ANI 01h)
```

This output is the same as the output already generated by "SHOW HISTORY" so this PR also changes how the DDT output is displayed.

It is important to note that it appears no other VMs use the `sim_vm_fprint_stopped` feature, so changing its implementation to be more flexible should not cause any negative impact to other VMs now or in the future.

I have been using this feature for about a year (it was previously rejected prior to open-simh) and it has helped tremendously with software development under the AltairZ80 simulator. As the default behavior remains unchanged, this PR should have no impact on users that prefer the standard SCP stopped output.

It is also possible to add other custom stopped displays in addition to DDT-styled format.

### DIFF
--- a/AltairZ80/altairz80_cpu.c
+++ b/AltairZ80/altairz80_cpu.c
@@ -161,6 +161,7 @@ static t_stat cpu_set_chiptype      (UNIT *uptr, int32 value, CONST char *cptr, 
 static t_stat cpu_set_size          (UNIT *uptr, int32 value, CONST char *cptr, void *desc);
 static t_stat cpu_set_memory        (UNIT *uptr, int32 value, CONST char *cptr, void *desc);
 static t_stat cpu_set_hist          (UNIT *uptr, int32 val, CONST char *cptr, void *desc);
+static t_stat cpu_set_ddt           (UNIT *uptr, int32 value, CONST char *cptr, void *desc);
 static t_stat cpu_show_hist         (FILE *st, UNIT *uptr, int32 val, CONST void *desc);
 static t_stat cpu_clear_command     (UNIT *uptr, int32 value, CONST char *cptr, void *desc);
 static void cpu_clear(void);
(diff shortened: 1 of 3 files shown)


markpizz:
The primary reason that this was rejected in the past was due to the fact that all of the simulator logic running within the sim_instr() routine (and ANY devices provided by the simulator) has 100% control over absolutely any error code specifically returned by any of that code and thus the sim_vm_fprint_stopped routine (if provided by the simulator) gets to deal with producing appropriate simulator specific output that might be desired.

Dragging logic from SCP into simulator code, and changing SCP to meet the desires of a specific user with a specific simulator isn't the right approach.

> Adds the ability for a VM to define its own stopped message for all codes, including SCPE errors. For example, on the AltairZ80 simulator, a custom stop message handler could also display CPU flags and registers:

Non simulator specific codes (SCPE codes) are the domain of SCP and its facilities and by design, not part of a simulator's problem space to address.

If you want to insert logic into the return path of the simulator specific sim_instr() routine to emit additional thing on specific SCPE errors, you're free to do that.

As suggested when you first proposed this, you could define things in your running environment.  That discussion was here: https://github.com/simh/simh/pull/962.

In that discussion you and @rcornwell suggested calling the sim_vm_fprint_stopped routine for all stop states.  

In that prior discussion, both you and @rcornwell stated that no other simulator would be affected.  That is absolutely not true.    Prior simulators exist which already implement the sim_vm_fprint_stopped, and have done so for a long time.  This PR changes the flow of message reporting and any other simulator which depended on the prior behavior will now be changed.

deltecent:
> Prior simulators exist which already implement the sim_vm_fprint_stopped, and have done so for a long time.

You are correct. My first search didn't find anyone else using it. I did my search again and both HP2100 and HP3000 appear to use it.

markpizz:
After PR#123 is adopted, you can handle any return codes you're interested in before sim_instr() returns and (or |) SCPE_NOMESSAGE into the return code before returning from sim_instr().

markpizz:
Clearly, you only (|SCPE_NOMESSAGE) codes you've handled.

deltecent:
Brilliant! I will update this PR to implement that change should PR #123 get adopted.

deltecent:
Mark wins.

PR #127, opened 2022-11-23
PDP11: Add RP11 controller and RP02/03 disk drive support

al20878:
This PR implements the RP11 (RP11C) controller and RP02/03 disk pack drives (the DP: device in many old DEC OSes).

It allows the user to switch from the existing RP device (which is currently only the later MASSBUS RH70 controllers for RP04/RM03 and later drives), to the RP11C controller with the use of "SET RP RP11" command, and back with "SET RP MASSBUS", if desired.  Both cannot be used on the same system, anyways, because they share the same IO address range and the vector.

As a side-effect, makefile was amended to include the -D_DEBUG and -DNDEBUG compiler flags for builds with and without the debugging information, respectively, and also to get aligned with that Windows (MSVC) builds were already using.  That helps with transparent use of `assert()` which greatly helps both document and debug the code. So hopefully, this addition will be useful.

### DIFF
--- a/PDP11/pdp11_io_lib.c
+++ b/PDP11/pdp11_io_lib.c
@@ -673,6 +673,8 @@ AUTO_CON auto_tab[] = {/*c  #v  am vm  fxa   fxv */
         {010700} },                                     /* KG11-A - fx CSR, no VEC */
     { { "RHA" },         1,  1,  0, 0, 
         {016700}, {0254} },                             /* RH11/RH70 - fx CSR, fx VEC */
+    { { "RP" },          1,  1,  0, 0, 
+        {016710}, {0254} },                             /* RP11 - fx CSR, fx VEC */
     { { "RHB" },         1,  1,  0, 0, 
         {012440}, {0224} },                             /* RH11/RH70 - fx CSR, fx VEC */
     { { "RHC" },  1,  1,  0, 0, 
@@ -702,8 +704,6 @@ AUTO_CON auto_tab[] = {/*c  #v  am vm  fxa   fxv */
         {0200,     0170,   0174,   0270,   0274} },     /* LP11 - fx CSR, fx VEC */
     { { "RB" },          1,  1,  0, 0, 
         {015606}, {0250} },                             /* RB730 - fx CSR, fx VEC */
-    { { "RL" },          1,  1,  0, 0, 
-        {014400}, {0160} },                             /* RL11 - fx CSR, fx VEC */
     { { "RL" },          1,  1,  0, 0, 
         {014400}, {0160} },                             /* RL11 - fx CSR, fx VEC */
     { { "DCI" },         1,  2,  0, 8, 
--- a/PDP11/pdp11_rp.c
+++ b/PDP11/pdp11_rp.c
@@ -598,6 +598,10 @@ t_stat rp_set_bad (UNIT *uptr, int32 val, CONST char *cptr, void *desc);
 int32 rp_abort (void);
 t_stat rp_help (FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const char *cptr);
 const char *rp_description (DEVICE *dptr);
+#if defined(VM_PDP11)  &&  !defined(UC15)
+t_stat rp_set_ctrl (UNIT *uptr, int32 val, CONST char *cptr, void *desc);
+t_stat rp_show_ctrl (FILE *st, UNIT *uptr, int32 val, CONST void *desc);
+#endif
 
 
 /* RP data structures
@@ -655,6 +659,12 @@ REG rp_reg[] = {
     };
 
 MTAB rp_mod[] = {
+#if defined(VM_PDP11)  &&  !defined(UC15)
+    { MTAB_VDV, 0, NULL, "RP11C", 
+        &rp_set_ctrl, NULL, NULL, "Switch controller type to RP11C/RP03" },
+    { MTAB_VDV, 0, "TYPE", NULL,
+        NULL, &rp_show_ctrl, NULL, "Display controller type" },
+#endif
     { MTAB_XTD|MTAB_VDV, 0, "MASSBUS", NULL, 
         NULL, &mba_show_num, NULL, "Display Massbus number" },
     { MTAB_XTD|MTAB_VUN, 0, "write enabled", "WRITEENABLED", 
@@ -1525,3 +1535,37 @@ const char *rp_description (DEVICE *dptr)
 {
 return "RP04/05/06/07 RM02/03/05/80 Massbus disk controller";
 }
+
+#if defined(VM_PDP11)  &&  !defined(UC15)
+/* Show / Switch controller type */
+
+t_stat rp_show_ctrl (FILE *st, UNIT *uptr, int32 val, CONST void *desc)
+{
+    fputs ("MASSBUS", st);
+    return SCPE_OK;
+}
+
+t_stat set_dev_enbdis (DEVICE *dptr, UNIT *uptr, int32 flag, CONST char *cptr);
+
+t_stat rp_set_ctrl (UNIT *uptr, int32 val, CONST char *cptr, void *desc)
+{
+    DEVICE* dptr = find_dev_from_unit (uptr);
+    t_stat rv;
+    int i;
+
+    if (cptr)
+        return SCPE_ARG;
+    if ((rv = set_dev_enbdis(dptr, 0, 0/*disable*/, NULL)) != SCPE_OK)
+        return rv;
+    for (i = 0;  sim_devices[i];  ++i) {
+        extern DEVICE rr_dev;
+        if (sim_devices[i] == dptr) {
+            sim_devices[i]  = &rr_dev;
+            if ((rv = set_dev_enbdis(sim_devices[i], 0, 1/*enable*/, NULL)) != SCPE_OK)
+                sim_devices[i] = dptr;
+            return rv;
+        }
+    }
+    return SCPE_NXDEV;
+}
+#endif
--- a/PDP11/pdp11_rp11.c
+++ b/PDP11/pdp11_rp11.c
@@ -1,0 +1,1275 @@
+/* pdp11_rp11.c: RP11-C/RP02/RP03 disk pack device
+
+   Copyright (c) 2022 Tony Lawrence
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+   ROBERT M SUPNIK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+   IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+   Except as contained in this notice, the name of Robert M Supnik shall not be
+   used in advertising or otherwise to promote the sale, use or other dealings
+   in this Software without prior written authorization from Robert M Supnik.
+
+   Inspired by PDP-11's RK/RL implementations by Robert Supnik.
+*/
+
+#if defined(VM_PDP11)  &&  !defined(UC15)
+
+#include <assert.h>
+
+#include "pdp11_defs.h"
+#include "sim_disk.h"
+
+/* Constants */
+
+#define RPCONTR         uint16
+#define RPWRDSZ         16
+#define MAP_RDW(a,c,b)  (Map_ReadW (a, (c) << 1, b) >> 1)
+#define MAP_WRW(a,c,b)  (Map_WriteW(a, (c) << 1, b) >> 1)
+
+/* RP02 parameters; PR03 doubles # of cylinders (both total and spare) */
+#define RP_NUMWD        256                             /* words/sector */
+#define RP_NUMCY        203                             /* cylinders/drive */
+#define RP_SPARE        3                               /* of those, spare */
+#define RP_NUMSF        20                              /* surfaces/cylinder */
+#define RP_NUMSC        10                              /* sectors/track */
+#define RP_NUMTR        (RP_NUMCY * RP_NUMSF)           /* tracks/drive */
+#define RP_NUMBL        (RP_NUMTR * RP_NUMSC)           /* blocks/drive */
+#define RP_NUMDR        8                               /* drives/controller */
+#define RP_MAXFR        (1 << 16)                       /* max transfer */
+#define RP_SIZE(n)      (RP_NUMWD * (n))                /* words in n blocks */
+#define RP_SIZE_RP02    RP_SIZE(RP_NUMBL)               /* RP02 capacity, words */
+#define RP_SIZE_RP03    RP_SIZE(RP_NUMBL*2)             /* RP03 capacity, words */
+#define RP_ROT_12       125                             /* Half rotation, 0.1ms */
+
+/* Flags in the unit flags word */
+#define UNIT_NOAUTO     DKUF_NOAUTOSIZE                 /* autosize disabled */
+#define UNIT_V_RP03     (DKUF_V_UF + 0)
+#define UNIT_RP03       (1 << UNIT_V_RP03)              /* RP03 vs RP02 (0) */
+#define GET_DTYPE(x)    (((x) & UNIT_RP03) >> UNIT_V_RP03)
+
+/* Controller / drive types */
+#define RP_RP11C        "RP11C"
+#define RP_RP02         "RP02"
+#define RP_RP03         "RP03"
+
+/* Parameters in the unit descriptor */
+#define CYL             u3                              /* current cylinder */
+#define FUNC            u4                              /* function */
+#define HEAD            u5                              /* current track */
+#define SEEKING         u6                              /* unit performing a seek */
+
+/* 12 UNIBUS registers */
+#define RP_IOLN         030
+
+/* RP02/RP03 particulars */
+static struct drv_typ {
+    const char*         name;                           /* device type name */
+    int32               cyl;                            /* cylinders */
+    int32               size;                           /* #blocks */
+    int32               spare;                          /* spare (out of cyl) */
+    int32               seek_1;                         /* one track move, 0.1ms */
+    int32               seek_ave;                       /* average seek, 0.1ms */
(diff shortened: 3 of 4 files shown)


pkoning2:
I'm glad to see the RP11 support, but this is the wrong way.  An RP11 is totally unrelated to the RH11/RH70 massbus controller, and combining them is very wrong.
Instead, you should just make it a different device type.  You might use as a model the way the PDP10 emulation does it, with a device called "DP" -- I assume that was done because RP was taken for the massbus disks.  Check out PDP10/kx10_dp.c.  In fact, I'm wondering if it would make sense to combine the two -- use the PDP10 code, with some conditionals, for the PDP11.  That sort of thing is done in other cases: for example, the pdp10 emulation uses a number of pdp11 (Unibus) devices, as does the vax emulation.

al20878:
I haven't combined anything; they are completely separate instances.  Like I said, you can't use both: only one or the other.  So you can switch the simulator to use RP11C (and RP02/03) or MASSBUS (and then other, later drives).

pkoning2:
There are two problems.

One is that you use RP for both massbus and RP11 disks.  That simply is technically bogus.  The other is that, as you said, you can use only one or the other.  That is also wrong because a PDP11 is perfectly capable of having both an RP11 and an RH11 (with massbus disks) on it.  I'm fairly sure I've seen such setups at DEC.

al20878:
SimH is insisting on having RHA (the first MASSBUS controller) on the first MASSBUS address range (and vector), which is also RP11C's. So they can't co-exist already!

al20878:
> One is that you use RP for both massbus and RP11 disks.  That simply is technically bogus. 

I don't think so.  When you configure your simulator you actually tell SimH, which devices your PDP11 computer is using, so saying the you're using the "RP" device, which in fact is RP11C is nothing bogus.  You're not supposed to be switching controllers or anything like that while running simulated software -- it's done once at the configuration stage before you boot.

And again, it's not BOTH, it's either / or.  The logic in MASSBUS configuration (as it's currently coded -- I could not judge by anything else, but I assume that reflects the then-reality) won't allow you to have any MASSBUS devices without the first controller -- which shares the RP11's space.

al20878:
>I could not judge by anything else

Actually, all hardware handbooks that I could have at my disposal, showed that address range shared, in favor of later controllers, RJP04 first, then RM02/03 (MASSBUS already)...  That made me almost certain that they are mutually exclusive with RP11.

Besides, RP11 did not even exist in SimH until now, so what's the problem with it sharing the RP name?

LOL, when I was coding it, I was using RR (seemingly unused abbreviation, next to RP)...  But again, suppose that if you have RR in your SimH.  Once you activate it, your MASSBUSS is out of luck -- because RR will grab the RHA IO and vectors.

IDK how to better explain this.  Or maybe I don't understand what's your point in here.

rcornwell:
The reason that RP01/2 disks are called DP on PDP10 is that is what they are  called under TOPS10. RC10 disks are FH. RP04/5/6 are RP and RS drives are FS. For the PDP11 I believe RP04/5/6 drives where called DB. So perhaps whatever the RP01/2/3 disks are called on PDP11 systems would be a good name.

The correct way to do this is to name the drives something else, and set the drives to disabled by default. If the address conflicts.

As to merging the PDP10 device, I doubt it since the PDP10 ran 576 byte sectors. (512 * 8 on simh).

al20878:
As for combining PDP10 and PDP11 for DP -- that's impossible because on PDP10, there was a different controller RP10 used for  RP02/03 drives.  RP11 (for PDP11) and RP10 (for PDP10) are two completely incompatible species.

rcornwell:
Please do not share the names. This will cause no end of trouble for the users.

al20878:
RP11' s disks were called DP: on DOS / RSTS/e and RSX.  MASSBUS disks were DB: and DM: (RSTS an RSX), but DP: was not reused

al20878:
> This will cause no end of trouble for the users.

How?  Please explain.  Maybe you check out the code and try it?

rcornwell:
If the drives where called DP on the system, then I would call them dp devices. Note other devices then disks could be attached to massbus controller.

al20878:
> If the drives where called DP on the system, then I would call them dp devices. 

That was almost never the case, RK -> DK:, RL -> DL:, RP -> DP:, RH-> DB:, RM ->DM: etc etc

 > Note other devices then disks could be attached to massbus controller.

They couldn't be.  I tried to explain that above already.  If you have an RP11, your system cannot use RHA, the first MASSBUS controller, because they are in conflict.  If you don't have RHA, you can't have RHB and RHC, that's endgame for MASSBUS.

Like I said, it can be named "RR", I have no problems with that.  But once RR is up, the MASSBUS is down.  And vice versa.

rcornwell:
Couple issues:

 1) Why are you removing the RL device from the list?

 2) Please do not edit the Makefile unless you are adding a new file. You introduced a DEBUG flag as part of default build. Also if adding in a new file you also need to update the VS project files.

al20878:
> Why are you removing the RL device from the list?

I'm not.  There was a duplicate entry (copy and paste thing, not the first time).

> Please do not edit the Makefile unless you are adding a new file. You introduced a DEBUG flag as part of default build. Also if adding in a new file you also need to update the VS project files.

Please read my comments.  _DEBUG and NDEBUG are in the VS builds already, but were missing from the makefile.

al20878:
> Also if adding in a new file you also need to update the VS project files.

Correct.  I'm not using MSVC.  Maybe someone can do that for me?  (Please?)  It did pass all the checks, though.  Probably because it does not build MSVC projects... I would have noticed that if it did...

rcornwell:
If _DEBUG is not in makefile, please do not add it.

I would call it DP, and set the default to disabled. The RP11 and RH11 are very different controllers. Also it would be possible for a user to configure the RP11 or RH11 at a nonstandard address and use both. Also the RH11 could have tapes and other types of drives attached. The RP11 can only be used to talk to only disks.

al20878:
> If _DEBUG is not in makefile, please do not add it.

Why not, BTW?  It's in MSVC already

markpizz:
You're missing an addition of a DEVICE to the pdp11_sys.c.

You should pick a simh DEVICE name that is unique and does not overlap or otherwise collide with existing device names.

@rcornwell said:
> If _DEBUG is not in makefile, please do not add it.

Absolutely.  Invoke make with DEBUG=1 will do a debug build.

Do not use assert().  Use ASSURE() and don't worry about performance.  The device reset routine is executed a minimal number of times.

Clearly, since the MASSBUS RP device and RP11 can not coexist, you should let the autoconfigure logic detect the problem and the user should manually disable the conflicting device.

TM and TS are examples of this same problem.  Hmm... That may need some fixing when crossing UBA and MBA details.

You usually never get to add a new device without reserving/adding bits relating to pending interrupts.  It seems you're overloading the RP values.  Maybe that's OK since they can't coexist....

al20878:
> You're missing an addition of a DEVICE to the pdp11_sys.c.

I wasn't missing that.  It wasn't necessary.

al20878:
> Invoke make with DEBUG=1

This is what I did. For `assert()` to go away in non-debug builds, one needs NDEBUG.  Besides, both _DEBUG and NDEBUG are perfectly defined by MSVC, so what's the problem with having them, for exactly same purposes in the UNIX builds?  I don't follow much of a logic here.

asserts() are extremely helpful and self-documenting way to write a better code, IMO.  I used them not only in reset() but other places,  and obviously they are not necessary once the debugging is done (for a moment), and they don't need to be compiled.

markpizz:
Please explain.  The contents of a particular DEVICE structure are private to the code which it is directly associated with.  You're not messing with the existing rp_dev.

```
sim> SHOW DEVICE
```

Lists the configured devices.

```
sim> SET <devname>/<unitname> something 
```

changes something about the device.

The device that's being changed needs to be in the device list defined in pdp11_sys.c

al20878:
> Clearly, since the MASSBUS RP device and RP11 can not coexist, you should let the autoconfigure logic detect the problem and the user should manually disable the conflicting device.

And I did!  If you want to comment on what's done please take a look at the code first.

markpizz:
> asserts() are extremely helpful and self-documenting way to write a better code, IMO. I used them not only in reset() but other places, and obviously they are not necessary once the debugging is done (for a moment), and they don't need to be compiled.

ASSURE is functionally equivalent to assure().  Just use it!

al20878:
>  You're not messing with the existing rp_dev.

I wasn't messing with anything.  Both devices were completely separate, they were trading placed when the controllers were changed.  Again, please read the code.

pkoning2:
On the subject of coexistence, you're mixing up standard addresses with what is actually possible in the hardware.   It is indeed correct that the standard address for the first massbus moving head disk controller is the same as that for the RP11.  That doesn't mean you couldn't build a config with both; you just have to move one of them to another address.  An obvious answer would be to put the RH11 on the second moving head controller address, 776300.  In fact, RSTS expects that.  It knows how to tell the difference between the two controller types at 776700, and if it sees that the controller at that address  is an RP11 it checks for an RH11 at that second address.
SIMH doesn't seem to handle the flexibility of massbus devices; it only has three (there should be four) and it assumes that there can't be mixed devices.  For some OS, like RSTS, that's correct; for others, like RSX, it's more restrictive than necessary.  What SIMH should really do, and especially with RP11 support added, is to support an RH at 776300.  For conventional (not mixed, RSTS compatible) configs you'd be able to put RM02/3/5/80 on there, and with later versions of RSTS also RP04/5/6.  (And RP07 works though was officially unsupported.)
BTW, about device names, the mapping from SIMH names to OS names is somewhat mixed up anyway.  RSTS and several others use names starting with D for disks, but RT11 starts them with R.  SIMH has some oddballs like HK (RK06/7 which is DM on RSTS) and RQ (from Unix conventions?) which DEC tends to call DU (does RT call that RU?  After my time...)  So for RP11 you'd want a new name with a separate controller, not mixed with massbus, and the name is somewhat arbitrary so long as it's documented.  Given that PDP10 calls is DP that would be an obvious choice.

al20878:
I am going to redo the PR, thanks for all the input.  It's my first (sizeable) contribution to this project so please be patient with me.

I do understand your point of a possibility for combining the incombinables -- by manually tweaking IO and interrupt, and so in order for that to work, I will need to add a few more things to completely sever from the original MASSBUS implementation of RP.

Like I said, since I thought that the devices were not really coexisting in any feasible manner, I could just switch them, retaining the same device name and resources (that's the only things they had in common).

I'm okay with renaming the device to DP, and I'll take a look at `ASSURE()`.

Best

al20878:
BTW, whatever was added, was perfectly working with whatever software I could throw it to -- the DP: drives were fully functional and happy.  Anyways, I guess it just needs a little more "administrative" upheave now.

al20878:
I still don't understand, though, why MSVC builds use `_DEBUG` and `NDEBUG`, and UNIX builds can't...  That beats me, really

pkoning2:
Microsoft isn't much for standards...

BTW, the reason I objected so strongly to mixing the RP11 into the Massbus controller setup is that SIMH devices correspond to real devices.  And an RP11 is a very different controller than an RH11 or RH70, so it belongs in a different SIMH device emulation.

The one place where you find multiple controller types in a single emulation is if they are essentially compatible, so RH11 and RH70 are combined since softwarily they are pretty much the same.  And all MSCP devices are combined since they have a common programming interface.  But RK05 and RK06/7 are separate because all they have in common are "R" and "K".

markpizz:
In simh, the places you care to do asserts will essentially always be in code executed once or rarely (ATTACH, or RESET).  Performance doesn't matter.  Meanwhile, ASSURE() came in to being when someone had a case where their code only worked in debug mode. As it turns out, the assert argument had side effects which assert didn't execute in non-debug mode.
Device code can have ASSURE engagements essentially anywhere since there will never be a performance impact since device simulators are instruction bound most of the time and only rarely engage devices.

pkoning2:
I'm looking forward to your reworked version.  When it shows up I'll throw some tests at it -- see what RSTS thinks of it, and in particular give it an RP11 plus RH11 config.  I may also be able to build an old RSTS (V4A) which supports just four disk types: RC11, RF11, RK05, and RP11.

al20878:
> Microsoft isn't much for standards...

`NDEBUG` is actually the standard C.  Take a look at `man assert`.  And MSVC did the right thing there, actually.  `_DEBUG` is not so much, but IMO the conditional macros must be maintained consistently across the platforms, which is why I added both to makefile -- so the same code pathways get conditionalized uniformly.

al20878:
>  I may also be able to build an old RSTS (V4A) which supports just four disk types: RC11, RF11, RK05, and RP11.

You can do it already (with this version).  Works fine with DOS and V4A.

> give it an RP11 plus RH11 config

Not with a hybrid mix, though, at the moment...  But wait up, I'll be back shortly ;-)

al20878:
> As it turns out, the assert argument had side effects which assert didn't execute in non-debug mode.

That's a well-known fact about how to use `assert()`.  The beauty of it is that it is completely removed when you do no longer need / want it in the release build.  OTOH, as I see it, `ASSURE()` always executes, which is exactly what I want to avoid.

markpizz:
Sorry about your desire to avoid.  Just use ASSURE and please move on.

markpizz:
Source modules which include anything that is between <> will be rejected in a PR.  Simulation code (devices and cpus) are only to use SCP provided capabilities.  Anything provided by a platform must be done within one of the SCP modules.

al20878:
Then can SCP please provide `ASSERT()` or `_ASSERT()`, which is functionally equivalent to the standard thing?  Like I said, `ASSURE()` is not doing the right thing (not for me, at least).

markpizz:
How are you harmed by ASSURE()?

al20878:
> How are you harmed by ASSURE()?

It's executing dead code.

markpizz:
> > How are you harmed by ASSURE()?
>
> It's executing dead code.

Let the dead rest in peace!

How are the dead actually harming you?

al20878:
Dead code means worse performance and also affects compiler optimization.  Why do you use -O2?

markpizz:
Clearly you don't believe it doesn't matter, so when you can measure the difference in performance for your simulator when ASSURE() is used vs when assert() is used and the difference is actually greater than the sampling difference between different measurement times, we'll consider it.

pkoning2:
Are we talking about the same thing here?
The standard meaning of "dead code" is code that appears in the source but can't be reached.  That clearly cannot affect performance.  The compiler might well remove it -- modern compilers are good at recognizing dead code and making it go away.
The harm in dead code is that it clutters up the source and may make it harder to read -- which impairs maintainability.

On assert vs. ASSURE, it seems we're talking about a run-time correctness check that is performed only in debug builds, but which evaluates its condition (a) only in debug builds (b) in all builds.  If you're doing these checks in the core emulation loop, then the difference certainly matters, but I don't think people have typically stuck asserts or variations thereof in there.  So, if you limit their use to infrequently used functions, as Mark suggests, then using ASSURE means any side effects will happen for all builds.

Then again, there are two reasons I don't particularly care for ASSURE.  One is that it's non-standard, both in name and in behavior.  The other is that anyone who sticks side effects into function arguments, especially functions of this type, needs to have his head examined.

al20878:
> Clearly you don't believe it doesn't matter

Clearly, you eschewed my question:  Why do you use -O2?

al20878:
Dead code in this case, is that it can be reached but it does nothing.  In essence it's the same as if it wasn't there (such as that code that is never reached) -- for the end results of the algorithm.  It takes CPU cycles and burns fossil fuel LOL It slows down your application, it also breaks optimization.  I know, I repeat myself.  But `assert()` is actually used quite frequently in places, which are not just one-offs.  It's in fact encouraged for such places that check all your invariants for different pathways and conditions, the more the better.

markpizz:
> Clearly, you eschewed my question: Why do you use -O2?

-O2 will have absolutely no measurable effect on simh device code.  I will absolutely effect CPU instruction simulation code and that's why we do it.

@pkoning2 said:
> On assert vs. ASSURE, it seems we're talking about a run-time correctness check that is performed only in debug builds, but which evaluates its condition (a) only in debug builds (b) in all builds. If you're doing these checks in the core emulation loop, then the difference certainly matters, but I don't think people have typically stuck asserts or variations thereof in there. So, if you limit their use to infrequently used functions, as Mark suggests, then using ASSURE means any side effects will happen for all builds.
>
> Then again, there are two reasons I don't particularly care for ASSURE. One is that it's non-standard, both in name and in behavior. The other is that anyone who sticks side effects into function arguments, especially functions of this type, needs to have his head examined.

ASSURE exists to provide an assert() equivalence since calling C facilities not provided by the SCP libraries is not allowed.  The reliable side effect is a side benefit of ASSURE() but not the primary reason.

If we were driving towards allowing any C standard APIs to be used willy-nilly, but then all sorts of other things within SCP's capabilities won't work as expected.

@al20878 Please measure the difference so you can justify your concerns.

pkoning2:
I can't see any good reason for being so rigid about "no C library calls except indirectly via SCP".  I view SCP as an abstraction layer that provides additional features or higher level services, like its disk or tape or network machinery.  But a shim layer that merely renames a feature of the C standard strikes me as something that removes, not adds, value.

It may have been useful in the dark ages of early C before standards were regularly used, but ANSI C has been around for 40 years now and it seems reasonable to take advantage of that fact.

al20878:
> Sorry was at a movie.

Cool. Hope it was a good one

> In terms of NDEBUG and other modification to makefile. Keep in mind that any changes to compile options effect ALL simulators, not just PDP11

I keep beating a dead horse here:  it's already there by the virtue of MSVC.  For ALL simulators.  Also, if they don't use it, how exactly are they affected?

> ASSERT vs ASSURE, just a name.

It's not.  One is debug build only, the other one is forever.

> And any unreachable code will be removed by the compiler.

True.  But here it's quite reachable yet doing nothing and only causing the climate change (if I'd go with `ASSURE`).

> SCP provides a wrapper to the OS 

Correct.  But we're not talking about OS here, rather the standard C, which comes with the standard library, which is supposed to be behaviorally identical on any platform that provides C as a programming language.  That's very different from any OS particulars.

markpizz:
Please prove your very concerned point with measurements.

Your big concern about dead code in your case is trivial compared to ALL the dead code in most simulator devices which is dead almost all the time.  This is all of the use of sim_debug().  Guess what, the dead case (no debug being produced) has no measurable impact on simulator performance.

> > SCP provides a wrapper to the OS
> 
> Correct. But we're not talking about OS here, rather the standard C, which comes with the standard library, which is supposed to be behaviorally identical on any platform that provides C as a programming language. That's very different from any OS particulars.

Arbitrary use of C APIs can impact how SCP needs to behave correctly.  Passing ALL system things through SCP APIs that have been written specifically with the concern about solving both the needs of API users and the needs of SCP. 

You're repeating yourself, so I will too:
Please prove your very concerned point with measurements!
Please prove your very concerned point with measurements!

al20878:
I had a chat with Bob on the question of naming the RP11C device, because it turned out that DP was an actual thing back in the day, DP11.  It existed, but it isn't implemented in the simulator.  Just the same as RP11 wasn't.  So hogging the "DP" mnemonic for RP11 seems a little off.  OTOH, Bob said that there was a trend that RK06/RK07 were named as "HK" because "RK" was taken by RK11/RK05, so he suggested to use "HP" for me. Also noting that "HP" was used in Ultrix as a device for RP04+ devices (when RP11 was long gone).  So as far as I understand it, using HP for RP11 would be backwards and confusing...  What is your take on the name?  He also said that RR (my original "working" name) was fine with him.  RR was never used for anything so is safe in this respect but a bit baffling as to what that is, just by looking at the name.

pkoning2:
RR seems fine.  I have a vague memory of it being used some place for the second RP/RM massbus interface, but that's obscure enough and it may not even be accurate.

al20878:
> the second RP/RM massbus interface

That is called RHB in SimH, currently, IIUC. (And then RHC for the 3rd.)

> RR seems fine

Okay, great!

Issue #81, opened 2022-10-19
RK05 image created by SIMH will not attach properly - complains of non-existent parameter RK05


decuser:
- #### Context
The default pdp11 simulation no longer attaches RK images created by the simulator:

(base) nebula:open-simh-newest wsenn$ BIN/pdp11 
PDP-11 simulator V4.0-0 Current        git commit id: 65410851
sim> att rk0 rk0
%SIM-INFO: RK0: Creating new file: rk0
sim> q
Goodbye

(base) nebula:open-simh-newest wsenn$ BIN/pdp11 

PDP-11 simulator V4.0-0 Current        git commit id: 65410851
sim> att rk0 rk0
%SIM-ERROR: RK device: Non-existent parameter - RK05
sim> q
Goodbye

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
sim> show version
PDP-11 simulator V4.0-0 Current
    Simulator Framework Capabilities:
        32b data
        32b addresses
        Threaded Ethernet Packet transports:PCAP:VDE:NAT:UDP
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        RAW disk and CD/DVD ROM support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC Apple LLVM 13.1.6 (clang-1316.0.21.2.5)
        Simulator Compiled as C arch: x64 (Release Build) on Oct 19 2022 at 11:59:05
        Build Tool: simh-makefile
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: SDL Version 2.24.0, PNG Version 1.6.38, zlib: 1.2.12
        PCRE RegEx (Version 8.45 2021-06-15) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        Ethernet packet info: libpcap version 1.9.1
        OS: Darwin nebula.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:25:27 PDT 2022; root:xnu-8020.141.5~2/RELEASE_X86_64 x86_64
        Processor Name: Intel(R) Xeon(R) CPU           X5670  @ 2.93GHz
[...]

PR #129, opened 2022-11-23
PDP11:  Fixes for RK11

al20878:
1. Byte / word mismatch in memory access counters;
2. Device size (per comments) in dev_tab in blocks;
3. Undefined memory access for registers covered by the bus I/O range but otherwise unused in rk_wr();
4. Logic error in attach that would make type iterations (the drive has only one type, which is RK05, and is current, and cannot be changed.

Severities of the found bugs:

1: Quite unlikely to hit unless there is an out of bound
   memory access in the simulated program; then the number of
   words transferred would be misrepresented (sometimes by a lot);
2. The parameter is unused (but was inconsistent with the description) -- so it's purely cosmetic;
3. Quite unlikely to run into, but writing into all registers by any test software (e.g. XXDP) may cause SEGV in SimH (in the first statement of rk_wr() that fetches the old value);
4. Actually described in #81;  applying the fix will silence the error message.

### DIFF
--- a/PDP11/pdp11_rk.c
+++ b/PDP11/pdp11_rk.c
@@ -85,15 +85,15 @@
 
 #define RKCONTR         uint32                          /* container format */
 #define RKWRDSZ         18                              /* word width */
-#define MAP_RDW(a,b,c)  Map_Read18 (a, b, c)
-#define MAP_WRW(a,b,c)  Map_Write18 (a, b, c)
+#define MAP_RDW(a,c,b)  (Map_Read18 (a, (c) << 1, b) >> 1)
+#define MAP_WRW(a,c,b)  (Map_Write18 (a, (c) << 1, b) >> 1)
 
 #else
 
 #define RKCONTR         uint16
 #define RKWRDSZ         16
-#define MAP_RDW(a,b,c)  Map_ReadW (a, b, c)
-#define MAP_WRW(a,b,c)  Map_WriteW (a, b, c)
+#define MAP_RDW(a,c,b)  (Map_ReadW (a, (c) << 1, b) >> 1)
+#define MAP_WRW(a,c,b)  (Map_WriteW (a, (c) << 1, b) >> 1)
 
 #endif
 
@@ -104,9 +104,9 @@
 #define RK_NUMTR        (RK_NUMCY * RK_NUMSF)           /* tracks/drive */
 #define RK_NUMDR        8                               /* drives/controller */
 #define RK_M_NUMDR      07
-#define RK_SIZE         (RK_NUMCY * RK_NUMSF * RK_NUMSC * RK_NUMWD)
+#define RK_NUMBL        (RK_NUMTR * RK_NUMSC)
+#define RK_SIZE         (RK_NUMBL * RK_NUMWD)           /* words/drive */
 #define RK_RSRVSEC      (3 * RK_NUMSF * RK_NUMSC)       /* reserved (unused) disk area */
-                                                        /* words/drive */
 #define RK_CTLI         1                               /* controller int */
 #define RK_SCPI(x)      (2u << (x))                     /* drive int */
 #define RK_MAXFR        (1 << 16)                       /* max transfer */
@@ -121,7 +121,7 @@ struct drvtyp {
     };
 
 static struct drvtyp drv_tab[] = {
-    { RK_NUMSC, RK_NUMSF, RK_NUMCY, RK_SIZE, "RK05" },
+    { RK_NUMSC, RK_NUMSF, RK_NUMCY, RK_NUMBL, "RK05" },
     { 0 }
     };
 
@@ -293,7 +293,7 @@ BITFIELD *rk_reg_bits[] = {
     rk_ba_bits,
     rk_da_bits,
     NULL,
-    NULL,
+    NULL
     };
 
 /* Debug detail levels */
@@ -323,6 +323,7 @@ int32 last_drv = 0;                                     /* last r/w drive */
 int32 rk_stopioe = 1;                                   /* stop on error */
 int32 rk_swait = 10;                                    /* seek time */
 int32 rk_rwait = 10;                                    /* rotate time */
+static int32 not_impl = 0;                              /* placeholder for unused regs */
 
 const char *rk_regnames[] = {
     "RKDS",
@@ -332,7 +333,7 @@ const char *rk_regnames[] = {
     "RKBA",
     "RKDA",
     "unused",
-    "RKDB",
+    "RKDB"
     };
 
 int32 *rk_regs[] = {
@@ -342,6 +343,8 @@ int32 *rk_regs[] = {
     &rkwc,
     &rkba,
     &rkda,
+    &not_impl,
+    &not_impl
     };
 
 t_stat rk_rd (int32 *data, int32 PA, int32 access);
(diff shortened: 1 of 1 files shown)


markpizz:
The pdp11_rk.c module (as is much of the PDP11 simulator) started with Bob Supnik's (@rms47's) original work and problems you're changing relating to your number 3 would be his to review or rewrite.  

It would be best if you either:
1)  just reported the very details of the problem to Bob, or 
2) made those specific changes to the pdp11_rk.c code in the Supnik-Current branch and sent it directly to him which he may then take or adjust as he sees fit.

We will then pick up his changes when the next merge of contents come along from him.

al20878:
What about number 1?  It looks like it's an original issue as well then.  The UNIBUS access routines take the count of bytes to process, and return the count of _bytes_ unprocessed (due to a mapping error), but then the return value is used to adjust the number of _words_ that needed to be transferred to/from the drive.

Is @rms47 following this repo in here?

I can split this patch in two separate ones: the first for number 1 and 3, and then the second for 2 and 4, respectively.  But then again, just recently some cosmetic mods added to the RK code just in the master branch in here, so I am a bit confused with your directions.

markpizz:
Oops.  The same with number 1

rms47:
The only actual bug here is the size of the WC adjustment following a memory overrun, The old code is:
```
        else {                                          /* normal store */
            if ((t = MAP_WRW (ma, wc << 1, rkxb))) {    /* store buf */
                rker = rker | RKER_NXM;                 /* NXM? set flag */
                wc = wc - t;                         /* adj wd cnt */
                }
            }
```
The adjustment is incorrect because the Map_X routines handle everything - addresses, counts, etc - as BYTES, while the RK11 counts in WORDS. So the fix (in both the read and write cases) is simply
```
                wc = wc - (t >> 1);                     /* adj wd cnt */
```
None of the other changes, such as screwing around with the macro definitions, is needed. As the module maintainer, I strongly prefer to put adjustments out in the open, rather than buried in macros.

The addition of
```
        uptr->CYL = track / RK_NUMSF;
```
is okay but not really necessary; it is intended for the next seek time calibration, which might be off by a cylinder or two in the current code.

All the rest is because of issues introduced in V4.

#2. I have said this again and again. The RK05 has 203 cylinders, not 200. SimH is a hardware simulator, not a software simulator. The reserving of 3 cylinders for "software" was a convention, and it was not honored by all software. The metadata junk just makes it worse.
#3. Undefined memory access. V3 does not have this old value folderol.
#4. Autosizing. Autosizing should not be applied to an RK05, as it has only one size.

In short, V3 needs just the two changes for adjusting the word count when an NXM or map error occurs. I'll add the cylinder update too; it's harmless. As for the rest, I would prefer to see Open-SimH rip out V4 errors than continue to paper over them.

al20878:
>  Map_X routines handle everything - addresses, counts, etc - as BYTES

Of course, and that's how I noticed the discrepancy.

> None of the other changes, such as screwing around with the macro definitions, is needed. As the module maintainer, I strongly prefer to put adjustments out in the open, rather than buried in macros.

I did not know the intention here, but I assumed that since the macros were called RDW and WRW, with the ending "W" standing for "WORDS", and since the following WC adjustment was done in the whole word decrements -- that they were the macros themselves to perform the word-to-byte count conversion and back -- and so I suggested the fix for those as such.

Also, fiddling with the macros' parameters, "a" was for "address", "c" was for "count", and "b" was for "blocks"; so I swapped the names for the last two, to make them a bit clearer (or so I thought).

Just to explain how it all came to be.

> it is intended for the next seek time calibration, which might be off by a cylinder or two in the current code.

True, and I noticed that was missing.  As for "one or two", I'd say statistically it might be half the number of cylinders (so a hundred, give or take).  It's not at all critical in a simulator, though, either way.

> # 2

There, I was just reconciling the structure field as documented in comments ("#blocks") with the value -- and like I said, that value is not used throughout the module -- so I replaced that SIZE in words with the _equivalent_ size in blocks.

As far as I understand it, the problem with V4 is that if the full size is indicated (elsewhere in code), the simulator would give you a 2494464-byte container right away when you create a new disk, even though the system might actually use only the "usual" 2457600.  With the "reduced" size, that container would be created of the smaller size, but if the system is actually using more (all 203 cyls) it will be allowed to extend.  The pre-allocation is the culprit but it seems to be a "feature" of the new and convoluted "autosize" logic with "footers" -- and the latter is the worst addition to the simulator.

> # 3  V3 does not have this old value folderol.

Indeed.  Then I guess it was added in V4 for the device register debugging / examine / deposit aid, and also re-used in the bus read / write routines.

> # 4 an RK05, as it has only one size.

That's what the fix is all about.

rms47:
While I understand the intent of your work, I prefer to make conservative changes, i.e., to the fewest lines of code possible. Accordingly, I will not approve this change as currently written. Please implement the bug fix for #1 as I suggested, changing just the two lines of code. You may do with V4-related fixes as you please.

The fix to #1, and the corresponding fix to the RL11 (if it proves to be wrong as well), will be part of 3.12-3.

al20878:
Yes, there are same issues in the RL11 code (both V3-related and V4-specific)

al20878:
On a more general note, though:

since most of the devices allow to be re-assigned to random bus addresses, is it still the correct logic to do the `PA<2:1>` things, just like that when determining the register number:
```
(PA >> 1) & 07
```
Because to me it looks like, the proper way of figuring out the number would be:
```
((PA - ba) >> 1) & 07
```
where `ba` is the assigned base address (auto-configured or otherwise).  When I was working on RP11, whose `ba` is not exactly the "round number" but begins with 010 (776710), the bit extraction like the original would not have worked, so it had to be offset with the base address first.

pkoning2:
On the register number thing, the existing code is correct because the addressing rules require the base address of the controller to be a multiple of the address block size.  So the RK11 has to be at a multiple of 020, the KL11/DL11 at a multiple of 010, and so on.  
If SIMH allows assigning addresses that violate those rules, that would be a problem.  Unless/until that is enforced the answer is "don't do that".

al20878:
> Unless/until that is enforced the answer is "don't do that".

I mentioned RP11 for a reason.  It's neither 8 nor 16, it's 12 words:  17776710 - 17776736.  Bit-fiddling with shifts and masks can't isolate not-a-power-of-two, 030 (=24.) addresses.  OTOH, I read somewhere that RP11 actually responds to a wider range, beginning at 17776700 (so actually covers 16 words), just without any registers in the first 4 words.  But when configuring SimH, should it be the actual 710 or 700?  Doc says 710 (and I was using just that, and it worked with DOS/Batch).  RSTS/E shows 700 for the base (in connection with RP02/03) but accesses at 710 and up (as it should).  If the blocks are always enforced for the other controllers, then I agree there's no need to adjust (and I wasn't going to) -- I was merely asking that because of the RP11 situation.  It's really not so cool to keep submitting and re-submitting the patches, so I want to make sure, I won't to have to re-do it.

pkoning2:
The rule is simple: 
1. Find the size of the address range used by the controller.
2. Round that up to a power of 2.
3. The number you have now is the required alignment of the controller base address
So, for example, the RP11 with 12 words of CSRs requires alignment to 16 words (040).

al20878:
Thanks, although that'd work well for finding the base address if it was for the 12 first words; with RP11, they are the last 12 words of the 16-word interval.  With the power-of-two rule in place, it'd then be impossible to set up something at the address xxx710, because it's not divisible by any power of 2.  Should then RP11 use the 4 ghost words, so that it'd be configured at the 16-word boundary, per the rule?  Meaning, it'd be set up at 776700 (rather than 776710) by default.  Like I said, when I configured the default to be 776710 with 12 words IO length, it was perfectly fine with SimH auto-configurator.  So it's in the conflict with "the rule" itself.

pkoning2:
I didn't realize the RP11 is so strange.  It's not clear that particular controller can actually be moved, but if you want to allow that I would handle it as follows:
It has a block of 12 registers so it uses a 16-word block aligned on an 040 boundary.  The 12 actual registers occupy the last 12 slots in the 16-word range. so if you want to do the `(address >> 1) & 017` thing you'd number the registers 4 through 15, not 0 to 11.
I don't know why using 776710 and 12 words would "work" with the SIMH auto-configurator, since that's not a valid case in the PDP11 addressing rules.

al20878:
Yep, that's exactly what I as asking about.  So it'd be 776700 for the base, with 4 ghost registers at the beginning of the 16-word block.  Also, that's how RSTS/E seems to handle this controller's IO range.  So I'm going to keep this in mind when re-submitting my RP11/RP02/03 addition.  Thanks

rms47:
The RP11 is far afield from this issue of the RK, but it might as well get settled. The default base address of the RP11, according to the schematics, is 776700, and the address base must be a multiple of 40(8) bytes. A pair of 3-to-8 decoders is used for address selection. One decoder does 776720..36 and is fully populated. The other does 776700..16 and has only the 10,12,14,16 outputs populated. See the address selection logic of the RP11-C schematic, page 20.

Please take any further discussion of implementation issues for new devices to the SimH mailing list.

al20878:
> One decoder does 776720..36 and is fully populated. The other does 776700..16 and has only the 10,12,14,16 outputs populated.

Thanks.  This just confirms that I have said about the 776700 base.

I'm going to close this PR now, and rework it: split in two adding RL11 to the first (V3-related) part, and prepare the V4 part separately.

Issue #130, opened 2022-11-24
SCP: SET ON in init scripts not working as described


deltecent:
> The values following the word "ON" are specific error codes (SCPE_xxxx) text or numeric values for xxxx. If your driving script otherwise doesn't deal with error trapping, then the general word ERROR covers everything:
> ```
> sim> # enable error trapping:
> sim> SET ON 
> sim> # trap all non 0 values
> sim> ON ERROR Echo %STATUS%-%TSTATUS%
> sim> # specifically for your extended command passing status in case it is interesting...
> sim> ON ERROR <your-extended-command> %STATUS%
> ```
> You wouldn't have to type the above except once into an init file: ~/simh.ini

I have a ~/simh.ini file that contains:

```
set on
on error reg
```

When I run `altairz80` I get the following:

```
% altairz80 -v

Altair 8800 (Z80) simulator V4.0-0 Current        git commit id: a06ab205
/Users/patrick/simh.ini-2> set on
/Users/patrick/simh.ini-3> on error reg
sim> show on
On Processing for input commands is disabled

sim>
```

I have to run altairz80 with the -o switch in order for my ON settings in ~/simh.ini to remain active.

```
% altairz80 -v -o

Altair 8800 (Z80) simulator V4.0-0 Current        git commit id: a06ab205
/Users/patrick/simh.ini-2> set on
/Users/patrick/simh.ini-3> on error reg
sim> show on
On Processing for input commands is enabled
    on ERROR     reg

on state and actions are inherited by nested do commands and subroutines
sim>
```

Is ~/simh.ini supposed to require inheritance with -o?

I have the same issue when using ./altairz80.ini

markpizz:
Oops.  My advice was a little off.

On actions default to only affect activities within a running command file since many cases ALL commands run in that command file and a user never sees a sim> prompt.

For your interactive use in either file ~/simh.ini or ./altair.ini file you should be able to:
```
set on inherit
on error reg
```
Or, as you've otherwise discovered, invoking the simulator with -O enables INHERITance globally.

I'll update the HELP information to describe the INHERIT and NOINHERIT options on the SET ON command.

deltecent:
I suspected that's what was going on, but I wanted to make sure. Thanks.

deltecent:
[...]

PR #131, opened 2022-11-24
SCP: Fix for HELP SET ON to describe INHERIT scoping

markpizz:
Issue #130 identified the missing info.

Very slow/busy hosts can let pdp9 diagnostic test fail.  Update max instruction count.


markpizz:
I removed the commit changing the pdp9_test.ini limit and fixed the missing description of SET ON INHERIT|NOINHERIT  See #130 for discussion

PR #132, opened 2022-11-27
SCP: Add VM-specific messages to %TSTATUS% expansion

deltecent:
Clean up variable expansion help

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -1690,8 +1690,8 @@ static const char simh_help2[] =
       " %%CTIME%%, %%DATE_YYYY%%, %%DATE_YY%%, %%DATE_YC%%, %%DATE_MM%%, %%DATE_MMM%%,\n"
       " %%DATE_MONTH%%, %%DATE_DD%%, %%DATE_D%%, %%DATE_WYYYY%%, %%DATE_WW%%,\n"
       " %%TIME_HH%%, %%TIME_MM%%, %%TIME_SS%%, %%TIME_MSEC%%, %%STATUS%%, %%TSTATUS%%,\n"
-      " %%SIM_VERIFY%%, %%SIM_QUIET%%, %%SIM_MESSAGE%% %%SIM_MESSAGE%%\n"
-      " %%SIM_NAME%%, %%SIM_BIN_NAME%%, %%SIM_BIN_PATH%%m %%SIM_OSTYPE%%\n\n"
+      " %%SIM_VERIFY%%, %%SIM_QUIET%%, %%SIM_MESSAGE%%, %%SIM_NAME%%, %%SIM_BIN_NAME%%,\n"
+      " %%SIM_BIN_PATH%%, %%SIM_OSTYPE%%\n\n"
       "+Token %%0 expands to the command file name.\n"
       "+Token %%n (n being a single digit) expands to the n'th argument\n"
       "+Token %%* expands to the whole set of arguments (%%1 ... %%9)\n\n"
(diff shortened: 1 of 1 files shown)


markpizz (changes requested):
`The change to the help text is OK.

The change in fprint_stopped_gen DOES NOT BELONG

The change in sim_error_text() is OK, but no need for a blank line before:
```
sprintf(msgbuf, "Unknown %s simulator stop code %d", sim_name, stat);```

**Comment by markpizz** [comment]
Actually after more thinking, the change to sim_error_text isn't OK either.  The reason being that it is called from many places with the specific purpose of returning the specific text related to SCPE_xxx codes.

Extending %TSTATUS% to include the simulator text is reasonable, but the logic to extend the expanded result belongs where TSTATUS is expanded in _sim_get_env_special.

**Comment by deltecent** [comment]
Maybe in the future I should post the intent as an issue and then implement after you've had a chance to look at the intent and recommend the proper course of action.

**Comment by deltecent** [comment]
@markpizz see if that is more of what you had in mind.

**Review by markpizz — COMMENTED** [review]
The:
```
int32 stat = SCPE_BARE_STATUS(sim_last_cmd_stat);
```
line should be:
```
t_stat stat = SCPE_BARE_STATUS(sim_last_cmd_stat);
```
AND it should be located right above this line:
```
if (stat >= 1 && stat < SCPE_BASE && sim_stop_messages[stat])
```

**Comment by deltecent** [comment]
Maybe add "#define SCPE_SIM_BASE 1" to scp_defs.h.

**Comment by markpizz** [comment]
Hmmm... 2 things:
1) to avoid adding another #define you could use stat > SCPE_OK
2) parenthesis are needed around the two separate terms of the && expression:

```
if ((stat > SCPE_OK) && (stat < SCPE_BASE && sim_stop_messages[stat]))
```

**Comment by markpizz** [comment]
Correcting again:
```
if ((stat > SCPE_OK) && (stat < SCPE_BASE) && (sim_stop_messages[stat] != NULL))
```

**Comment by markpizz** [comment]
Upon further review, the parentheses aren't really necessary, but some folks, like me, are too weak minded to memorize the complete operator precedence rules so the parentheses makes things unambiguous.

**Comment by deltecent** [comment]
> to avoid adding another #define you could use stat > SCPE_OK

I was attempting to emphasize that simulator error codes start at 1.

> Upon further review, the parentheses aren't really necessary, but some folks, like me, are too weak minded to memorize the complete operator precedence rules so the parentheses makes things unambiguous.

I agree with the parenthesis. I didn't include "!= NULL" because other parts of the code don't, so I figured that was the desired style. I prefer having it there as it's more clear to me.

**Comment by deltecent** [comment]
> for (i=1; i<SCPE_BASE; ++i)

Would this be good place for a new definition?

**Comment by markpizz** [comment]
> > to avoid adding another #define you could use stat > SCPE_OK
>
> I was attempting to emphasize that simulator error codes start at 1.

Having the #define for SIM_SIM_BASE, in theory would make things more clear if it happened to be staring us in the face where we need to reference it, but as you say, that would be in sim_defs.h.  Referring directly to SCPE_OK and SCPE_BASE in this context seems to be particularly specific.

You ask:
> > for (i=1; i<SCPE_BASE; ++i)
> 
> Would this be good place for a new definition?

I'm not sure what you're asking.  The author(s) of a particular simulator can decide about how the stop codes should be assigned.  They don't need to be contiguous, but they should (but aren't required to) have a related text string at the appropriate offset in the sim_ stop_messages array.

**Comment by deltecent** [comment]
This is from show_on in scp.c:

```
    fprintf(st, "On Processing");
fprintf(st, " is %s\n", (sim_on_check[lvl]) ? "enabled" : "disabled");
for (i=1; i<SCPE_BASE; ++i) {
    if (sim_on_actions[lvl][i])
        fprintf(st, "    on %6d    %s\n", i, sim_on_actions[lvl][i]); }
```

Just asking while I'm in there.

**Comment by markpizz** [comment]
The code in question is displaying any specific ON actions which may have been specified using numeric codes specific to the current simulator.  The next for loop runs through the SCPE specific codes >= SCPE_BASE and displays any that have been explicitly specified using the text form of the ON action.

As for your question, the code that you're pointing at very specifically shows that 1 is the beginning of the range. while using a symbol would again not be staring you in the face and might create a need to look up the symbol...

The section of sim_defs.h which you showed earlier that describes specifically:

> sim_defs.h says:
>
```
/* Simulator status codes

   0                    ok
   1 - (SCPE_BASE - 1)  simulator specific
   SCPE_BASE - n        general
*/
```

Makes things clear enough for anyone writing or using a simulator.

The loop in question could, in theory, be rewritten as:
```
for (i=SCPE_OK+1; i<SCPE_BASE; ++i) {
```
But I don't see that as adding to anyone's understanding...  If you want to extend the language in the HELP output to describe simulator specific codes as starting at 1, that might be useful..

Besides looking at the output of SHOW ON, you might find it interesting and provide a better understanding of all that is going on if you also look at SHOW DO.

markpizz:
This change looks good.

PR #133, opened 2022-11-27
AltairZ80: Fix CPU_HALT stop code conflict with SCPE_OK and "ON 0"

deltecent:
AltairZ80 uses 0 for the VM-specific HLT instruction stop code.
SCP uses 0 to mean all errors ("ON ERROR").

The command "ON 0" will generate "%SIM-ERROR: Invalid argument: 0".

This PR changes the HALT stop code from 0 to 5.

### DIFF
--- a/AltairZ80/altairz80_defs.h
+++ b/AltairZ80/altairz80_defs.h
@@ -64,11 +64,11 @@ typedef enum {
 } ChipType;
 
 /* simulator stop codes */
-#define STOP_HALT       0   /* HALT                                             */
 #define STOP_IBKPT      1   /* breakpoint   (program counter)                   */
 #define STOP_MEM        2   /* breakpoint   (memory access)                     */
 #define STOP_INSTR      3   /* breakpoint   (instruction access)                */
 #define STOP_OPCODE     4   /* invalid operation encountered (8080, Z80, 8086)  */
+#define STOP_HALT       5   /* HALT                                             */
 
 #define UNIT_CPU_V_OPSTOP       (UNIT_V_UF+0)               /* stop on invalid operation                    */
 #define UNIT_CPU_OPSTOP         (1 << UNIT_CPU_V_OPSTOP)
--- a/AltairZ80/altairz80_sys.c
+++ b/AltairZ80/altairz80_sys.c
@@ -160,11 +160,12 @@ DEVICE      *sim_devices[]  = {
 static char memoryAccessMessage[256];
 static char instructionMessage[256];
 const char *sim_stop_messages[SCPE_BASE] = {
-    "HALT instruction",
+    "Unknown error",            /* 0 is reserved/unknown */
     "Breakpoint",
     memoryAccessMessage,
     instructionMessage,
-    "Invalid Opcode"
+    "Invalid Opcode",
+    "HALT instruction"
 };
 
 static const char *const Mnemonics8080[] = {
--- a/scp.c
+++ b/scp.c
@@ -9435,7 +9435,7 @@ t_value pcval;
 
 fputc ('\n', st);                                       /* start on a new line */
 
-if (v >= SCPE_BASE)                                     /* SCP error? */
+if (v == SCPE_OK || v >= SCPE_BASE)                                     /* SCP error? */
     fputs (sim_error_text (v), st);                     /* print it from the SCP list */
 else {                                                  /* VM error */
     if (sim_stop_messages [v])


markpizz:
This makes sense.  AltairZ80 seems to be the only simulator that had a 0 value for a potential simulator specific stop code.

The simulator stop codes overlap the SCPE_ value ranges, where everything from SCPE_BASE thru SCPE_MAX_ERR and SCPE_OK are reserved to SCP and from 1 thru SCPE_BASE-1 is reserved to simulators.

In the ON processing SCP leverages the fact that ON OK (or ON 0) is NOT a meaningful trap condition since you shouldn't trap on every command return, so ON 0 is reserved to cover any case where no specific trap was defined.

psco:
I am not a fan of "specification by voting". In particular the current approach (0 is reserved) leads to the, in my opinion, bad outcome that each and every simulator still must define a value in `sim_stop_messages` for 0 which is never used. The values range from "Unknown error" (most popular choice), "OK" (!) to "---".

In case 0 becomes officially reserved, the change is still incorrect:
- There is no need for `#define STOP_ERROR      0`
- The message `sim_stop_messages[0]` must be "Unknown error" (majority vote) and not "Reserved"

deltecent:
I'm good with "Unknown error". I had not looked at the other VMs. The intent was that if this PR was accepted, that future contributors would know that 0 was reserved and off limits. I'll make your recommended changes.

deltecent:
sim_defs.h says:

```
/* Simulator status codes

   0                    ok
   1 - (SCPE_BASE - 1)  simulator specific
   SCPE_BASE - n        general
*/
```

sim_error_text(t_stat, stat) returns "No Error" when stat == SCPE_OK. It seems that it shouldn't matter what a simulator puts in sim_stop_messages[0] as it should never be used.

```
const char *sim_error_text (t_stat stat)
{
static char msgbuf[64];

stat &= ~(SCPE_KFLAG|SCPE_BREAK|SCPE_NOMESSAGE);        /* remove any flags */
if (stat == SCPE_OK)
    return "No Error";
```

The problem is with fprint_stopped:

```
if (v >= SCPE_BASE)                                     /* SCP error? */
    fputs (sim_error_text (v), st);                     /* print it from the SCP list */
else {                                                  /* VM error */
    if (sim_stop_messages [v])
        fputs (sim_stop_messages [v], st);              /* print the VM-specific message */
    else
        fprintf (st, "Unknown %s simulator stop code %d", sim_name, v);
    if ((sim_vm_fprint_stopped != NULL) &&              /* if a VM-specific stop handler is defined */
        (!sim_vm_fprint_stopped (st, v)))               /*   call it; if it returned FALSE, */
        return;                                         /*     we're done */
    }
```

sim_error_text() handles stat == SCPE_OK while fprint_stopped() punts stat == SCPE_OK to the simulator. To be consistent with sim_error_text, fprint_stopped should probably be:

```
if (v == SCPE_OK || v >= SCPE_BASE)          /* SCP error? */
    fputs (sim_error_text (v), st);                     /* print it from the SCP list */
```
and AltairZ80 should not use 0 for STOP_HALT.

markpizz:
> I am not a fan of "specification by voting". In particular the current approach (0 is reserved) leads to the, in my opinion, bad outcome that each and every simulator still must define a value in sim_stop_messages for 0 which is never used. The values range from "Unknown error" (most popular choice), "OK" (!) to "---".
> 
> In case 0 becomes officially reserved, the change is still incorrect:

As @deltecent pointed out, it specifically reserved in sim_defs.h and has been there since 2001.

> There is no need for #define STOP_ERROR 0

Agreed.  Nothing in @deltecent 's change added such a define.

No simulator should return from sim_instr() with the value 0.  It should either be a simulator specific value OR a SCPE defined one which may have been returned by an event dispatch routine.

> The message sim_stop_messages[0] must be "Unknown error" (majority vote) and not "Reserved"

sim_stop_messages[0] can be anything you want (even NULL) it should never be referenced by anything.

At one point, the sim_stop_messages array was simply defined as:
```
const char *sim_stop_messages[] = {
    "HALT instruction",
    "Breakpoint",
    memoryAccessMessage,
    instructionMessage,
    "Invalid Opcode"
};
```

and now it is defined as:

```
const char *sim_stop_messages[SCPE_BASE] = {
    "HALT instruction",
    "Breakpoint",
    memoryAccessMessage,
    instructionMessage,
    "Invalid Opcode"
};
```

This was needed to properly allow the SCP code which references the array (and assumes it to be SCPE_BASE entries long) to not reach beyond the original array's bounds when lookup up the text for a sim_instr() returned value.  Someone in some simulator added a new simulator specific STOP_xxx value without adding an entry in the sim_stop_messages array and when they returned that value SCP reached beyond the original bounds of the array and picked up garbage data as a string pointer...  Dealing with this illuminated the specific need for the SCPE_BASE array size.  All existing simulators had the array size added back in September of 2020.

Meanwhile, the HALT stop code for the AltairZ80 simulator should be changed to something other than 0 and the appropriate element of the sim_stop_messages array should say "HALT instruction".

markpizz:
>
@deltecent said:
> sim_error_text() handles stat == SCPE_OK while fprint_stopped() punts stat == SCPE_OK to the simulator. To be consistent with sim_error_text, fprint_stopped should probably be:
```
if (v == SCPE_OK || v >= SCPE_BASE)          /* SCP error? */
    fputs (sim_error_text (v), st);                     /* print it from the SCP list */
```
>
> and AltairZ80 should not use 0 for STOP_HALT.

Agreed.

psco:
@markpizz The `#define STOP_ERROR 0` was in the version of the PR I commented on. @deltecent removed it at my request and now its gone.

markpizz:
> @markpizz The #define STOP_ERROR 0 was in the version of the PR I commented on. @deltecent removed it at my request and now its gone.

Sorry, that's a side effect of forced pushes.  Since forced pushes are preferred (just to keep the whole discussion in one place), then if such a thing (crossed messages when context changed), it would be fine with me to just ignore the currently irrelevant comments.  The rest of my points stand and I hope you agree with the reasoning.

psco:
The AltairZ80 related change is ok - thanks Patrick (@deltecent).

markpizz:
This PR is fine.

markpizz:
Parentheses wouldn't hurt in the scp change as well....

PR #135, opened 2022-11-27
PDP9 VAX SEL32: Adjust test limits

pkoning2:
Adjust the instruction limits for three simulators so they pass on Apple M1 systems.


markpizz:
The other day when I had a suggested pending change to the pdp9's test, I had upped the limit to 40M....  As you suggested there really isn't a penalty to allow extra limit time, while a headache when it is too small.

pkoning2:
Makes sense.  Done.

PR #143, opened 2022-12-03
PDP11: Fixes for RK11

al20878:
This is a redo of PR#129 with byte/word mismatch separared as PR#140 suitable for V3.

1. Device size (per comments) in dev_tab in blocks;
2. Undefined memory access for registers covered by the bus I/O range but otherwise unused in rk_wr();
3. Logic error in attach that would make type iterations (the drive has only one type, which is RK05, and is current, and cannot be changed.

### DIFF
--- a/PDP11/pdp11_rk.c
+++ b/PDP11/pdp11_rk.c
@@ -104,9 +104,9 @@
 #define RK_NUMTR        (RK_NUMCY * RK_NUMSF)           /* tracks/drive */
 #define RK_NUMDR        8                               /* drives/controller */
 #define RK_M_NUMDR      07
-#define RK_SIZE         (RK_NUMCY * RK_NUMSF * RK_NUMSC * RK_NUMWD)
+#define RK_NUMBL        (RK_NUMTR * RK_NUMSC)
+#define RK_SIZE         (RK_NUMBL * RK_NUMWD)           /* words/drive */
 #define RK_RSRVSEC      (3 * RK_NUMSF * RK_NUMSC)       /* reserved (unused) disk area */
-                                                        /* words/drive */
 #define RK_CTLI         1                               /* controller int */
 #define RK_SCPI(x)      (2u << (x))                     /* drive int */
 #define RK_MAXFR        (1 << 16)                       /* max transfer */
@@ -121,7 +121,7 @@ struct drvtyp {
     };
 
 static struct drvtyp drv_tab[] = {
-    { RK_NUMSC, RK_NUMSF, RK_NUMCY, RK_SIZE, "RK05" },
+    { RK_NUMSC, RK_NUMSF, RK_NUMCY, RK_NUMBL, "RK05" },
     { 0 }
     };
 
@@ -293,7 +293,7 @@ BITFIELD *rk_reg_bits[] = {
     rk_ba_bits,
     rk_da_bits,
     NULL,
-    NULL,
+    NULL
     };
 
 /* Debug detail levels */
@@ -323,6 +323,7 @@ int32 last_drv = 0;                                     /* last r/w drive */
 int32 rk_stopioe = 1;                                   /* stop on error */
 int32 rk_swait = 10;                                    /* seek time */
 int32 rk_rwait = 10;                                    /* rotate time */
+static int32 not_impl = 0;                              /* placeholder for unused regs */
 
 const char *rk_regnames[] = {
     "RKDS",
@@ -332,7 +333,7 @@ const char *rk_regnames[] = {
     "RKBA",
     "RKDA",
     "unused",
-    "RKDB",
+    "RKDB"
     };
 
 int32 *rk_regs[] = {
@@ -342,6 +343,8 @@ int32 *rk_regs[] = {
     &rkwc,
     &rkba,
     &rkda,
+    &not_impl,
+    &not_impl
     };
 
 t_stat rk_rd (int32 *data, int32 PA, int32 access);
@@ -773,7 +776,7 @@ if (wc && (err == 0)) {                                 /* seek ok? */
             awc = (wc + (RK_NUMWD - 1)) & ~(RK_NUMWD - 1); /* clr to */
             for (i = wc; i < awc; i++)                  /* end of blk */
                 rkxb[i] = 0;
-            sim_disk_data_trace (uptr, (uint8 *)rkxb, da/RK_NUMWD, awc, "sim_disk_wrsect", RKDEB_DAT & dptr->dctrl, RKDEB_OPS);
+            sim_disk_data_trace (uptr, (uint8 *)rkxb, da/RK_NUMWD, awc*sizeof(*rkxb), "sim_disk_wrsect", RKDEB_DAT & dptr->dctrl, RKDEB_OPS);
             err = sim_disk_wrsect (uptr, da/RK_NUMWD, (uint8 *)rkxb, NULL, awc/RK_NUMWD);
             }
         break;                                          /* end write */
@@ -816,6 +819,7 @@ if ((uptr->FUNC == RKCS_READ) && (rkcs & RKCS_FMT))     /* read format? */
 else da = da + wc + (RK_NUMWD - 1);                     /* count by words */
 track = (da / RK_NUMWD) / RK_NUMSC;
 sect = (da / RK_NUMWD) % RK_NUMSC;
+uptr->CYL = track / RK_NUMSF;
 rkda = (rkda & RKDA_DRIVE) | (track << RKDA_V_TRACK) | (sect << RKDA_V_SECT);
 rk_set_done (0);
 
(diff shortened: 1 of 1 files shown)


bscottm:
Suggestion: Could replace the `#define`-s with `static const`. They'll get optimized down to values in Release builds, ensures better type safety.

al20878:
> They'll get optimized down to values

They are already constant numeric expressions (all resulting in small integers) -- there's nothing to optimize.

bscottm:
@al20878: Constants (static or not) are usually emitted to memory, requiring a load. Most compilers figure out that the load can be eliminated, even for debug builds. And you get better type safety as well. Just a suggestion, especially when those preprocessor constants are only used in this one file.

pkoning2:
I interpreted @rms47 Bob's comment to mean: RK11 should simply say it has 203 cylinders because the actual hardware does.  The fact that software, in general, only uses 200 out of the 203 is not relevant to the simulation.

pkoning2:
Constants don't have to go to memory, and don't need a load; that's up to the compiler to sort out.  The only real difference between `static const int foo=1` and `#define foo 1` is that in the former, `foo` has a stated type (const int) while in the latter case it is implicit.  But that isn't much of a difference, especially given the automatic type conversion the compiler will do for either of them.

pkoning2:
Part of my comment was from misreading the code.  But it still has the notion of "reserved" and passes that to sim_disk_attach_ex2.  Why that function exists at all I have no idea, and the point is that, hardware wise, there isn't any such thing as reserved tracks.

al20878:
> sim_disk_attach_ex2. Why that function exists at all I have no idea, and the point is that, hardware wise, there isn't any such thing as reserved tracks.

That goes back to # 1065 in the old days simh/simh, which, incidentally, was reported by me.  What followed was a "fix" (not by me) for otherwise unattachable images of "unreasonable" size.  I agree that the fix is a band-aid, but without it, it'd be currently impossible to mount 200-cylinder RK images (with recognized file systems) read-only.  Which I am sure is worse that having this "artificial" split (for reserved) for that function.

pkoning2:
Wouldn't it suffice to turn off autosizing for RK11 since that's a meaningless operation?

al20878:
> Wouldn't it suffice to turn off autosizing for RK11 since that's a meaningless operation?

I did not experiment with that.  But what I know is that RK images mostly come in two different flavors, 200- and 203- cylinders. `sim_disk.c` is not capable of handling that gracefully (in fact, it's very inflexible esp. once it "detects" a file system in there).
IDK what is the side-effect of the side-effect of turning auto-size off.  That's because it's not only about the actual hardware aspects of the drive (which can be 200-cyl, 203-cyl or plain under-allocated in the image file), but also a lot of software aspects (which I oppose quite strongly when they are used in simulator except for purely informational purposes).  Time and again I stressed that SimH can't understand software on disk, it can't make any judgments or assumptions as to what's in there.  But `sim_disk.c` does!

al20878:
> I did not experiment with that. 

This PR, however, hopefully fixes the issue #81 as reported, by disabling to iterate over the RK05's only type, thus causing the spurious error messages.

pkoning2:
What I meant is that autosizing in "attach rk<n>" is not a meaningful operation.  Clearly autosizing is meaningful for controllers that support drives of differing sizes.

FWIW, I have an RK05 image that is only 811 blocks long -- a RSTS V3C image.  It attaches with the expected warning, but apart from that it works just fine.

al20878:
> absolutely everyone missed the primary metadata detail which inspired it's addition.

It may very well may be, although this PR has nothing to do with that.

al20878:
@pkoning2

> RSTS V3C image

Incidentally, can I take a look at that, please?  (I want to see if there was RP11 support in there by default.)  Thx

pkoning2:
I got it from [here](http://www.bitsavers.org/bits/DEC/pdp11/dectape/rsts/v3/).  That's a pre-built system; sysgen appeared in V4 so before then you got the system built for you by DEC and shipped on DECtape or paper tape.  Also, RP11 support was added in V4.
It's a bit tricky to use because V3 manuals don't seem to exist.  There are some V1 manuals (?) but fun little details like "what does L D R S A mean when you boot the system???" are a bit obscure.

al20878:
Below is off topic:
Thanks, @pkoning2 !

> There are some V1 manuals (?)

Where?

>  fun little details like "what does L D R S A mean when you boot the system???" are a bit obscure

http://bitsavers.org/pdf/dec/pdp11/rsts-11/PL-11-71-02-01-A-D_RSTS-11_System_Managers_Guide_197105.pdf
Page 5 explains all but A (or O, below)...
I was able to transfer the system to an RF disk but it does not boot properly so I must have been doing something wrong. Where can we discuss that if there's any interest?
```
sim> B TC
ROLLIN VERSION 3A-02
#DF:<DT:
THE REEL LABEL INDICATES THAT THE REST OF THE TAPE WAS
NOT DUMPED.  TYPE K TO KILL REQUEST AT THIS POINT, ANYTHING
ELSE TO PROCEED IN FACE OF DANGER:
HUNG DEVICE DT1, TYPE  K  TO KILL REQUEST,
ANYTHING ELSE TO PROCEED: K
#/RSTS

L D R S O ? R
SURE? Y

L D R S O ? S
[1,2]? Y
DD-MON-YY? 5-DEC-99
HH:MM? 12:00

HALT instruction, PC: 000010 (000012)
```
Same with V3C, the halt is at different PC, though...

al20878:
>  I did succeed with the V3C RK version.

I'd appreciate to know more because I couldn't... (Like I said, it'd just halt on RSTS bootstrap.)  Is there any way out of this PR that we talk about it?  Thx

rms47:
This was fixed after 3.12-3 was released. It can be found here:

http://simh.trailing-edge.com/sources/current/PDP11/pdp11_rk.c

Merging this will then show a conflict with the PR.

al20878:
This was supposed to be the V4-specific patch.  Can you please confirm?

markpizz:
Your changes approved by Bob are in https://github.com/simh/simh master branch for some 2 months now.

pkoning2:
> Your changes approved by Bob are in https://github.com/simh/simh master branch for some 2 months now.

... but of course that isn't this repository so I'm not sure how that answers the question.

al20878:
> Your changes approved by Bob are in https://github.com/simh/simh master branch for some 2 months now.

This PR is not there, anyways.  It deals with V4-specific issues, which Bob was not concerned about (and so he said it, and so I had to split this PR in two, for SupnikCurrent and master, respectively (it was your suggestion, IIRC, as well).

The V3 path was applied to SupnikCurrent, but the remaining part (this PR) -- wasn't, in either repo (open-simh or simh).

markpizz:
> The V3 part was applied to SupnikCurrent, but the remaining part (this PR) -- wasn't, in either repo (open-simh or simh).

You are right.  I missed this.  The relevant parts of this commit are now in the master branch of https://github.com/simh/simh

pkoning2:
Sorry to leave things dangling as long as I did, but Bob's V3.12-3 is now in this repository (Supnik-Current branch and v4 "master" both).  Could you update this PR to account for that?

al20878:
> Bob's V3.12-3 is now in this repository

Hmmm...  It looks weird because despite it says that the word count was fixed, the actual patch to do that (follows) wasn't applied in the final V4 source code (only the CYL repositioning)...  IDK how that's possible
```
--- a/PDP11/pdp11_rk.c
+++ b/PDP11/pdp11_rk.c
@@ -752,7 +752,7 @@ if (wc && (err == 0)) {                                 /* seek ok? */
         else {                                          /* normal store */
             if ((t = MAP_WRW (ma, wc << 1, rkxb))) {    /* store buf */
                 rker = rker | RKER_NXM;                 /* NXM? set flag */
-                wc = wc - t;                            /* adj wd cnt */
+                wc = wc - (t >> 1);                     /* adj wd cnt */^M
                 }
             }
         break;                                          /* end read */
@@ -769,7 +769,7 @@ if (wc && (err == 0)) {                                 /* seek ok? */
         else {                                          /* normal fetch */
             if ((t = MAP_RDW (ma, wc << 1, rkxb))) {  /* get buf */
                 rker = rker | RKER_NXM;                 /* NXM? set flg */
-                wc = wc - t;                            /* adj wd cnt */
+                wc = wc - (t >> 1);                     /* adj wd cnt */^M
                 }
             }
         if (wc) {                                       /* any xfer? */
```

pkoning2:
Looks like my pilot error.  I merged the part that Bob approved for V3.12 earlier (https://github.com/open-simh/simh/commit/3f9ea0f6bd26080b31ad1348001d8ad8c49f0716).  When I merged 3.12-3 to V4, I merged only the 5 commits that made up the delta to the Supnik-Current branch, and I didn't realize that your earlier fix hadn't been merged into V4 yet.  Sorry for that confusion.
Simplest fix is probably for you just to pick up that change, which should amount to the part that you were expecting to see already but didn't.

PR #148, opened 2022-12-11
AltairZ80: Adds to MMD device ability to set drive as single/double sided

deltecent:
While working on archiving disk images for the Morrow Micro Decision, I realized that I could not create a single-sided disk image from scratch using the simulator because there was no way to tell the simulator that the disk "in the drive" was single-sided as it would always assume double-sided if the size of the attached image was not single-sided. When the simulator creates a new file, the size is 0. This PR adds a "SET MMDx SIDES={1|2}" command to force the simulator to treat the attached disk image as single or double-sided, regardless of its size.

This PR also changes the default from double-sided to single-sided.

```
sim> set mmd ena
sim> attach mmd0 ss.dsk
%SIM-INFO: MMD0: creating new file    # size is 0
sim> attach mmd1 ds.dsk
%SIM-INFO: MMD1: creating new file    # size is 0
sim> show mmd
MMD     NODIAG, ROM=v3.1, 5 units
  MMD0  204KB, attached to ss.dsk, SS
  MMD1  204KB, attached to ds.dsk, SS
  MMD2  0MB, not attached, N/A
  MMD3  0MB, not attached, N/A
  MMD4  not attached, N/A
sim> set mmd1 sides=2
sim> show mmd
MMD     NODIAG, ROM=v3.1, 5 units
  MMD0  204KB, attached to ss.dsk, SS
  MMD1  409KB, attached to ds.dsk, DS
  MMD2  0MB, not attached, N/A
  MMD3  0MB, not attached, N/A
  MMD4  not attached, N/A
sim>
```

This PR also removes extraneous ^M carriage return characters that were appearing in my vi editor.  It seems like when I copied in the ROMs this file ended up partially formatted DOS and partially Unix. I made it all DOS formatted. If someone knows a better way to fix this without creating 2,000 changes, please let me know.

### DIFF
--- a/AltairZ80/mmd.c
+++ b/AltairZ80/mmd.c
@@ -70,7 +70,7 @@
 
     Track = (2 * Cylinder) + Side
 
-    MTAB contains one 9 byte entry for each logical drive.
+    CP/M MTAB contains one 9 byte entry for each logical drive.
     The bytes of each entry are defined as follows:
 
     Byte 0          DSKDEF0:
@@ -180,6 +180,8 @@ static t_stat mmd_set_diag(UNIT *uptr, int32 val, CONST char *cptr, void *desc);
 static t_stat mmd_show_diag(FILE *st, UNIT *uptr, int32 val, CONST void *desc);
 static t_stat mmd_set_rom(UNIT *uptr, int32 val, CONST char *cptr, void *desc);
 static t_stat mmd_show_rom(FILE *st, UNIT *uptr, int32 val, CONST void *desc);
+static t_stat mmd_set_sides(UNIT *uptr, int32 val, CONST char *cptr, void *desc);
+static t_stat mmd_show_sides(FILE *st, UNIT *uptr, int32 val, CONST void *desc);
 static int32 mmddev(int32 Addr, int32 rw, int32 data);
 static int32 mmdrom(int32 Addr, int32 rw, int32 data);
 static int32 mmdmem(int32 Addr, int32 rw, int32 data);
@@ -2550,7 +2552,8 @@ static uint8 *mmd_rom = mmd_rom_31;    /* Default 3.1 ROM */
 #define MMD_BPT         (MMD_SPT * MMD_SECTOR_LEN)
 #define MMD_TRACKS      40
 #define MMD_SIDES       2
-#define MMD_CAPACITY    (MMD_BPT * MMD_TRACKS * MMD_SIDES)  /* Default Disk Capacity */
+#define MMD_SS_CAPACITY (MMD_BPT * MMD_TRACKS)              /* SS Disk Capacity */
+#define MMD_DS_CAPACITY (MMD_BPT * MMD_TRACKS * MMD_SIDES)  /* DS Disk Capacity */
 
 #define MMD_IO_BASE     0xf0
 #define MMD_IO_SIZE     14        /* FE is used by SIMH */
@@ -2569,7 +2572,6 @@ static uint8 *mmd_rom = mmd_rom_31;    /* Default 3.1 ROM */
 #define MMD_REG_FDCDATA  0xfb
 #define MMD_REG_CDATA    0xfc
 #define MMD_REG_CSTAT    0xfd
-
 
 #define MMD_CH_1         0x7e    /* Select CTC Channel 1 */
 #define MMD_CH_2         0xbe    /* Select CTC Channel 2 */
@@ -2661,7 +2663,7 @@ typedef struct {
     uint8   stat1[MMD_MAX_DRIVES];          /* Status 1 Register */
     uint8   stat2[MMD_MAX_DRIVES];          /* Status 2 Register */
     uint8   stat3[MMD_MAX_DRIVES];          /* Status 3 Register */
-    uint8   c[MMD_MAX_DRIVES];
+    uint8   c[MMD_MAX_DRIVES];              /* Cylinder Register */
     uint8   hd[MMD_MAX_DRIVES];             /* Head Register */
     uint8   h;                              /* Head Register */
     uint8   r;                              /* Sector Register */
@@ -2715,10 +2717,10 @@ static MMD_CTX mmd_ctx_data = {
 static MMD_CTX *mmd_ctx = &mmd_ctx_data;
 
 static UNIT mmd_unit[MMD_UNITS] = {
-    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, MMD_CAPACITY), 10000 },
-    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, MMD_CAPACITY), 10000 },
-    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, MMD_CAPACITY), 10000 },
-    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, MMD_CAPACITY), 10000 },
+    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, 0), 10000 },
+    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, 0), 10000 },
+    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, 0), 10000 },
+    { UDATA (mmd_svc, UNIT_FIX + UNIT_ATTABLE + UNIT_DISABLE + UNIT_ROABLE, 0), 10000 },
     { UDATA (mmd_sio1_svc, UNIT_ATTABLE + UNIT_DISABLE, 0), 10000 },
 };
 
@@ -2744,8 +2746,10 @@ static REG mmd_reg[] = {
 static MTAB mmd_mod[] = {
     { MTAB_XTD|MTAB_VDV|MTAB_VALR, 0, "DIAG", "DIAG={ENABLE|DISABLE}",
         &mmd_set_diag, &mmd_show_diag, NULL, "Set/Show ROM DIAG enabled/disabled status"},
-    { MTAB_XTD|MTAB_VDV|MTAB_VALR, 0, "ROM", "ROM={25|31}",
+    { MTAB_XTD|MTAB_VDV|MTAB_VALR, 0, "ROM", "ROM={13|23|24|25|31}",
         &mmd_set_rom, &mmd_show_rom, NULL, "Set/Show ROM version"},
+    { MTAB_XTD|MTAB_VUN|MTAB_VALR, 0, "SIDES", "SIDES={1|2}",
+        &mmd_set_sides, &mmd_show_sides, NULL, "Set/Show disk sides"},
     { 0 }
 };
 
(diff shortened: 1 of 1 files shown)


markpizz:
I haven't looked at this source module before, but please explain what the MMD device is.  Your comments above suggest you're talking about a disk/floppy drive.  That's fine.  

Your comments at the top of the module suggest that there is an additional device possible with this hardware, which is, in fact, a modem.  As a consequence, you reasonably have a separate DEVICE which provides the modem functionality.

However, looking at the code, I see the MMD DEVICE's attach routine mmd_attach has this code:
```
static t_stat mmd_attach(UNIT *uptr, CONST char *cptr)
{
    t_stat r;
    unsigned int i = 0;

    /* Attaching to console interface? */
    if (uptr == &mmd_dev.units[MMD_SIO1_UNIT]) {
        if ((r = tmxr_attach(&mmd_tmxr, uptr, cptr)) == SCPE_OK) {
            mmd_tmln[0].rcve = 1;

            sim_debug(VERBOSE_MSG, uptr->dptr, "attached '%s' to console interface.\n", cptr);
        }

        return r;
    }

    r = attach_unit(uptr, cptr);    /* attach unit  */
    if (r != SCPE_OK) {             /* error?       */
        sim_debug(ERROR_MSG, &mmd_dev, "ATTACH error=%d\n", r);
        return r;
    }

    /* Determine length of this disk */
    if (sim_fsize(uptr->fileref) != 0) {
        uptr->capac = sim_fsize(uptr->fileref);
    } else {
        uptr->capac = MMD_CAPACITY;
    }

    for (i = 0; i < MMD_UNITS; i++) {
        if (mmd_dev.units[i].fileref == uptr->fileref) {
            break;
        }
    }

    if (i >= MMD_UNITS) {
        mmd_detach(uptr);

        return SCPE_ARG;
    }

    /* Set Status 3 */
    mmd_ctx->MMD.stat3[i] |= MMD_STAT3_RY;
    if (uptr->flags | UNIT_RO) {
        mmd_ctx->MMD.stat3[i] |= MMD_STAT3_WP;
    }

    if (uptr->capac == MMD_CAPACITY) {
        mmd_ctx->MMD.stat3[i] |= MMD_STAT3_TS;
    }

    /* Default for new file is DSK */
    uptr->u3 = IMAGE_TYPE_DSK;

    sim_debug(VERBOSE_MSG, uptr->dptr, "unit %d, attached to '%s' size=%d\n",
        i, cptr, uptr->capac);

    return SCPE_OK;
}
```

Notice the early call to  tmxr_attach().  It is hard to imagine a device which could be either a serial port or a disk drive.  What is the intention here?

deltecent:
> Notice the early call to tmxr_attach(). It is hard to imagine a device which could be either a serial port or a disk drive. What is the intention here?

@markpizz, that is a good question. I set out to emulate an entire computer system, the Morrow Micro Decision, because I had purchased one and had no boot media. This is a single board computer system that has a Z80, RAM, ROM, 1-4 floppy drives, two serial ports and a parallel port. I initially made the serial ports a separate device from the floppy drives. I then ran into a problem where the second serial port uses port 0xFE which is used by the SIMH pseudo device. A modification was made to move the SIMH pseudo device to another port but the READ and WRITE utilities are fixed at 0xFE.

After waffling back and forth, I ended up making MMD0-3 drives, MMD4 the console port, and MMDM the second serial port so it could be disabled allowing access to the SIMH pseudo device. I also have another PR in the queue that adds MMDP for the printer port (which brings up another question about determining if attach is being asked to attach a file or TMXR device as the printer could be connected to either).

Since MMD4 is actually a serial port, not a drive, it allows TMXR connections and has its own service routine.

This is also the device where you explained THROTTLE, as some of the diagnostics I found for the system would not run without timing close to a 2MHz Z80. This was also my reason for adding Z80 interrupt support.

With AltairZ80 and FlashFloppy I was able to get the system booted and make some real boot disks for it.

I was reading somewhere today that it may be possible to disable specific units rather than an entire device. If that's the case, I don't think I knew that at the time and probably thought if I disabled 1 serial port, I'd have to disable the other, or have a whole slew of separate devices.

The MMD is a new addition to AltairZ80 so if there's a better way to handle it, I'd certainly be open to changing it.

markpizz:
If the MMD (Morrow Micro Decision) is a system with a collection of separate devices, then you should model it as a set of DEVICEs each with their own UNITs, REGs and MTAB entries for configuration.  I have no idea how the AltairZ80 simulator manages and/or validates device addressing and interrupt dispatching, and/or what mechanisms exist to avoid conflicts, but surely @psco and @hharte could  provide help.

BTW, It is possible for a DEVICE to have UNITs which are individually disabled (UNIT_DISABLE).  I.e. the MMDM device could have 2 units, each separately disablable and separately attachable possibly to separate TMXR structures, but otherwise similar logic.

deltecent:
> BTW, It is possible for a DEVICE to have UNITs which are individually disabled (UNIT_DISABLE). I.e. the MMDM device could have 2 units, each separately disablable and separately attachable possibly to separate TMXR structures, but otherwise similar logic.

I think my lack of understanding that individual units could be disabled is probably why I did it the way I did. Unless @psco or @hharte have objections to changing it around, I'd like to make the drives, serial ports, and printer separate devices and get the first serial port out of the drive device. That makes a lot more sense and should simplify the code.

For clarification, which simulator function gets called when "SET DEVICEn DIS" is called? It is my understanding that if the entire device is disabled the reset routine is called and resources can be released. How does the simulator know the unit was enabled/disabled so it can release resources?

markpizz:
No simulator specific function is called when a UNIT is disabled or enabled.  The DEVICE reset routine is called as part of the setup for a RUN or a BOOT command.  The DEVICE reset routine can make sure the DEVICE's units are properly/reasonably setup and complain about things which don't look good.  The DEVICE reset routine is called when a DEVICE is enabled or disabled, reasonable default setup of the DEVICE's UNIT's would normally be the best approach.  UNITs can be changed mostly on the fly, but for the most part, the MTAB configuration routines should probably not allow for setting UNIT configurations once the UNIT is attached.  That detail is completely up to the logic in the UNIT's set validation logic.

deltecent:
As a device has no idea if a unit gets disabled, this makes me think I need 4 devices for the Micro Decision:

MMD[0-3] - The "main" device for ROM, RAM, and floppy controller
MMDC - The serial console port
MMDM - The serial modem/aux port
MMDP - The printer port

Enabling MMD can then enable MMDC, MMDM, and MMDP, but they can also be disabled individually freeing up those I/O resources if necessary.

markpizz:
> As a device has no idea if a unit gets disabled....

Nothing cares until the UNIT could be used which will come as a consequence of a RUN or BOOT.

> MMD[0-3] - The "main" device for ROM, RAM, and floppy controller

You're describing 3 devices: MMDRO, MMDRA, MMDC.  There is nothing saying that enabling/disabling any one of the 3 separate devices then gets its reset routine be called and that routine looks at the state of the DEVICE that just changed and reaches over and does the same (enable or disable) to the other two.  Maybe that might also make sense for some or all of the other MMDC, MMDM and MMDP devices...

This allows for the floppy device to have as many UNITs as appropriate.

deltecent:
When I enable a device, the reset routine starts a timer to poll for data. When I disable that device, it won't disable it because it thinks it's busy:

```
sim> set mmdm dis
%SIM-ERROR: MMDM has attached or busy units
```

I'm assuming that's the timer since I had not attached anything.

Is it possible to tell SCP that it's okay to terminate any timers on a disable or is there something else I should be doing?

deltecent:
Let's see if reset gets called..... and stop the timer there.....

markpizz:
> When I enable a device, the reset routine starts a timer to poll for data.

What data would be possible if nothing has yet been attached.  Polling without a place to poll shouldn't ever return data and thus is a waste of time....

Start your polling in your attach logic.

Your detach logic should call sim_cancel(uptr) where uptr is the unit being detached.

> When I disable that device, it won't disable it because it thinks it's busy:

You can retain your prior DEVICE  enabled state locally and only when the reset routine is called and the enabled state is enabled AND it hasn't transitioned, then you know you're getting called from RUN or a BOOT and that would be the time to start your timer.  BTW, any scheduled units will always be unscheduled at the beginning phase of a RUN or BOOT command and thus you'd get control.

If you're trying to disable one or all of the devices from one of the other devices as I suggested, you can explicitly call sim_cancel() for all of the units on the devices you're disabling.  Calling sim_cancel() for a unit that isn't scheduled is harmless.

markpizz:
Like I said, if the DEVICE was previously enabled when reset is called, only then do you start the timer since that call to DEVICE reset will be due to a RUN or BOOT command.

deltecent:
I apologize but I'm not understanding something. When I ENABLE a device, it is my understanding that the device is now ready to go. I can input a small program by hand, single step it, and send a character out the serial port without having to RUN or BOOT first. We are emulating hardware with a front panel that may not have any drives attached to it (hex loader through MMDM port, for example).

Reset is called when the simulator starts and on enable, disable, run, and boot. The reset function has no idea which.

I know I'm going to kick myself for bringing up SCP, but it seems that a device's timer shouldn't make it "busy" when trying to disable (unplug) it. Something attached, yes. A timer, no.

This entire exercise is so the MMDM device can be enabled and disabled at will in order to going access to the SIMH psuedo driver. My understanding is that the service routine, which executes from a timer, is there to simulate hardware processing in the background, like a disk controller or UART, all of which are active the minute a machine is powered on.

```
sim>ena mmd
sim>ena mmdm
sim>boot
Do some stuff with MMDM device
sim>dis mmdm
sim>c
Use the SIMH pseudo device - upload new files from host
sim> ena mmdm
sim> c
Use the MMDM device again
<repeat>
```

I can't seem to do this because I can't disable the MMDM device because of a timer.

Looking forward to your solution.

deltecent:
What is the reasoning behind that behavior. All I can think of is it may not be  good practice to have SCP cancel a device's resources on its behalf or possibly allow a disabled device to leave a timer on the queue by it not cancelling the timer in its reset function. In other words, SCP shouldn't have to (or won't) clean up after a device.

markpizz:
> What is the reasoning behind that behavior. All I can think of is it may not be good practice to have SCP cancel a device's resources on its behalf or possibly allow a disabled device to leave a timer on the queue by it not cancelling the timer in its reset function. In other words, SCP shouldn't have to (or won't) clean up after a device.

I already answered this, with "In general, simh simulators really don't expect their configuration to change dynamically while executing." but please provide examples of systems of this vintage which would allow you to rip out and/or randomly plug in different pieces of hardware while it was running.

Like I also said, "You wish it would behave that way, meanwhile the other 77 simh simulators expect it to behave as it now does." which means SCP shouldn't be changed to accommodate your situation.

I then offered a suggestion how you could solve what you want: "In theory, you could have an MTAB entry that implemented your own SET MMDM OFFLINE that called sim_cancel on the mmdm unit and then set the DEVICE disabled."

Do you have questions about how to do this?

deltecent:
> but please provide examples of systems of this vintage which would allow you to rip out and/or randomly plug in different pieces of hardware while it was running

Really? You do realize these are simulated machines that allow us to do a lot of things we couldn't do with real systems of this vintage. That's one of the things that makes them so useful. I couldn't make a socket connection to Altair 8800 in 1975 but here we are with TMXR. I couldn't shove a thumb drive in the Altair's USB port and have immediate access to thousands of floppy disks. I couldn't just change the value of a register. I couldn't send debug output to a remote console or file. I couldn't boot CP/M, transfer a file from another host, assemble it, and copy listing files back the host in under 2 seconds, but here we are. The whole idea of somehow not being able to "rip out and/or randomly plug in different pieces of hardware while it was 'running'" in a fantasy environment because we couldn't do that with real hardware is just nonsense.

You typically have good technological reasons backing your opinions and I was hoping to learn what that was. This isn't one of them.

> meanwhile the other 77 simh simulators expect it to behave as it now does.

Please provide examples of how SCP cancelling timer resources on a DISABLE would somehow break 77 simulators due to their expectations that such an action isn't allowed.

markpizz:
>> meanwhile the other 77 simh simulators expect it to behave as it now does.

Changing core functionality to satisfy you, especially when a trivial solution exists is not going to happen.

STOP ARGUING AND PLEASE TRY:
```
static t_stat mmdm_offline(UNIT *uptr, int32 val, CONST char *cptr, void *desc)
{
    DEVICE *dptr = find_dev_from_unit(uptr);

    sim_cancel (uptr);
    dptr->flags |= DEV_DIS;
    return SCPE_OK;
}

static MTAB mmdm_mod[] = {
    { MTAB_XTD|MTAB_VDV, 0, "OFFLINE", "OFFLINE",
        &mmdm_offline, NULL, NULL, "Set MMDM device disabled"},
    { 0 }
};
```
```
sim> show mmdm
MMDM    disabled
sim> set mmdm enable
sim> show mmdm
MMDM    OFFLINE
        not attached
sim> set mmdm offline
sim> show mmdm
MMDM    disabled
sim>  
```

PR #154, opened 2022-12-27
Fix operand order bugs in PDP-11 FP11 disassembly

nickd4:
### Issue

I was using the SIMH PDP-11 disassembler for some fuzzing tests in another project... and I discovered a disassembly issue. Instructions like `addf 001234,f0` are incorrectly being disassembled with the operands swapped, thus `addf f0,001234`.

### Fix

This PR fixes the issue, according to the reference I used which is at: 
  https://archive.org/details/bitsavers_decpdp11haorHandbook1981_37639914

Basically all of the 2-operand floating point instructions such as `modf`, `addf`, etc get their operands swapped in the disassembly printout but otherwise should work as before, *except* storage instructions like `stf`, `std`, `stexp`, `stcfi` etc which I have not changed.

### DIFF
--- a/PDP11/pdp11_sys.c
+++ b/PDP11/pdp11_sys.c
@@ -342,6 +342,9 @@ return SCPE_CSUM;
 #define I_V_CCC         14                              /* CC clear */
 #define I_V_CCS         15                              /* CC set */
 #define I_V_SOPR        16                              /* operand, reg */
+#define I_V_FOPA        17                              /* flt operand, fac */
+#define I_V_SOPA        18                              /* operand, fac */
+#define I_V_SMDA        19                              /* moded int op, fac */
 #define I_NPN           (I_V_NPN << I_V_CL)
 #define I_REG           (I_V_REG << I_V_CL)
 #define I_3B            (I_V_3B << I_V_CL)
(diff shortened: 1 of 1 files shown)


markpizz:
I haven't looked at the details of what you've done, but assembly instructions in this and other simulators are not just displayed.  They are also interpreted by the DEPOSIT command.  So, if a fix is needed to the EXAMINE side of things, a related fix is likely needed on the deposit side.

Meanwhile, this issue is specific to the PDP11 simulator which Bob Supnik (@rms47) is the author and the primary maintainer. 
 Any problems should be filtered through him for specific approval or for him to fix himself.

rms47:
Mark is correct. I've sent him an updated module that fixes both sides of the equation.

nickd4:
That sounds good. Thank you guys.

pkoning2:
Nick, so if you have updated code from Bob, could you revise this PR to reflect that?

nickd4:
My reading of @rms47's comment was that he had sent the further fix directly to @markpizz, is that correct Bob? I can look at examine/deposit if need be, I have not done so yet.

markpizz:
Commit id 9319f6027e2b3cff70b375714d9df2726f7f858b in the https://github.com/simh/simh repo contains Bob's solution to this problem.

rms47:
This fix was generated after 3.12-3 was released. It can be found here:

[http://simh.trailing-edge.com/sources/current/PDP11/pdp11_sys.c](url)

wfjm:
@rms47 wrote:
> This fix was generated after 3.12-3 was released.

True indeed, 3.12-3 was released 6 weeks ago. And before other changes are introduced into the PDP-11 code base this latest release from Bob should first be merged into the `SupnikCurent` branch, and from there integrated into `master`. As it was done for many years. See also #159.

pkoning2:
3.12-3 has been merged.
I just created PR #170 with Bob's code, which is a superset of what you have here.  So I think the right thing is to close this PR and merge the other one.

PR #160, opened 2023-01-15
AltairZ80: Adds -H switch to LOAD command for loading Intel hex files

deltecent:


psco:
@pkoning2 This looks fine and can be merged.

markpizz:
@psco wait a minute:

@deltecent, your coding sequence:
```
            sim_printf("Hex file format error at line %d\n", line);
            return SCPE_IERR | SCPE_NOMESSAGE;
```
could more simply be replaced by:
```
            return sim_messagef (SCPE_IERR, "Hex file format error at line %d\n", line);
```
This appears 4 times.

deltecent:
```
sim> load -h blank.hex
%SIM-INFO: 0 bytes loaded at 0.
sim> load -h pc2flop.hex
%SIM-ERROR: Checksum error at line 22
   :100250003E01CD5603CA0802BBC208023A65092F00
```

Is there any way to hide the superfluous `%SIM-XXXX` message prefix? "Checksum **error**" is self-explanatory.

markpizz:
> Is there any way to hide the superfluous %SIM-XXXX message prefix? "Checksum error" is self-explanatory.

Well, originally sim_messagef() messages were without the prefix (which, when specified as sim_messagef(SCPE_OK, "Message text.\n"); appear as SIM-INFO: Message text.

The prefix resolved numerous confusing concerns various folks had about the messages.  

Messages with SIM-ERROR: prefix are completely unambiguous and will not produce ambiguities for folks.  Hopefully someone who sees such a message will be very rare since most folks will be providing good input data.

You might want to include a sim_messagef(SCPE_OK, "With a summary of the success of the operation that was performed\n"); at the end of the load operation.

SIM-INFO messages can be suppressed if the invoking command has a -Q switch specified.
SIM-ERROR messages can be suppressed if SET NOMESSAGE has been specified.  SIM-ERROR is emitted for any return value that is not SCPE_OK.

The SCPE_XXX error value used in a sim_messagef() return statement will trigger any ON xxx trapping as you hopefully would expect.

PR #169, opened 2023-02-02
Floating loads are src,dst (nickd4)

pkoning2:


rms47:
This was fixed after v3.12-3 in
http://simh.trailing-edge.com/sources/current/PDP11/pdp11_sys.c

pkoning2:
> This was fixed after v3.12-3 in http://simh.trailing-edge.com/sources/current/PDP11/pdp11_sys.c
> […](#)

Yes, that's the fix I picked up and have here in this PR.  So the intent is to reflect that newer fix in this repository branch.

PR #171, opened 2023-02-02
Identify Open SIMH builds in version strings

tlhackque:
With three simulators sharing a common heritage and name, it can get confusing when questions arise from the user community.

This edit changes the Minor version from 0 to 1 (e.g. V4.1) and adds "Open SIMH" to the banner and show version output.


markpizz:
I have no issue with the purpose of this PR.

However, we've strived to avoid merge commits, so that detail should be addressed before accepting the PR.

PR #174, opened 2023-02-04
AltairZ80, sim_imd: Fix static code analysis warnings.

hharte:
Summary of changes:

* Fix static code analysis issues in sim_imd.c.
* Fix static code analysis issues in Flashwriter and CompuPro Disk3.

Compiled with:

VS 2022 17.5.3 (Debug and Release)
VS 2008 (Debug and Release)
MacOS (x64) clang-1400.0.29.202
Linux gcc 10.2.1
Automated tests: SIMH running on Windows, Linux (ARM), MacOS (x64):

[CompuPro CP/M-80](https://schorn.ch/cpm/zip/cpm86.zip)
[CompuPro CP/M-86](https://schorn.ch/cpm/zip/cpm86.zip)
[SCP 86-DOS](https://schorn.ch/cpm/zip/86dos.zip)
[MS-DOS 2.11](https://github.com/hharte/altairz80-packages/tree/main/msdos211_test)
[CompuPro CP/M-68K v1.3](https://github.com/hharte/altairz80-packages/tree/main/ccpm68k13) (68000 and 68010 versions)
[CP/M-68K v1.2](https://schorn.ch/cpm/zip/cpm68k.zip)
[Advanced Digital Super-Six CP/M 2.2](https://github.com/hharte/altairz80-packages/tree/main/adc_cpm2)
[Advanced Digital Super-Six CP/M 3.0](https://github.com/hharte/altairz80-packages/tree/main/adc_cpm3)
[DIGITEX CP/M 2.2](https://github.com/hharte/altairz80-packages/tree/main/dig_cpm2)
[DIGITEX CP/M 3.0](https://github.com/hharte/altairz80-packages/tree/main/dig_cpm3)
[DIGITEX OASIS multi-user version 5.6](https://github.com/open-simh/simh/pull/github.com/hharte/altairz80-packages/tree/main/dig_oasis56)
[CP/M-80 (Altair)](https://schorn.ch/cpm/zip/cpm2.zip) ZEXALL Z80 instruction set test.

FYI, @psco , @markpizz

### DIFF
--- a/AltairZ80/altairz80_cpu.c
+++ b/AltairZ80/altairz80_cpu.c
@@ -6628,7 +6628,7 @@ static const char* m68kVariantToString[] = {
 
 static t_stat chip_show(FILE *st, UNIT *uptr, int32 val, CONST void *desc) {
     fprintf(st, cpu_unit.flags & UNIT_CPU_OPSTOP ? "ITRAP, " : "NOITRAP, ");
-    if (chiptype < NUM_CHIP_TYPE) {
+    if ((chiptype >= 0) && (chiptype < NUM_CHIP_TYPE)) {
         fprintf(st, "%s", cpu_mod[chiptype].mstring);
         if (chiptype == CHIP_TYPE_M68K) {
             fprintf(st, " (%s)", m68kVariantToString[m68kvariant]);
@@ -6677,7 +6677,7 @@ static t_stat cpu_show(FILE *st, UNIT *uptr, int32 val, CONST void *desc) {
             }
         fprintf(st, "]");
     }
-    if (chiptype < NUM_CHIP_TYPE) {
+    if ((chiptype >= 0) && (chiptype < NUM_CHIP_TYPE)) {
         first = TRUE;
         /* show verbose CPU flags */
         for (i = 0; cpuflags[chiptype][i].mask; i++)
@@ -6861,7 +6861,7 @@ static int32 bankseldev(const int32 port, const int32 io, const int32 data) {
 }
 
 static void cpu_set_chiptype_short(const int32 value) {
-    if ((chiptype == value) || (chiptype >= NUM_CHIP_TYPE))
+    if ((chiptype == value) || (value < 0) || (value >= NUM_CHIP_TYPE))
         return; /* nothing to do */
     if (((chiptype == CHIP_TYPE_8080) && (value == CHIP_TYPE_Z80)) ||
         ((chiptype == CHIP_TYPE_Z80) && (value == CHIP_TYPE_8080))) {
@@ -7090,7 +7090,7 @@ static t_stat cpu_set_hist(UNIT *uptr, int32 val, CONST char *cptr, void *desc) 
    uint32 i, lnt;
    t_stat r;
 
-    if ((chiptype != CHIP_TYPE_8080) && (chiptype != CHIP_TYPE_Z80)) {
+    if ((chiptype >= 0) && (chiptype != CHIP_TYPE_8080) && (chiptype != CHIP_TYPE_Z80)) {
         sim_printf("History not supported for chiptype: %s\n",
                (chiptype < NUM_CHIP_TYPE) ? cpu_mod[chiptype].mstring : "????");
         return SCPE_NOFNC;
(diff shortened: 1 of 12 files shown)


AltairZ80/altairz80_cpu.c
```diff
@@ -7359,6 +7369,9 @@ static t_stat cpu_hex_load(FILE *fileref, CONST char *cptr, CONST char *fnam, in
 
         bufptr = datastr;
 
+        /* Ensure datastr is not NULL */
+        datastr[strlen(datastr)] = '\0';
```
pkoning2, line 7373: That doesn't do anything.  "strlen" returns the offset at which a 0 byte is found, so if strlen works then this just writes 0 over that existing 0.  And if there isn't a 0 termination in place, strlen isn't valid.

AltairZ80/altairz80_sio.c
```diff
@@ -61,14 +61,18 @@ uint8 *URLContents(const char *URL, uint32 *length);
 #define RESULT_LEAD_IN          "URL is not supported on this platform. START URL \""
 #define RESULT_LEAD_OUT         "\" URL END."
 uint8 *URLContents(const char *URL, uint32 *length) {
-    char str[RESULT_BUFFER_LENGTH] = RESULT_LEAD_IN;
+    char str[RESULT_BUFFER_LENGTH];
     char *result;
+    strncpy(str, RESULT_LEAD_IN, strlen(RESULT_LEAD_IN));
+    str[strlen(RESULT_LEAD_IN)] = '\0';
```
Rhialto, line 67: Why would copying in the string be better than initializing it? And even when this would be better, why not use strcpy, instead of the complicated dance with strncpy and strlen? In any case, with initialization, the compiler can warn when the string doesn't fit; gcc would warn `warning: initializer-string for array of chars is too long`.

pkoning2: Strnlen is meant to protect the destination from overrun, but for that to works requires using the size of the destination, not the size of the source, as the length limit.  As written it's merely a complicated way to say "strcpy" with all the hazards of that function.

markpizz:
In general, strlcpy() and strlcat() should be used these days and avoiding potential errors with strncpy().

The simh codebase has working strlcpy() and strlcat() for platforms which don't natively have these APIs.

AltairZ80/altairz80_cpu.c
hharte: Thanks, @pkoning2 for pointing that out.  I will upload a new patch set to resolve.

AltairZ80/altairz80_sio.c
hharte: Thanks, @Rhialto, @pkoning2 for pointing these out.  The VS2022 static analysis tool flags the original code with as "str" not being initialized.  I will upload a new patch set using snprintf() that makes the code more readable and should resolve the issue.

hharte:
Thanks @markpizz for the information.  I will upload a new changes to resolve this.

AltairZ80/altairz80_sio.c
Rhialto: Hm, if a tool looks at `char str[RESULT_BUFFER_LENGTH] = RESULT_LEAD_IN;` and then complains that `str` is not initialized, I'd have to take the results from this tool with a large amount of salt...

Your other changes seem sensible though, at first glance.

hharte:
FYI, @deltecent : The last commit resolves the PC != 0 issue after PR #189. as suggested by @psco .

deltecent:
I was just looking at that and was just about to write you. The precalibrate function resets all the devices when it’s done and cpu_reset wasn’t setting PC to 0.

The PR you sent me looks good.

Thanks.

markpizz:
The routine find_unit_index could be much simpler if it merely looked like:

```
int32 find_unit_index (UNIT *uptr)
{
    DEVICE *dptr;

    dptr = find_dev_from_unit(uptr);
    if (dptr == NULL)
        return -1;
    return (uptr - dptr->units);
}

```

deltecent:
That is a useful function buried in an AltairZ80 device. What about moving that functionality to SCP ?

markpizz:
Well, most cases where it would be used will already have (or found) the dptr, so calling a function to replace 1 line of code (uptr - dptr->units) wouldn't be an advantage.  

Additionally, in theory it could be added to SCP, then that would imply someone willing to change all the places in both the altairz80 as well as all other simulators where such logic is present, otherwise having it in the altairz80 as exists now, is fine.  

I mentioned the simplification since someone is already messing with that code.

hharte:
Thanks Mark for the suggestion.  I looked at several uses of the find_unit_index() in AltairZ80 and most don’t already have the dptr, so I will retain this function and simplify this function as you suggest.

hharte:
@psco this change also addresses your feedback.

psco:
Thank you Howard. @pkoning2 This looks good and can be merged.

hharte:
FYI, @tlhackque , @deltecent , I rebased the original 9 commits on top of master, and added one commit to (hopefully) fix the Windows CI builds.  The AltairZ80.vcxproj was out of date, missing the video changes and two source files: s100_vdm1.c, sol20.c.

hharte:
@pkoning2 this is ready to merge.  Thanks!

Issue #54, opened 2022-08-31
Please can I have prebuilt binaries for simh?


MidnightThunder:
I can't get it to compile by myself and compiling code is not something i really understand, could I please have prebuilt binaries for windows?

pkoning2:
Yes, that's a goal.  It needs someone conversant with Windows to do the necessary work.

Xuser641:
So at the moment it does not work with Windows, correct?

pkoning2:
I haven't tried it, but I assume it does work.  The missing part as far as I know is only the prebuilt binaries, not the ability to run.

LegalizeAdulthood:
I believe we should be able to include pre-built binaries from the continuous integration workflow into the releases here on github.

pkoning2:
Interesting idea.  Is that enough for what is needed?  For the Windows case, installers are typical, though I suppose a simple ZIP archive may be sufficient if nothing installer-flavored is needed.  That would be worth investigating.  Any volunteers?

markpizz:
Well, every commit (or at least every push) produces the CI activity.  Each of those then produces almost 120MB Windows of binaries.  These can be compressed down to about 11MB for all the binaries, if that is stored for each relatively minor commit it seems like it's quite wasteful.

I think it would be better to drive the release concept packaging on some sort of calendar process (at most quarterly maybe).

pkoning2:
Good point.  But the other part of the suggestion was that the CI process is all that is needed to produce the ingredients for kits.  Is that true?

LegalizeAdulthood:
Starting with a ZIP file is an easy place to start; an installer can come later.  It's not hard to do with CMake but would be a giant PITA with the existing makefile, unless it was already there.  (I haven't looked.)

LegalizeAdulthood:
It wouldn't necessarily be producing a package per CI build, it would be leveraging the CI build to create a package associated with a release, either manually added to the release or somehow automated into the release based on a manual trigger.  My point was that we have automated Windows stuff building via the CI infrastructure, so you're about 80-90% of the way towards an automated release mechanism for Windows.

markpizz:
Maybe if the CI build process for Windows produces XP capable binaries (not yet with cmake), then maybe some fancy CI scripting could even automatically determine if the reasonable "release" interval had been passed and then push a formal release.

LegalizeAdulthood:
[...]

PR #178, opened 2023-02-08
Add a workflow for building windows executables

tlhackque:
First pass:
  - Uses current Visual Studio
  - Refers to the pre-fork repo for windows libraries
  - Manually triggered - several options can be selected
  - Triggered on pull request merged

Result is an "artifact" ZIP file that contains all the simulators.  It's retained for 90 days.

To run  manually:
Goto "Actions", select "Windows_Build", Click on the "Run workflow" pulldown.  The defaults are probably what you want.

Click 'Run workflow'.  With a tailwind, wait about 40 minutes.
When the checkmark appears, click on 'windows build'.  Under 'Artifacts', find the zip file, which will (depending on what options you picked) be named something like 'Win32-x86-Release.  Click on the name to download the ZIP file.

Will need some tweaks to switch to an Open SIMH repo for the windows libraries, but for now, this gives us windows builds and binaries.

Note that this doesn't touch the existing windows builds; it has its own SLN file and project files (migrated from the standard 2008 version).  This makes it possible to easily run in the CI infrastructure.  The 2008 version allows us to still build for XP.

The size of the commit is due to adding the migrated project files.  It has no effect on anything except the new workflow.
First step to addressing #54


tlhackque:
Taking a new approach.
Stay tuned.

PR #188, opened 2023-02-27
VIDEO: Several enhancements

deltecent:
Moved SDL_Init(SDL_INIT_VIDEO) to EVENT_OPEN to prevent video window from opening and taking focus until needed.
Added SIM_VID_RESIZABLE flag to allow windows to be resized.
Added SIM_VID_IGNORE_VBAR flag to allow windows to ignore video buffer aspect ratio.
Added vid_set_window_size() function.
Added EVENT_FULLSCREEN to SDL_MAIN_AVAILABLE event processor.

I don't believe these will have any impact on existing simulators. SDL_INIT_VIDEO was moved for Mac users that require `main` being replaced with `SDL_main`.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -313,17 +313,19 @@ static int SDL_SavePNG_RW(SDL_Surface *surface, SDL_RWops *dst, int freedst)
      
  */
 
-#define EVENT_REDRAW     1                              /* redraw event for SDL */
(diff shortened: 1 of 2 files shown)


deltecent:
Included changes in #191 for ease of compiling and testing.

PR #190, opened 2023-02-27
AltairZ80: Correct strange line endings in mmd.c

deltecent:
No other changes.


markpizz:
This should be merged.  It merely corrects mixed CRLF and LF line endings.

deltecent:
Included in #192.

PR #191, opened 2023-02-28
AltairZ80: Adds SOL20 and VDM1 devices

deltecent:
This PR adds support for the Processor Technology VDM-1 display adapter and Sol-20 computer system. The VDM-1 was integrated into the Sol-20 but has been separated out into its own device. This PR is dependent on other PRs currently pending.

This PR adds the following devices to the AltairZ80 simulator:

VDM1   - Processor Technology VDM-1 display adapter
SOL20  - Processor Technology SOL20 with SOLOS ROMs
SOL20K - SOL20 Keyboard (callback from VDM1)
SOL20T - SOL20 Tape (reads/writes cassette file images)
SOL20S - SOL20 Serial Port (TMXR capable)
SOL20P - SOL20 Printer Port (TMXR capable)

I am unable to test other operating systems. I have only tested on OS X Ventura. Testing and feedback would be greatly appreciated. The VDM-1 device uses sim_video which has not been used in AltairZ80 before. I expect there to be some issues.

The simplest way to play around with the Sol-20 is to paste ENT files into SOLOS using the SIMH console.

An example INI file looks something like this:
```
set ptr dis
set ptp dis
set hdsk dis
set sol20 ena
attach sol20t blank.cas
attach sol20s -U 8080
boot sol20
```
At the ">" prompt in the video window, paste ENT file into SIMH console. After the ENT file has been loaded, enter "EX 0" at the ">" prompt.

A lot of Sol-20 software will not work properly running at full simulator speed therefore the device itself enables throttling. If you want to run at full speed, you may disable throttling in your INI file after enabling the SOL20 device. Also, without the precalibration PR, expect the simulator to hang for about 10 seconds after startup.


markpizz:
You were on the right track to update the sim_video code in a separate PR.

As for the changes in that sim_video code, there is no need to change:
```
status = SDL_Init (SDL_INIT_VIDEO);
```
status = SDL_Init (SDL_INIT_EVENTS);```
```
Look at the documentation for SDL_INIT and you will see that SDL_INIT_VIDEO explicitly also initializes the events, as described here: https://wiki.libsdl.org/SDL2/SDL_Init

Once that change is removed, the other SDL_init logic isn't needed.

You needed to shift the numeric values in the list of event types since the new event type had a double digit number and had a long name.  That's fine.  Even though we avoid tabs in simh source code files, it is reasonable to try and align things on 4 or 8 character boundaries since we'll often merely have the editor configured to insert spaces.  Not a big issue.



Your 2nd to last change to sim_video.c added 4 spaces where they don't belong.

@larsbrinkhoff should review the changes to the sim_video PR.

**Comment by deltecent** [comment]
The reason SIM_INIT_EVENTS was separated from SIM_INIT_VIDEO was for the Mac (and maybe other operating systems) that need SDL_Main. With SDL_Main, SIM_INIT_VIDEO is called even if there aren't any devices enabled that need video. This starts the video system which in turn takes focus away from the SIMH console which isn't the intent. There doesn't seem to be any reason to initialize the VIDEO until it is needed by a device and I think it is backwards compatible.

I could encapsulate this change with SDL_Main ifdefs, but I don't think that's necessary.

**Comment by markpizz** [comment]
Your makefile changes in this PR need fixing.

1) Where you set VIDEO_USEFUL for the AltairZ80, you shouldn't be messing with anything about the BESM6 simulator.  It is perfectly valid for someone to build multiple simulators in a single invocation of make, either by explicit name or implied via all.

2) Since you're not using the display APIs, then the ALTAIRZ80_OPT shouldn't have ```${DISPLAY_OPT}```, but it should have ```-DUSE_SIM_VIDEO ${VIDEO_CCDEFS}```

**Comment by markpizz** [comment]
Since you've now got a collection of 6 devices which all must be enabled or all must be disabled at the same time, your logic should produce well behaved results if/when a user enables or disables any one of these devices.  A related problem might happen if a user enables these devices and then disables them.

**Comment by deltecent** [comment]
>Since you've now got a collection of 6 devices which all must be enabled or all must be disabled at the same time, your logic should produce well behaved results if/when a user enables or disables any one of these devices.

The Sol-20 only needs 3 devices to boot in SOLOS: SOL20, SOL20K, and VDM1. The SOL20T tape device or floppy controllers make it more functional. The VDM-1 was integrated into the Sol-20. Since the VDM-1 originally existed as its own S-100 card, I made it work on its own.

>A related problem might happen if a user enables these devices and then disables them.

The devices stand on their own. Someone could replace them in their own custom Sol-20 configuration or use them elsewhere. I thought enabling the devices when SOL20 is enabled would be best for most people.

**Comment by deltecent** [comment]
>Since you're not using the display APIs, then the ALTAIRZ80_OPT shouldn't have ${DISPLAY_OPT}, but it should have -DUSE_SIM_VIDEO ${VIDEO_CCDEFS}

It also needs `${VIDEO_LDFLAGS}`.

**Comment by deltecent** [comment]
After some effort, I got AltairZ80 to compile for Windows using Visual Studio and SDL2. One thing I noticed is `sim_messagef()` output doesn't appear on the SIMH console while the simulator is running with a video window open like it does on the Mac. It looks like 'stdout` isn't displayed until ^C is pressed and the simulator stops. Is this normal for Windows?

**Comment by markpizz** [comment]
In general, sim_messagef() is useful while users are configuring or managing the setup or shutdown of the simulator when at the sim> prompt.

Why would you be using sim_messagef() while the simulator is executing instructions?  Please explain the use case and how it relates to how original hardware worked.

**Comment by markpizz** [comment]
BTW, Building with VS2008 produces:
```
2>s100_vdm1.c
2>..\AltairZ80\s100_vdm1.c(368) : error C2143: syntax error : missing ';' before 'type'
2>..\AltairZ80\s100_vdm1.c(368) : error C2143: syntax error : missing ';' before 'type'
2>..\AltairZ80\s100_vdm1.c(368) : error C2143: syntax error : missing ')' before 'type'
2>..\AltairZ80\s100_vdm1.c(368) : error C2143: syntax error : missing ';' before 'type'
2>..\AltairZ80\s100_vdm1.c(368) : error C2065: 'i' : undeclared identifier
2>..\AltairZ80\s100_vdm1.c(368) : warning C4552: '<' : operator has no effect; expected operator with side-effect
2>..\AltairZ80\s100_vdm1.c(368) : error C2065: 'i' : undeclared identifier
2>..\AltairZ80\s100_vdm1.c(368) : error C2059: syntax error : ')'
2>..\AltairZ80\s100_vdm1.c(368) : error C2143: syntax error : missing ';' before '{'
2>..\AltairZ80\s100_vdm1.c(369) : error C2065: 'i' : undeclared identifier
```

To fix this, you need to declare local variables immediately after an open brace and not inline in for statements or in the middle of code.

Note that the s100_vdm1.c file has mixed line endings.

**Comment by markpizz** [comment]
Looking in sol20.c, I see you are using sim_messagef() to indicate a hardware error state to the system user rather than actually passing a hardware error back to the simulator.  I'm certain that the hardware you're simulating must have actually returned some sort of hardware state and didn't have a display a user could see error messages on.  If it  had such a display, then simulating that should be part of the project, and in any case the error condition should be passed back into the simulated system.  

The types of things you're using sim_messagef() and sim_printf() would usually be reserved to debug output that, in general, normal users don't see unless they're deep into debugging problems...

**Comment by deltecent** [comment]
>Note that the s100_vdm1.c file has mixed line endings.

Thank you for noticing those!

**Comment by deltecent** [comment]
I think it's a problem with my Windows build. I can't get `sim_poll_kbd()` to work right with the master branch under Windows and VS 2019. It returns `SCPE_OK` when keys have been entered on the keyboard.

**Comment by deltecent** [comment]
>I'm certain that the hardware you're simulating must have actually returned some sort of hardware state and didn't have a display a user could see error messages on.

Certainly not. The Sol-20 had no idea if there was a cassette in the tape player. The user would physically have to look at the cassette player and notice there wasn't a tape in it. The Sol-20 only provided motor on-off control with no feedback to the OS whatsoever. The purpose of `sim_messagef` is to remind the user that there is no file attached to (inserted in) the tape device. A normal user would probably want to have that feedback rather than scratch their head, exit to `simh>` prompt and run `show sol20t`.

**Comment by deltecent** [comment]
Windows version works just like the Mac version. I had VS2019 installed for use with QT5 and a bunch of Ming stuff to make Windows behave like Unix. I installed VS2008 in another VM, built AltairZ80 and it works fine with all the Sol-20 output. Happy days!

**Comment by deltecent** [comment]
Nope, I was wrong. Stdout works on Windows like the Mac, but `USE_SIM_VIDEO` breaks keyboard input on the console using `sim_poll_kbd`, even if no video window is opened.

**Comment by markpizz** [comment]
There certainly other simulators which reliably detect console WRU with video windows open.

Please provide enough details so that I can reproduce what you're seeing.

**Comment by deltecent** [comment]
I'm trying my best.  I have to get around this:

```
Linking...
Embedding manifest...
mt.exe : general error c101008d: Failed to write the updated manifest to the resource of file "..\BIN\NT\Win32-Debug\AltairZ80.exe". The process cannot access the file because it is being used by another process.
Build log was saved at "file://c:\Users\Patrick\open-simh\BIN\NT\Project\simh\AltairZ80\Win32-Debug\BuildLog.htm"
AltairZ80 - 1 error(s), 0 warning(s)
========== Rebuild All: 0 succeeded, 1 failed, 0 skipped ==========
```
I can't figure out what has the file open. Probably time for a Windows fix-it reboot.

**Comment by markpizz** [comment]
Right click on a blank spot on the task bar and select "Task Manager".  Then select details and then find the AltairZ80.exe process and right click on it and select End Process.  Alternatively, you can start the "Task Manager" if you can enter a Control-Alt-Delete into the VM you'll be presented with a menu where you select "Task Manager".

**Comment by deltecent** [comment]
That was the first thing I tried. No AltairZ80 running. Also tried a `ps` but then remembered I was running Windows. Closed re-opened all running programs. No joy. Logout/Login - Fixed. I'm guessing some VS process had it locked.

**Comment by deltecent** [comment]
@markpizz, I'm so sorry for wasting your time. I have no idea what was going on. I don't know if it had something to do with switching from VS2019 to VS2008 or something else. I ended up removing my entire `open-sim` directory in Windows and cloned from scratch. Now it all works fine. I can start the Sol-20 sim and all status messages appear on the SIMH console, all keyboard input on the SIMH console shows up on the video window, and Ctrl-E stops the simulator, just like the Mac version.

**Comment by markpizz** [comment]
You said:
> I think it's a problem with my Windows build. I can't get sim_poll_kbd() to work right with the master branch under Windows and VS 2019. It returns SCPE_OK when keys have been entered on the keyboard.

This suggests that you "might" be using sim_poll_kbd() to get keyboard input for your GUI presented session.  If I'm wrong, hit me in the head.  Meanwhile, if you're getting general keyboard input using sim_poll_kbd() you should not be doing this, but using the sim_video APIs to get keyboard activity in your window from the SDL event system...

**Comment by deltecent** [comment]
I've now tested this on the Mac and Windows with and without video support. All looks good.

**Comment by markpizz** [comment]
Where are you going when adding AltairZ80V.vcproj to the "Visual Studio Projects" directory?

All of what is in AltairZ80V.vcproj should be in AltairZ80.vcproj except the string ```Name="AltairZ80V"``` should be ```Name="AltairZ80"```

**Comment by markpizz** [comment]
And AltairZ80V.vcproj should not exist

**Comment by deltecent** [comment]
>And AltairZ80V.vcproj should not exist
There shouldn't be a AltairZ80 with and without Video support?

**Comment by markpizz** [comment]
> > And AltairZ80V.vcproj should not exist
> There shouldn't be a AltairZ80 with and without Video support?

All simulators which have devices capable of video are built on Windows without issue.  The components needed to do this are specifically provided.

Notice that the only reason the makefile might not build the altairz80 binary with video support would be if the needed libraries weren't available in the build environment.  There is no separate altairz80v target, nor separate targets for any other video capable simulators.

Notice that

**Comment by markpizz** [comment]
To be consistent with the rest of the simh project, the sol20.c and s100_vdm1.c files should have CRLF line endings.

**Comment by deltecent** [comment]
That’s what they’re supposed to have. I will see how those could have changed and redo.

**Comment by deltecent** [comment]
I looks like they do have CRLF endings:
```
00000000: 2f2a 2a2a 2073 6f6c 3230 2e63 3a20 5072  /*** sol20.c: Pr
00000010: 6f63 6573 736f 7220 5465 6368 6e6f 6c6f  ocessor Technolo
00000020: 6779 2053 6f6c 2d32 300d 0a20 0d0a 2020  gy Sol-20.. ..  
```
I'm not sure how they are being uploaded without CRLF. I don't think I've done anything different.

**Comment by markpizz** [comment]
> I'm not sure how they are being uploaded without CRLF. I don't think I've done anything different.

These are newly created files, which by default get created with LF line endings.

This problem relates to the mmd.c issue that previously got fixed.

I have:
```
$ git config -system core.autocrlf false
```
which means checkout-as-is and commit-as-is and don't mess with converting line endings.  And thus the files stored in the repo and the files extracted on check-out will both be the same.

**Comment by deltecent** [comment]
@markpizz I'm not following you. Are you seeing line ending issues with the files that are part of this PR (`s100_vdm1.c` and `sol20.c`) or `mmd.c` which has nothing to do with this PR? I am going to add your config to my config. I typically run `set ff=dos` in vi, but sometimes forget. I'm trying to remember to do `unix2dos` before pushing anything.

Maybe we should eradicate Windows from the planet and then we can all live in harmony. Think of all the bytes of storage that would be saved without the need to store all those unnecessary carriage returns. It should reduce our carbon footprint by at least 15%. Elon?

**Comment by deltecent** [comment]
> `git config -system core.autocrlf false`

Is this something that could/should be added to `.git\config` in this project?

**Comment by markpizz** [comment]
I tried changing the --local config which didn't end up as something that could be committed or that would have been done long ago.

I have it set in system since I end up creating new clones whenever I look at someone's PR and I'd have to do the config change ALL THE TIME, which would be a pain.

**Comment by deltecent** [comment]
> I have it set in system since I end up creating new clones whenever I look at someone's PR and I'd have to do the config change ALL THE TIME, which would be a pain.

Sounds like a good plan to follow. It's difficult enough getting PRs through (not a complaint, I really do benefit from other people reviewing PRs, finding problems and making recommendations) without formatting issues getting in the way.

Have you tried using hooks to run `dos2unix` and `unix2dos` on checkouts, commits, and amends? I am considering trying it unless you've already been down that path.

**Comment by markpizz** [comment]
> I'm not following you. Are you seeing line ending issues with the files that are part of this PR (s100_vdm1.c and sol20.c) or mmd.c which has nothing to do with this PR? I am going to add your config to my config. I typically run set ff=dos in vi, but sometimes forget. I'm trying to remember to do unix2dos before pushing anything.

Those 2 new files are stored in your repo with LF line endings.  I mentioned the mmd.c only due to the confusion about line endings.  Maybe it was or was not relevant.
How you edit these new files and unix2dos doesn't change the default of how they are stored in the repo.

**Comment by markpizz** [comment]
> Have you tried using hooks to run dos2unix and unix2dos on checkouts, commits, and amends? I am considering trying it unless you've already been down that path.

Hooks were used by this project early on, not to manage line endings, but for something else.  It turned out that they weren't actually engaged by all git environments on all platforms with all versions of git, so they were abandoned and their functions were melded into the various build mechanisms.

**Comment by deltecent** [comment]
I just pushed again. Would you check and see if they look correct now? I didn't know I had control over how they're stored in the repo. I thought the repo would do conversion to what it wanted regardless of how I pushed them. Of course, I want to make sure I'm always pushing files with CRLF, but I seem to be missing the mark.

I think there may be issues with these two also:
```
AltairZ80/s100_djhdc.c
AltairZ80/s100_tdd.c
```

markpizz:
I won't be near my desk for about 5-6 hours.  I'll check then.

markpizz:
Things look good now.

deltecent:
> Things look good now.

Thank you.

PR #194, opened 2023-03-11
Add ".vimrc" to ".gitignore"

deltecent:


markpizz:
Since there are absolutely no .vimrc files in the simh repo, adding this serves no purpose.

In general, you shouldn't be storing unrelated files in a simh working directory,

deltecent:
It serves a security purpose. For people that want a .vimrc file specific for their work in simh, like setting file types to DOS, 4 character indents, etc, it prevents someone else from placing their own .vimrc in the repository and doing all kinds of nasty things to others on the next pull because Git will ignore it without a force.

hharte:
+1, I agree this is helpful.

markpizz:
If you want to have extra files (to support windows like this case), put them in the "Visual Studio Projects" directory which, already ignores everything except specific files that are manually added.

deltecent:
> If you want to have extra files (to support windows like this case), put them in the "Visual Studio Projects" directory which, already ignores everything except specific files that are manually added.

I don't think `vim` works that way. Windows, being the "let's not put anything in any sort of standard location and scatter files all over the place where not even the operator knows where they are" operating system, it would make sense to pick any random place and shove files there. *nix systems have standards about where files go and where applications look for them, such as `vim` and `.vimrc`.

I don't see any harm with having `.vim` in `.gitignore`. What is the difference between `.vim` and `.Thumbs.db`, `*.bak` (which is in there twice, BTW), `*.save`, and `*.log`? What about `*.obj` and `*.exe` that have nothing to do with non-Windows systems?

In regards to this repository, what is the definition of "unrelated" and who came up with that definition?

>The purpose of gitignore files is to ensure that **certain files** not tracked by Git **remain untracked**.

deltecent:
>  It would be hard to imaging that folks doing simulator development have their home directory as the simh source directory.

Bingo! Yes, that would be hard to imagine.

~/.vimrc can be told to look for local a .vimrc in the current working directory. So, rather than having project-specific configurations in a single ~/.vimrc file, users can have a .vimrc file in their simh directory that is specific for that project. The issue is .vimrc is like a shell script and can do all sorts of bad things. Putting .vimrc in .gitignore would prevent someone from propagating their own .vimrc files to others.

The purpose of .gitignore is to let Git know which files it shouldn't track. It seems reasonable to place common files, like .vimrc, that may be used by a developer, but not tracked, in .gitignore as that's its intended purpose. This is more true when it improves security.

> Now that I'm beyond that stupidity

It's assumed and expected that all PRs will be subjected to some myopic and dogmatic scrutiny. It's the price of admission for contributing to `open-simh`.

markpizz:
It seems to me that the security concern would be in the original .vimrc which executes scripts from directories that might contain untrusted code which comes from elsewhere.

I wouldn't think that a someone's preferred .vimrc would ever make it into simh repo.  Who would approve such a PR?

Similarly, nothing would force your ~/.vimrc that executes other scripts possibly in the current directory to actually name these other scripts .vimrc so a .gitignore wouldn't protect you from anything.

It would seem that a better way to have a custom vim session when editing files in a simh working directory would be to have an alias for vim that invokes it with a -u argument that pointed to a simh specific .vmrc file that wasn't in the simh directory structure...

deltecent:
> It would seem that a better way to have a custom vim session when editing files in a simh working directory would be to have an alias for vim that invokes it with a -u argument that pointed to a simh specific .vmrc file that wasn't in the simh directory structure...

I disagree and think this PR should be added.

markpizz:
Think about this:

You are worried about security exposure which you've argued having .vimrc as an entry in the .gitignore file would protect you from such an exposure.

Actually, even if there was a specific mention of .vimrc in the .gitignore file, someone could commit a .vimrc into the repo merely by a "git add -f .vimrc" and then committing what ever it may contain.

With your .vimrc file which reads another .vimrc file from the current directory you would then execute whatever that person put in the .vimrc file when you happen to open any file in the simh root directory.  You will achieve the security protection you have described ONLY by making sure that the simh root directory repo does not contain a .vimrc file.  

You could enrich your current personal .vimrc file by having it check to see if scp.c is in the current directory and then check for the existence of a .vimrc file and if it exists, then complain and exit.

deltecent:
Mark, you are absolutely correct. The information was reading on protecting the use of .vimrc in a repository was incorrect. .gitignore is ignored on a pull request providing the user no protection at all.

pkoning2:
So what is the conclusion?
Mark's point is that .gitignore only prevents accidental adding of a file, not intentional adding.  True.  Intentional adding can be caught by PR review, and for this sort of file presumably would be.

PR #199, opened 2023-03-28
sigma: Correct recognition of pending DP seek interrupt.

kenrector:
This corrects an error that caused SIO reject when SIO occured before a pending seek interrupt on a different device.
 - Move the interrupt pending test from sigma_io.c into each device.
 - Make the sigma_dp.c test a special case that looks for pending seek interrupts.


kenrector:
Correct recognition of pending DP seek interrupt.  Moved interrupt pending test from sigma_io into individual device modules, except sigma_dp.c.  Added code in sigma_dp.c to test for pending seek interrupts instead.

PR #202, opened 2023-03-29
Initial support for the Nord-100,  implements the base instruction set.

ragge0:
Passes the test program INSTRUCTION-B.


markpizz:
Comments:
1) All source files have LF line endings and tabs.  simh project convention is that text files have CRLF line endings and tabs expanded to spaces.  The forcing of CRLF line endings means that CRLF are the line endings stored in the repo which is not the default.   You can get git to store the line endings that are in the local files by ```$ git config -system core.autocrlf false```.  You can change the local files to use CRLF line endings with ```$ unix2dos ND100/*```.  To change the tabs to spaces ```$ “expand -t4``` might be helpful...
2) simh project commit comments have the form:
```
affected-component: 1 line summary of commit < 80 characters
<blank-line-if-additional-commit-comments>
<additional-commit-comments with lines less than 80>
```
So for this commit a suggestion would be:
```
ND100: Initial support for Nord-100, implements the base instruction set.

Passes the test program  INSTRUCTION-B.
```
3) The nd100.h files defines a number of symbols which start with SCPE_.  These should start with STOP_ and the code which references them should be changed accordingly.  nd100_sys.c correctly defines text for these STOP messages in the sim_stop_messages array.
4) In nd100_sys.c sim_load routine uses uint16_t for the variable s rather than uint16.
5) In nd100_mm.c the wrmem function sometimes returns a value, but it has a signature of void wrmem(int vaddr, int val).  Both here and in nd100.h
6) In nd100_cpu.c, there is an include of <unistd.h> which needs to be removed.
7) In nd100_floppy.c the function alloca() is used which doesn't exist on all platforms.  malloc() should be used and then the allocated memory should be freed as needed on the return paths.
8) In nd100_cpu.c uses an function ffs() which isn't doesn't exist everywhere. I'd suggest that a local implementation be provided and used.
9) In nd100_cpu.c, various references to the m3 64bit variable assign results into 16bit variables.  Appropriate casts should be used to explicitly declare the intention and avoid truncation warnings.
10) In nd100_cpu.c, line 829: ```if ((regSTL & STS_K) != 0 ^ ((R[rd] >> n) & 1) != 0)```  has potential operator precedence ambiguity with the ^ operator, The compiler suggests using parentheses to avoid this ambiguity.
11) It's not clear what the code controlled by TIMERHACK does, but SCP provides significant timer/clock related facilities which might be appropriately used.
12) In nd100_floppy.c there is a call to fread() which should be changed to sim_fread().
13) in nd100_cpu.c there are a good number of unused variables.  GCC build like this to see those:
```$ make nd100 "GCC=gcc -std=gnu99 -U__STRICT_ANSI__  -O2 -finline-functions -fgcse-after-reload -fpredictive-commoning -fipa-cp-clone -fno-unsafe-loop-optimizations -fno-strict-overflow -I . -Werror -D_GNU_SOURCE -DUSE_READER_THREAD -DSIM_ASYNCH_IO  -DHAVE_PCRE_H -DHAVE_LIBEDIT -DHAVE_SEMAPHORE -DHAVE_SYS_IOCTL -DHAVE_LINUX_CDROM -DSIM_HAVE_DLOPEN=so -DHAVE_UTIME -DHAVE_LIBPNG -DHAVE_ZLIB -DHAVE_GLOB -DHAVE_SHM_OPEN -Wunused-variable"```
14) Your simulator specific include file is named nd100.h, while most others name this include file simname_defs.h.  No problem with this, but maybe a rename to follow convention - your choice.
15) Here is a zip file ([Visual Studio Projects.zip](https://github.com/open-simh/simh/files/11105272/Visual.Studio.Projects.zip)) that contains the promised windows build components which should be dropped in the "Visual Studio Projects" directory.

You'll probably commit a series of changes to address the above issues probably as separate commits.   Once you have some or all of these addressed you should be able to ```$ git push -f``` your current state to update the PR so we can see it.  Once all of these and any new issues are addressed, we'll want a single squashed commit to merge, but we'll get to that detail later.

**Comment by ragge0** [comment]
Den 2023-03-29 kl. 22:35, skrev Mark Pizzolato:
>
> Comments:
>
>  1. All source files have LF line endings and tabs. simh project
>     convention is that text files have CRLF line endings and tabs
>     expanded to spaces. The forcing of CRLF line endings means that
>     CRLF are the line endings stored in the repo which is not the
>     default. You can get git to store the line endings that are in the
>     local files by |$ git config -system core.autocrlf false|. You can
>     change the local files to use CRLF line endings with |$ unix2dos
>     ND100/*|. To change the tabs to spaces |$ “expand -t4| might be
>     helpful...
>
Ok, I'll try to change this.  I'm grown up with always use tabs in 
source files, hence "unexpand" is usually a great friend of me.
Also, file lines ending with crlf looks terrible in vi (with ^M at the 
end of all lines...).
(IMHO the project defaults should be changed, but that is another issue).

>  1. simh project commit comments have the form:
>
> |affected-component: 1 line summary of commit < 80 characters 
> <blank-line-if-additional-commit-comments> <additional-commit-comments 
> with lines less than 80> |
>
> So for this commit a suggestion would be:
>
> |ND100: Initial support for Nord-100, implements the base instruction 
> set. Passes the test program INSTRUCTION-B. |
Thanks, will do!
>
>  3. The nd100.h files defines a number of symbols which start with
>     SCPE_. These should start with STOP_ and the code which references
>     them should be changed accordingly. nd100_sys.c correctly defines
>     text for these STOP messages in the sim_stop_messages array.
>
FIxed.
>
>  3. In nd100_sys.c sim_load routine uses uint16_t for the variable s
>     rather than uint16.
>
Mistake, fixed.
>
>  3. In nd100_mm.c the wrmem function sometimes returns a value, but it
>     has a signature of void wrmem(int vaddr, int val). Both here and
>     in nd100.h
>
No, it doesn't.  It returns void (and shadowwr() returns void) which is 
perfectly legal.
>
>  3. In nd100_cpu.c, there is an include of <unistd.h> which needs to
>     be removed.
>
Fixed.
>
>  3. In nd100_floppy.c the function alloca() is used which doesn't
>     exist on all platforms. malloc() should be used and then the
>     allocated memory should be freed as needed on the return paths.
>
True, fixed!
>
>  3. In nd100_cpu.c uses an function ffs() which isn't doesn't exist
>     everywhere. I'd suggest that a local implementation be provided
>     and used.
>
Thanks, replaced with just a loop.
>
>  3. In nd100_cpu.c, various references to the m3 64bit variable assign
>     results into 16bit variables. Appropriate casts should be used to
>     explicitly declare the intention and avoid truncation warnings.
>
Added.
>
>  3. In nd100_cpu.c, line 829: |if ((regSTL & STS_K) != 0 ^ ((R[rd] >>
>     n) & 1) != 0)| has potential operator precedence ambiguity with
>     the ^ operator, The compiler suggests using parentheses to avoid
>     this ambiguity.
>
Correct, my mistake.
>
>  3. It's not clear what the code controlled by TIMERHACK does, but SCP
>     provides significant timer/clock related facilities which might be
>     appropriately used.
>
This is for timing debug of test programs that count instruction 
execution time.  Need to be able to do strange things when finding out 
how binaries work.  Commented out.
>
>  3. In nd100_floppy.c there is a call to fread() which should be
>     changed to sim_fread().
>
Fixed, thanks.
>
>  3. in nd100_cpu.c there are a good number of unused variables. GCC
>     build like this to see those:
>     |$ make nd100 "GCC=gcc -std=gnu99 -U__STRICT_ANSI__ -O2
>     -finline-functions -fgcse-after-reload -fpredictive-commoning
>     -fipa-cp-clone -fno-unsafe-loop-optimizations -fno-strict-overflow
>     -I . -Werror -D_GNU_SOURCE -DUSE_READER_THREAD -DSIM_ASYNCH_IO
>     -DHAVE_PCRE_H -DHAVE_LIBEDIT -DHAVE_SEMAPHORE -DHAVE_SYS_IOCTL
>     -DHAVE_LINUX_CDROM -DSIM_HAVE_DLOPEN=so -DHAVE_UTIME -DHAVE_LIBPNG
>     -DHAVE_ZLIB -DHAVE_GLOB -DHAVE_SHM_OPEN -Wunused-variable"|
>
Don't have gcc around, I use llvm on OSX, and it screams really much 
(about the rest of the simh code) when trying to compile with any 
warning flags.
Anyway, I tried to remove some unused variables.

>  3. Your simulator specific include file is named nd100.h, while most
>     others name this include file simname_defs.h. No problem with
>     this, but maybe a rename to follow convention - your choice.
>
No problem, renamed!
>
>  3. Here is a zip file (Visual Studio Projects.zip
>     <https://github.com/open-simh/simh/files/11105272/Visual.Studio.Projects.zip>)
>     that contains the promised windows build components which should
>     be dropped in the "Visual Studio Projects" directory.
>
Great, thanks, added!

> You'll probably commit a series of changes to address the above issues 
> probably as separate commits. Once you have some or all of these 
> addressed you should be able to |$ git push -f| your current state to 
> update the PR so we can see it. Once all of these and any new issues 
> are addressed, we'll want a single squashed commit to merge, but we'll 
> get to that detail later.
>
Ok, I think I have managed to update the pull request now?
Or is there something I have missed?

-- Ragge

**Comment by pkoning2** [comment]
The handling of line endings is one of the places where Git clearly gets it wrong where Subversion gets it right.  Line endings are a client side concern, so generating the desired line endings at the client as Subversion does is the correct design and forcing it into the server (and forcing it to be the same for all the users of a repository) is wrong.  But that ship has sailed, unfortunately.

File: ND100/nd100_cpu.c [hunk 1]
```diff
@@ -1,1737 +1,1748 @@
...
+		break;
+
+	case 004: /* BSKP zero/one */
+	case 005: 
+		if (((R[rd] >> n) & 1) == BIT7(IR))
+			regP++;
+		break;
+
+	case 006: /* BSKP BCM K == ~bit */
+		if (((regSTL & STS_K) != 0) ^ (((R[rd] >> n) & 1) != 0))
```
pkoning2, line 825: Since `^` is a bit operator, not a Boolean operator, it might be clearer to write it as `!=`.  But ok either way.

ND100/nd100_cpu.c
```diff
@@ -1,1737 +1,1748 @@
...
+	int sh;
+
+	sh = (regT & 077777) - 16384 + SEXT8(ins);
+	if (sh < 0) {
+		val = regA;
+		val >>= -sh;
+	} else {
+		
+	}
+#ifdef notyet
```
pkoning2, line 943: Why "not yet"?  When and why would this be enabled?

markpizz:
>> Comments: 1. All source files have LF line endings and tabs. simh project convention is that text files have CRLF line endings and tabs expanded to spaces. The forcing of CRLF line endings means that CRLF are the line endings stored in the repo which is not the default. You can get git to store the line endings that are in the local files by |$ git config -system core.autocrlf false|. You can change the local files to use CRLF line endings with |$ unix2dos ND100/*|. To change the tabs to spaces |$ “expand -t4| might be helpful...
>
> Ok, I'll try to change this.  I'm grown up with always use tabs in source files, hence "unexpand" is usually a great friend of me. Also, file lines ending with crlf looks terrible in vi (with ^M at the end of all lines...). (IMHO the project defaults should be changed, but that is another issue).

Line endings changed, but tabs are still present.  

BTW, on my macOS system, vi is really vim and it doesn't show ^M for files with CRLF line endings on any of your source code files...

>> 3. In nd100_mm.c the wrmem function sometimes returns a value, but it has a signature of void wrmem(int vaddr, int val). Both here and in nd100.h
>
> No, it doesn't.  It returns void (and shadowwr() returns void) which is perfectly legal.

Well, we have:
```
void
wrmem(int vaddr, int val/* , int how*/)
{
	vaddr &= 0177777;
	if ((vaddr >= 0177000) && is_shadow(vaddr))
		return shadowwr(vaddr, val);
	PM[vaddr] = val;
}
```

Surely shadowwr() is declared as returning void, but the syntax "return shadowwr(vaddr, val);" suggests otherwise and produces a warning on some platforms.  
```
void
wrmem(int vaddr, int val/* , int how*/)
{
	vaddr &= 0177777;
	if ((vaddr >= 0177000) && is_shadow(vaddr)) {
               shadowwr(vaddr, val);
		return;
                }
	PM[vaddr] = val;
}
```
Doesn't.

----
Besides the wrmem() issue, I missed one thing before:
In nd100_cpu.c on line 948, we have:
```
        val = -val;
```
where val is uint32, and we get a warning that says: unary minus operator applied to unsigned type, result still unsigned
Maybe:
```
        val = ~val + 1;
```
would produce the same and avoid the warning...

pkoning2:
I tried `return foo(x)` where `foo` is a void function.  GCC accepts it even with `-Wall` but objects with `-Wpedantic`.  So that tells me (without actually inspecting the standard) that using a void function in a return statement expression like that is not allowed by the C standard.
It would be a good idea to fix the place(s) where we do this.

markpizz:
Two things:
1) nd100_mm.c still has tabs
2) After fixing 1, your repo has 3 (or 4) commits for this PR.  These should be squashed...

ragge0:
Den 2023-03-30 kl. 19:15, skrev Mark Pizzolato:
>
> Two things:
>
>  1. nd100_mm.c still has tabs
>
Gee, this seems to be a forever problem.  Fixed now.
>
>  1. After fixing 1, your repo has 3 (or 4) commits for this PR. These
>     should be squashed...
>
Much googling told me how.  Hopefully fixed now :-)

-- R

markpizz:
From my point of view, this is ready for merge.

At some point I'd like to understand what your TIMERHACK stuff is/was doing, but that's not needed now.

pkoning2:
Thanks!

Ragge, you might send an announcement to the list about the new simulator.

Issue #201, opened 2023-03-29
pdp11_dh.c doesn't correctly handle byte-wide access to its registers


bjorend:
- #### Context
I found this problem when trying to run Unix v7.  Its DH-11 driver does do byte-wide access to the "CSR" or "SCR" register (from [https://www.tuhs.org/cgi-bin/utree.pl?file=V7/usr/sys/dev/dh.c]):
```
struct device
{
	union {
		int	dhcsr;
		char	dhcsrl;
	} un;
	int	dhnxch;
	int	dhlpr;
	char	*dhcar;
	int	dhbcr;
	int	dhbar;
	int	dhbreak;
	int	dhsilo;
};
 ````

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
```
sim> show version
PDP-11 simulator Open SIMH V4.1-0 Current
    Simulator Framework Capabilities:
        32b data
        32b addresses
        Threaded Ethernet Packet transports:PCAP:VDE:NAT:UDP
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        RAW disk and CD/DVD ROM support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC Apple LLVM 14.0.0 (clang-1400.0.29.202)
        Simulator Compiled as C (Release Build) on Mar 28 2023 at 20:17:21
        Build Tool: simh-makefile
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: SDL Version 2.26.3, PNG Version 1.6.39, zlib: 1.2.13
        PCRE RegEx (Version 8.45 2021-06-15) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        Ethernet packet info: libpcap version 1.10.1
        OS: Darwin titus.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000 arm64
        Processor Name: Apple M1 Pro
        tar tool: bsdtar 3.5.3 - libarchive 3.5.3 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.8
[...]

PR #204, opened 2023-03-30
PDP11: Support byte write access to DHCSR.

pkoning2:
This fixes issue #201, using a fix proposed bu Bjoren Davis.


larsbrinkhoff:
Approved.

PR #209, opened 2023-04-04
sigma: MT: AIO must mask unit number before calling TDV status.

kenrector:
sigma_mt.c - AIO must mask unit number before calling TDV status.


markpizz:
This is OK to merge.

PR #211, opened 2023-04-05
AltairZ80: Remove Mac OS9 support, clean up function declarations

psco:
This is purely code clean up, no change in functionality


markpizz:
This looks good to me.

PR #213, opened 2023-04-07
SCP: Fix possible memory corruption issue

psco:
The original statement
`argv[argc + 1] = NULL;`
was addressing a memory location beyond the bounds of `argv`.


markpizz:
Good to merge.

pkoning2:
Done.
The (existing) code shows the confusion seen from time to time, between NULL (a standard macro intended to represent a null pointer) and the ASCII character NUL, a.k.a., `'\0'` (the C string terminator).

PR #219, opened 2023-04-20
AltairZ80: Add additional checks for stop_cpu

psco:
There are two additional conditions (SIO attached to a file, SIO attached to a port) where a check for stop_cpu is required. This is needed when polling is needed to receive the WRU character.

### DIFF
--- a/3B2/3b2_timer.c
+++ b/3B2/3b2_timer.c
@@ -92,9 +92,13 @@
 #define QUICK_DELAY 100
 #endif
 
-#define DELAY_US(C,N) ((TIMER_MODE(C) == 3) ?                \
-                       (TIME_BASE[(N)] * (C)->divider) / 2 : \
-                       TIME_BASE[(N)] * (C)->divider)
+#define MIN_US 100
+
+#define CALC_US(C,N) ((TIMER_MODE(C) == 3) ?      \
+                      (TIME_BASE[(N)] * (C)->divider) / 2 : \
+                      TIME_BASE[(N)] * (C)->divider)
+
+#define DELAY_US(C,N) (MAX(MIN_US, CALC_US((C),(N))))
 
 #if defined(REV3)
 /* Microseconds per step (Version 3 system board):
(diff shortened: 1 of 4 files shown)


markpizz:
Each line that looks like this:
```
if ((ch == SCPE_OK) && stop_cpu) {
```
Would be more simply stated as:
```
if (ch == SCPE_STOP) {
```

psco:
Unless  `SCPE_OK == SCPE_STOP` the two conditions above do not look equivalent. Furthermore I do not see that `SCPE_STOP` is an explicit return value of `t_stat sim_poll_kbd (void)`.

markpizz:
You are right.  I stand corrected.

Meanwhile, what about the loop in altairz80_cpu.c  sim_instr() which currently says:
```
        do {
            result = (chiptype == CHIP_TYPE_8086) ? sim_instr_8086() : sim_instr_mmu();
        } while (switch_cpu_now == FALSE);
```
and should say:
```
        do {
            result = (chiptype == CHIP_TYPE_8086) ? sim_instr_8086() : sim_instr_mmu();
        } while ((switch_cpu_now == FALSE) && (result == SCPE_OK));
```

psco:
I would leave it as is: Normally this `do {} while` is executed exactly once since `switch_cpu_now` is almost always true (it is easy to misread this loop as an infinite loop requiring some additional guard). The condition `switch_cpu_now == false` happens only in the (rare) event that the CPU type is switched programmatically but in this case the result is also set to `SCPE_OK`, so the additional condition `result == SCPE_OK` is also true. The only problem would be if it were possible to create a program which would switch from a 8080/Z80 to an 8086 and then immediately back (or vice versa) but I do not believe this is possible.

markpizz:
As you say, ```normally...```, but it would seem that covering the potential for one of those dispatched to routines returning something otherwise unexpected should be covered.  

What would the downside of changing that to cover an unexpected return value be?

psco:
Probably I was not clear enough: `while (switch_cpu_now == FALSE);` is here equivalent to `while ((switch_cpu_now == FALSE) && (result == SCPE_OK));` As I wrote before: `switch_cpu_now == FALSE` implies `result == SCPE_OK`. I do not see the added value of adding this redundant condition. And if `switch_cpu_now == TRUE` the loop will immediately terminate and the the expected or unexpected result value will be returned.

psco:
This PR also solves this issue: [AltairZ80: ^E doesn't work when attaching SIO to Telnet (port 23)](https://github.com/simh/simh/issues/1172).

markpizz:
This is good to merge.

PR #227, opened 2023-05-10
sigma implement WAIT instruction

kenrector:
Added code to implement WAIT operation in sigma_cpu.c and correct the register returned for SIO, TIO, and other io instructions


pkoning2:
Does this enable SIMH idling when in the WAIT instruction, as it does on some other machines that have such an instruction (like PDP11)?

PR #228, opened 2023-05-10
AltairZ80: Code cleanup and some simplifications

psco:


psco:
Corrected formatting, removal of unnecessary includes, some function simplifications

psco:
It could be done without the platform check but I want to clearly point out that this polling is only needed on Windows or when the video functionality is enabled. No need to do an unnecessary poll otherwise. Furthermore there are checks for _WIN32 in other simulators as well.

markpizz:
> It could be done without the platform check but I want to clearly point out that this polling is only needed on Windows or when the video functionality is enabled. No need to do an unnecessary poll otherwise. 

Those are completely correct ideas that are properly considered in the SCP layer routines, not in simulator code.

> Furthermore there are checks for _WIN32 in other simulators as well.

Well, this is true:
1)  for the PDP11 where the help text for the ethernet devices provides an example of the the raw windows device names and the required external components for network support on Windows.
2) for the 3B2, similar help text language for ethernet devices.
3) for the IBM1130 there is a "grandfathered" windows specific GUI based front panel
4) the SAGE simulator has platform includes in several modules which aren't needed at all (compiles fine without them).  I've submitted this change as a PR today.
5) The HP2100 has some logic, but the HP2100 simulator is not supported in open-simh/simh or simh/simh but directly by Dave Bryan.  He changed various things about the SCP APIs he uses and thus his current codebase is no longer relevant to the 4.x codebase and beyond.  When running either the HP2100 or the HP3000 simulators, a message saying ```"This is the last version of this simulator which is API compatible with the 4.x version of the simh framework.  A supported version of this simulator can be found at: http://simh.trailing-edge.com/hp"``` appears before the initial ```sim>``` prompt.

So, none of these other cases relates to simulator device logic with platform \#ifdef's

markpizz:
While you're removing unnecessary includes, the include of sim_sock.h in altairz80_sio.c can be removed.

psco:
Thank you @markpizz ! Done.

hharte (approved):

psco:
@pkoning2 I believe this is now good to go. Otherwise, please let me know.

psco:
Will restart from a fresh repository.

PR #229, opened 2023-05-10
Fix library-related compile errors.

larsbrinkhoff:
- libpng is fussy about setjmp.h.

This happens on Ubuntu 16.  My other machines are on Ubuntu 22, where is works fine.

- Older versions of libsdl don't have SDL_SetWindowResizable.

This API was introduced with SDL 2.0.5.

### DIFF
--- a/sim_video.c
+++ b/sim_video.c
@@ -27,6 +27,9 @@
    11-Jun-2013  MB      First version
 */
 
+#if defined(HAVE_LIBPNG) && defined(USE_SIM_VIDEO) && defined(HAVE_LIBSDL)
+#include <png.h>
+#endif
 #include "sim_video.h"
 #include "scp.h"
 
@@ -148,7 +151,6 @@ static char tmp_key_name[40];
  * http://www.libpng.org/pub/png/src/libpng-LICENSE.txt
  */
 #include <SDL.h>
-#include <png.h>
 #include <zlib.h>
 
 #define SUCCESS 0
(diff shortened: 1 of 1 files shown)


larsbrinkhoff:
The problem is that the libpng distributed with some operating systems (such as Ubuntu 16) has this:

```
#      ifdef _SETJMP_H
       /* If you encounter a compiler error here, see the explanation
        * near the end of INSTALL.
        */
           __pngconf.h__ in libpng already includes setjmp.h;
           __dont__ include it again.;
```

larsbrinkhoff:
There's also this which suggests using one of the symbols so skip the check:

```
#ifdef PNG_SETJMP_SUPPORTED
/* This is an attempt to force a single setjmp behaviour on Linux.  If
 * the X config stuff didn't define _BSD_SOURCE we wouldn't need this.
 *
 * You can bypass this test if you know that your application uses exactly
 * the same setjmp.h that was included when libpng was built.  Only define
 * PNG_SKIP_SETJMP_CHECK while building your application, prior to the
 * application's '#include "png.h"'. Don't define PNG_SKIP_SETJMP_CHECK
 * while building a separate libpng library for general use.
 */

#  ifndef PNG_SKIP_SETJMP_CHECK
```

larsbrinkhoff:
I changed to Mark's proposed fix.

pkoning2:
So that fixes it?  Amazing.
I agree with Mark, it doesn't make any sense for libpng to be doing any of this.

shattered:
>     Older versions of libsdl don't have SDL_SetWindowResizable.
> This API was introduced with SDL 2.0.5.

SDL_WINDOW_RESIZABLE flag is an enum, so this #ifdef test shouldn't work.  I propose to delete it, since SDL 2.0.x are likely obsolete by now.

PR #237, opened 2023-05-27
AltairZ80: Migrate NET device from raw socket APIs to TMXR APIs

markpizz:
The use of RAW socket APIs burden's simulator DEVICE code with potentially complex interactions with the SCP event system.  Given this complexity, new simulator code is required to use the TMXR APIs to leverage TCP behaviors since these APIs have been written to be well behaved with respect to the SCP event system.

This module, written long before TMXR exposed robust TCP support, was the only module in the base simulators that directly used the sim_sock APIs.  This change addresses that.

### DIFF
--- a/AltairZ80/altairz80_net.c
+++ b/AltairZ80/altairz80_net.c
@@ -25,7 +25,7 @@
 */
 
 #include "altairz80_defs.h"
-#include "sim_sock.h"
+#include "sim_tmxr.h"
 
 /* Debug flags */
 #define ACCEPT_MSG  (1 << 0)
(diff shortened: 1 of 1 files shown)


markpizz:
The failure of the CI build on xcode is at least odd.  It builds and tests cleanly everywhere else.  

I have no way to try and reproduce this failure.  

Suggestions?

markpizz:
Deep thinking while walking the dog identified the problem.  Analysis was base on 2 facts.  Basic tests for this simulator involved only 2 things in the simulator:  1) Reset all devices, and 2) analyze registers in all devices.  Register analysis was clean, so something in the reset logic for the modified device (NET) had to be a problem.  There were 3 possible device address combinations and only 2 lines.  Inadvertent reference to 3 possible lines was the problem.

markpizz:
The current CI build failure is failing in the IBM1130 test for a windows build.  This has nothing to do with anything in the AltairZ80.

The CI build author should look into solving the CI build issue.

psco:
Unfortunately this change removes some existing functionality and introduces a **major defect**.

1. Unwanted removal of NET device debug flags ACCEPT and DROP

The expected behavior with 

```
set net debug
set debug stdout
```

for cpnetclient and cpnetserver is:

Once the user executes the login command in the client, a message similar to this one shows up on the CP/NET server side
`DBG(7520367)> NET ACCEPT: NET: [0x0000ada3] Accepted connection 1 with socket 11.`

When the user terminates the server with ^E and BYE, a message similar to this one shows up on the CP/NET client side
`DBG(1815000)> NET DROP: NET: [0x0000f3fa] Drop connection 0 with socket 7`

Since there is no support for the debug flags ACCEPT and DROP anymore, these messages will not show although they are quite useful.

2. Incorrect behavior in case of a third CP/NET or CPNOS client

When attempting to connect a third CP/NET client the expected behavior is that the execution of the third "login" command will block until one of the other CP/NET clients is terminated with ^E and BYE. As soon as one of these clients is terminated, the blocked "login" command will continue. The new version will start continuously showing in the third CP/NET client
`data dropped (no connection) 5 to 51`
once the "login" command is attempted. This confusing "error message" is incorrect as there is no error in the simulated system. Furthermore, terminating one of the other CP/NET clients will not cause the blocked "login" command to continue.

A similar problem exists with CPNOS: In the third CPNOS client one can press RETURN after the "LOGIN=" prompt but this will block even when one of the other CPNOS clients is terminated. Note that the correct behavior (login will continue once a CPNOS client is terminated) is even described in the documentation in section 5.5.

This is a **major defect**.

3. The coding style for function calls in altairz80_net.c is `func()` and not `func ()`.

4. The comment in line 66 has not been updated.

5. There is a trailing "\\" on line 182.

1 and 2 are show stoppers btw.

pkoning2:
Could we get the issue sorted out and a finished PR generated?

psco:
I have no interest in this PR and the current proposal just introduces two major regression defects. I propose to close it.

markpizz:
Please leave this may as a draft.  I've been busy with other things to finish  addressing what Peter pointed out.

pkoning2:
You can reopen it, right?  I'm not sure how to mark a PR as draft.

markpizz:
I'll reopen it when I'm back on it.
I saw a way to mark it as a draft earlier before I got otherwise busy.

Issue #235, opened 2023-05-19
Setup ASAN, UBSAN, TSAN CI jobs


XVilka:
Address sanitizer, UB (Undefined Behavior) sanitizer, Thread sanitizer will be helpful for catching obvious issues in the code while running tests.

Something like this could help:
```cmake
if(CMAKE_BUILD_TYPE MATCHES "Debug")
  add_compile_options(-fsanitize=address)
  add_compile_options(-fsanitize=undefined)
  add_link_options(-fsanitize=address)
  add_link_options(-fsanitize=undefined)
endif()
```

bscottm:
For which compiler(s)?

XVilka:
@bscottm from my experience both modern GCC and Clang work well with sanitizers, there are some differences but largely they overlap in reporting the same issues.

bscottm:
I do think this is a good idea, with better scope

1.. Should this be done for any debug build or should sanitizers be run as separate rules in `build.yaml`?
2. Is there a simh project position on this proposal that guides (1)?

Scoping to GCC and Clang is easy by matching the CMake compiler ID, which exists in a couple of places already. All Debug vs. separate CI/CD builds has the potential for unexpected consequences (not counting the usual Pizz on things.)

XVilka:
1. It's better to make it a separate job, from my experience, unless you have the reason to save CI workload/minutes
2. no idea

bscottm:
@XVilka: I copied sanitizer support from `arsenm/sanitizers-cmake`, which supports GCC, Clang and MSVC. You will have to set the configuration option flags from the command line. These are summarized in `Sanitizers-README.md` (renamed `README.md` from `arsenm/santizers-cmake`.) The files are all licensed with the MIT license, so they ought to be compatible with `open-simh` and `simh`.

I had to tweak the code to ensure that sanitization is only added for Debug build configurations. And then did a test run on my RPi ARM64 with `-DSANITIZE_ADDRESS=On`, which failed for all of the right reasons. Excerpt from `simh-vax-780` test:

```
[ctest] =================================================================
[ctest] ==2060569==ERROR: LeakSanitizer: detected memory leaks
[ctest] 
[ctest] Direct leak of 677 byte(s) in 20 object(s) allocated from:
[ctest]     #0 0x7fb38f2e7c in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
[...]

PR #239, opened 2023-06-07
3B2: Properly NULL-terminate IU mod arrays

sethm:
This addresses an issue raised in #235, two arrays in the 3B2 IU device were not properly null-terminated.


markpizz:
This is OK to merge.

pkoning2:
Merged.

PR #240, opened 2023-06-08
PDP11: RK11: Complete fix for WC as implemented in V3 + other fixes

al20878:
This is a revised PR#143 because it was never fully applied.  Still needs to fix the word count, and other things as previously posted.

### DIFF
--- a/PDP11/pdp11_rk.c
+++ b/PDP11/pdp11_rk.c
@@ -105,9 +105,9 @@
 #define RK_NUMTR        (RK_NUMCY * RK_NUMSF)           /* tracks/drive */
 #define RK_NUMDR        8                               /* drives/controller */
 #define RK_M_NUMDR      07
-#define RK_SIZE         (RK_NUMCY * RK_NUMSF * RK_NUMSC * RK_NUMWD)
+#define RK_NUMBL        (RK_NUMTR * RK_NUMSC)
+#define RK_SIZE         (RK_NUMBL * RK_NUMWD)           /* words/drive */
 #define RK_RSRVSEC      (3 * RK_NUMSF * RK_NUMSC)       /* reserved (unused) disk area */
-                                                        /* words/drive */
 #define RK_CTLI         1                               /* controller int */
 #define RK_SCPI(x)      (2u << (x))                     /* drive int */
 #define RK_MAXFR        (1 << 16)                       /* max transfer */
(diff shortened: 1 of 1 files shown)


pkoning2:
Merged.

al20878:
Somehow this commit (even though it appears in master) does not look proper -- red, as PR was rejected, not actually merged.  IDK it that's a github issue or something was amiss when the merge occurred.  I checked the other closed PR and the same thing shows up there as well.  I expect the resolution to be "merged" shown by github.

al20878:
@pkoning2  please take a look.  TY

markpizz:
The repo looks OK.  You as Author and @pkoning2 the committer.

Don't know about the web UI.

al20878:
OpenSimhPRs

al20878:
Only a few PRs appear as properly merged (purple). Most recent ones are just "closed", which is not a proper resolution, because it does not propagate as merged to the forked repo (branch), which then notifies that it can be dropped / deleted...  Something's amiss

markpizz:
Well, the CI builds have been broken for at least a few weeks.  This PR had one of the CI failures which had nothing to do with the content of this PR (hence @pkoning2 merged it).  How all of the magic pieces flowed together may indeed be broken, but what's in the repo matters.  Since the PR has been merged, you can delete the related branch in your personal repo.

al20878:
The CI is a check mark or a cross that follows the PR name, the resolution is what shows up front (in the pic that I attached).
There are PRs with broken CI (red cross) that are marked "merged" (purple), and there is one okay CI (green check) yet still red closed (instead of purple merged).  So CI is probably not the culprit.

markpizz:
I'm not sure that even if the CI builds were successful, the merge success status you're looking for would be there most of the time.
@pkoning2 I think this relates to how and when PR content is merged.  If users follow the previous required strategy of pull before committing and then immediately creating a PR, and then I pick up the user's changes for review and if OK, I merely pushed the user's branch (which has all of the master plus their commit) to the repo master.  That would then have the specific user's commit id now in the master branch (and the user would be both the Author and the committer).  It would then be easy for git clients (or the web UI) to know that the user's personal branch has been digested since the commit-id's will match.  Otherwise, some out of band bookkeeping would have to happen.  Maybe merging via the web UI would have the same results under the same conditions (pull before commit and merge before anything else comes along), or maybe merging via the web UI when PRs have collected and thus a real merge will be needed for all but possibly one resulting in new commits for these merging activities and you as the committer.

al20878:
>  Github uses a bad color choice for closed PRs.

Not only that but also it shows as an unmerged branch in the icon: 
ClosedUnmerged

PR #242, opened 2023-06-08
MAKEFILE: Move CYGWIN library search out of VIDEO section; sync compi…

al20878:
…ler flags with MSVC
This PR moves high-precision timer outside the VIDEO section for CYGWIN.  It also synchronizes _DEBUG/NDEBUG with MSVC (they are predefined in Windows builds, and NDEBUG is required for ANSI compatibility with <assert.h> if any used in code).


pkoning2:
Merged.

PR #243, opened 2023-06-08
SCP: Don't malloc ep->match_pattern twice

sethm:
`ep->match_pattern` was being allocated twice, leading to a memory leak, as mentioned in #235

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -13119,8 +13119,6 @@ else {
     ep->match = match_buf;
     ep->size = match_size;
     }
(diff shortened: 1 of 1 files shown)


markpizz (changes requested):
The correct solution is not to remove the 2 lines (13122 and 13123), rather: 13122 should merely be changed to:
ep->match_pattern = (char *)realloc (ep->match_pattern, strlen (match) + 1);

markpizz:
After closer review, the original change is correct.

pkoning2:
Merged.

Issue #245, opened 2023-06-08
PDP11: timer calibration division by zero


al20878:
- #### Context
PDP11 failed to start, core dumped.  This is a rather rare occurrence.  It happened right after the fresh build from github sources,
at the "post build" register check procedure...  The build was on CYGWIN with gcc-11.

More context:
I was building with `NOVIDEO=1` (and also: `DEBUG=1 NONETWORK=1`), which as it turns out, used to bypass the WINMM high-precision timer on Cygwin (see my PR#242), so it was not using that.  When built with WINMM, I could not reproduce this problem, so it only happened when the timer was "coarse" (default).

coredump should not occur -- division by 0 is very much avoidable if the terms are duly checked.
- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
latest build from github as of yesterday
- #### how you built the simulator or that you're using prebuilt binaries

```
LPATH=/usr/local/lib make DEBUG=1 NONETWORK=1 NOVIDEO=1 pdp11
```
- #### the simulator configuration file (or commands) which were used when the problem occurred.

none
- #### the expected behavior and the actual behavior

none
- #### you may also need to provide specific pointers to data files that may be necessary to demonstrate the problem
gdb output pulled post-mortem from the core file:
```
Division by 0:

1168        sim_os_tick_hz = 1000/(sim_os_clock_resoluton_ms * (sim_idle_rate_ms/sim_os_clock_resoluton_ms));
(gdb) p sim_os_tick_hz
$1 = 0
(gdb) p sim_os_clock_resoluton_ms
$2 = 10
(gdb) p sim_idle_rate_ms
$3 = 3
```
Please note the values of the parameters in the formula above (`sim_timer.c`).  Since 'sim_idle_rate_ms / sim_os_clock_resolution_ms` happened to be 0 (the first is smaller than the last), the denominator for division of the 1000 into was 0.  The host machine that was doing the build was quite busy at that time, so maybe that's how the values were "underestimated".  But at any rate, there should be a protection against such thing.

markpizz:
This comes down to if building with cygwin (or MinGW) works for you, great.  If it doesn't you should either build with a free Microsoft Compiler or use pre-built windows binaries.  Cygwin is a Posix emulation on top of Win32 or Win64 and as such may or may not provide precisely correct behavior in various corner cases (such as this).  A well behaved OS syscall layer will never produce the 0 case you are reporting about.

The value of building simh with these tool chains on Windows is essentially 0.

[...]

PR #248, opened 2023-06-08
TIMER: Properly report clock calibration inconsistency

markpizz:
This avoids a potential crash when timer calibration produces less than useful results.

As discussed in #245

### DIFF
--- a/sim_timer.c
+++ b/sim_timer.c
@@ -1083,7 +1083,7 @@ do {
         sim_os_clock_resoluton_ms = clock_diff;
     clock_last = clock_now;
     } while (clock_now < clock_start + 100);
-if ((sim_idle_rate_ms != 0) && (sim_os_clock_resoluton_ms != 0))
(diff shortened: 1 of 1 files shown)


al20878:
> This avoids a potential crash when timer calibration produces less than useful results.

No, it does not.  Please see #245.

This would:
```
if ((sim_idle_rate_ms != 0) && (sim_os_clock_resoluton_ms != 0) && !(sim_idle_rate_ms < sim_os_clock_resoluton_ms))
```

markpizz:
Thanks again.

al20878:
> Thanks again.

It still needs an "&&" not "||"

al20878:
Basically, that all boils down to this:
```
if (sim_os_clock_resoluton_ms != 0  &&  sim_idle_rate_ms >= sim_os_clock_resoluton_ms)
```
It covers all bases

al20878:
The third time is the charm!  Now it looks good.

pkoning2:
Merged.

PR #257, opened 2023-06-22
sigma: Correct DP Sense Length error(Bob Supnick)

kenrector:
sigma_dp.c:   SENSE operation is returning a length error when it shouldn't.
CP-V requests x10 (16) bytes and status returned from chan_WrMemB() is 0x8001, and
CP-V requests x4 bytes and returned the same status.

This fix accounts for the 'no errors on 10 byte model' and 'error if > 16'. A 0 transfer length is 
considered to be 65536, so it will always cause a length error in a 16 byte model.


pkoning2:
Merged.

PR #262, opened 2023-06-26
SCP: Assure that ASSURE() gets used instead of assert()

markpizz:


display/ng.c
```diff
@@ -123,7 +122,10 @@ int
 ng_init(void *dev, int debug)
 {
   /* Don't change this number. */
-  assert (DISPLAYS == 8);
```
pkoning2, line 126: This seems like a good place for STATIC_ASSERT.  Having a compile time invariant checked at runtime is not the best answer.

slirp/tftp.c
```diff
@@ -387,7 +387,10 @@ static void tftp_handle_rrq(Slirp *slirp, struct tftp_t *tp, int pktlen)
   }
 
   if (nb_options > 0) {
-      assert(nb_options <= ARRAY_SIZE(option_name));
+      if (nb_options > ARRAY_SIZE(option_name)) {
```
pkoning2, line 390: Why not simply change `assert` to `ASSURE` ?

display/ng.c
markpizz: assert() is a runtime functionality only on DEBUG builds and otherwise does nothing.

The suggested change keeps that effect.

I've changed this to a static check.

The other case was looking at runtime determined data, so a static check wouldn't be useful

This file and the only other place in the codebase where assert() was used doesn't specifically depend on any other simh code.

markpizz:
As I mentioned elsewhere, this externally source component (slirp) doesn't otherwise depend on any simh pieces, so it isn't built with sim_defs.h.  When I added NAT to sim_ether I created glue so that it could be leveraged by sim_ether essentially located in slirp_glue and made almost no modifications to the original slirp code.

pkoning2:
> As I mentioned elsewhere, this externally source component (slirp) doesn't otherwise depend on any simh pieces, so it isn't built with sim_defs.h. When I added NAT to sim_ether I created glue so that it could be leveraged by sim_ether essentially located in slirp_glue and made almost no modifications to the original slirp code.

Thanks, that makes sense.

al20878:
What's wrong with the standard `assert()`?

The haste with which this was created after the debate in PR244 suggests yet another `sim_disk.c`-like complex.

And no, `ASSURE` is not good as a replacement for `assert()`, and I already explained, why.

markpizz:
There is absolutely no haste here.  ASSURE has been around and in use in simh for 8 years.

The point of the PR is to avoid rehashing the discussion the next time this comes up.

al20878:
>  ASSURE has been around and in use in simh for 8 years.

That's totally fine!  But you're missing the point that `ASSURE` is **not a drop-in replacement** for `assert()`, as the two have very different semantics.  One can choose to use `ASSURE` if that fits the bill, but sometimes, it's `assert()`, which is really wanted.

markpizz:
Yes, it truly has different semantics.  

You deep explanation has not in any way explained what in your code will break if you replace assert() with ASSURE().  

If you want something like assert(), you can invent your own function, BUT it should NOT behave differently when compiled with or without DEBUG.

al20878:
It's pointless to discuss this any longer as you clearly don't understand the concept of the code invariant, which `assert()` were designed for.  They are for code development _only_, and they are a great way to comment your code without actually commenting your code, if you know what I mean.  And as they are not supposed to have any side effects, the code runs the same _with_ and _without_ them.  Yet they are a great aid when you're developing, to assure (no pun intended) that your assumptions remain correct.  As comments, they are wiped out when you're done, and build your code without debugging.

FWIW, your "protection" is easily bypassed by inclusion of `<assert.h>` after any simh headers.  Sad.

markpizz:
Please use comments to comment your code and ASSURE to validate reasonable or correct values.

al20878:
>  ASSURE to validate reasonable or correct values.

I don't need to "validate" the values.  They are all correct.  I'd rather `#define assert(x) /* nothing */` than add the ballast and pipeline-disturbing code that `ASSURE` would inject.  Thanks but no thanks.

markpizz:
> #define assert(x) /* nothing */ 

You could do that, but then anyone looking at the code might thing those lines of code actually do something.

Given your voicing deep performance concern, you might try to actually perform some testing with a PDP11 simulator using this code when compiled with ASSURE and with assert() using this disk device with as much robust activity as you can imagine.  Enumerate the amazing performance gain by not having the pipeline-disturbing code.  As I've said previously, I don't think you will be able to measure any difference, but I'll be glad to be proven wrong.

al20878:
> anyone looking at the code might thing those lines of code actually do something.

Those should start reading "The C Language" by K&R from page one.  I can't care less.

> Enumerate the amazing performance gain

I have better things to do.

sim_defs.h
```diff
@@ -1145,6 +1145,16 @@ struct MEMFILE {
                                                     sim_printf("%s failed at %s line %d\n", #_Expression, __FILE__, __LINE__);       \
                                                     abort();}
 
+#if defined(assert)
+#error "Don't user assert().  Feel free to use ASSURE(expression) in its place."
```
XVilka, line 1149: A typo, should be `use`

XVilka: Probably makes sense to reword into something like "It's advised to use ASSURE(expression) instead"

sim_defs.h
```diff
@@ -1145,6 +1145,16 @@ struct MEMFILE {
                                                     sim_printf("%s failed at %s line %d\n", #_Expression, __FILE__, __LINE__);       \
                                                     abort();}
 
+#if defined(assert)
+#error "Don't user assert().  Feel free to use ASSURE(expression) in its place."
+#else
+#define assert(_Expression) do {                                                                                                \
+                                fprintf (stderr, "Don't user assert().  Feel free to use ASSURE(expression) in its place.\n");  \
```
XVilka, line 1152: Same

XVilka (changes requested):

sim_defs.h
markpizz: Good Catch.  Fixed as suggested

al20878:
> reword into something like "It's advised to use ASSURE(expression) instead"

@XVilka : This hypocrisy does not change that from being an ultimatum, as coded

pkoning2:
Thanks for the proposal, but after discussion we've decided not to take this change.  Instead we'll try to get an explicit coding standard written and published.

PR #263, opened 2023-06-28
FIO: Add a global variable that indicates the normal host path separator

markpizz:
This avoids the potential need for platform specific #ifdef blocks in simulator code which need/have awareness of host system file paths.


pkoning2:
But nothing uses this as far as I can see, right?

PR #264, opened 2023-06-30
AltairZ80: Correct spelling, formatting and years

psco:


pkoning2:
I see a bunch of files with just a copyright date change.  Did you actually change those in earlier PRs?  If not, please don't do that.  I know some projects change the copyright year on every file every year.  In my view this is incorrect and a misunderstanding of how copyright works.  If you change a file,  by all means adjust the copyright date accordingly, but not if the file hasn't changed.

psco:
I'd like to update the copyright dates about every 10 years. I think this should be ok.

pkoning2:
It doesn't make a whole lot of practical difference because of the duration of copyright, but that really isn't right.  Copyright attaches to a "work", meaning something you created, or revised.  So a change comes with a copyright date matching that change.  But the elapsing of time doesn't change the effective date of the copyright.  In other words, no matter what the header may say, legally (as I, not a lawyer, understand it) the actual copyright dates are the creation and revision dates of the work in question.

psco:
If I understand you right you are opposed to just updating the copyright year if there is no change otherwise. If this is the case I'll make sure that all files where I just updated the year will also have some other change. Would this be acceptable?

markpizz:
As I understand the current copyright concepts, a copyright exists explicitly with the authorship of a work without any regard to explicit declaration/language claiming copyright.  Authorship, by itself, is all that is needed.

pkoning2:
> As I understand the current copyright concepts, a copyright exists explicitly with the authorship of a work without any regard to explicit declaration/language claiming copyright. Authorship, by itself, is all that is needed.

Yes, that's right.  So the copyright header is redundant, but serves a useful purpose in making things explicit.

PR #272, opened 2023-07-10
Memory sanitizer fixes

bscottm:
Fixes suggested by running the Clang memory santizer on the Github CI/CD pipeline:

- sim_exp_check: Use calloc() to allocate and initialize the PCRE captured string offset vector. Also check the offsets to ensure they are captured substrings:

  -- If a substring wasn't captured, the offset pair is { -1, -1 }. (Of course, this would never happen in SIMH. :-)

  -- Remove the corresponding "_EXPECT_MATCH_GROUP_{pat}" variable from the environment when its substring isn't captured.  unsetenv() is locally implemented in scp.c, which sets the environment variable to an empty string; it isn't actually removed.

  Failure occurs in multiple tests.

- _sim_debug_write_flush: Ensure that debug_line_buf_last's underlying buffer is initialized when realloc-ed. (i650 failure.)

- vid_version: Initialize local variables "compiled" and "running". MSan claims that these aren't initialized. Reading the SDL  source, the patch level might not be initialized on some platforms. (multiple sim_video tests.)

NOTE: Do not attempt to run interactive memory-santized simulators on Linux platforms where ncurses <= 6.2+20201114-2. MSan detects uninitialized variables within ncurses; newer versions of ncurses fix them.

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -13241,6 +13241,7 @@ for (i=0; i < exp->size; i++) {
     if (ep->switches & EXP_TYP_REGEX) {
 #if defined (USE_REGEX)
         int *ovector = NULL;
+        int ovector_elts;
         int rc;
         char *cbuf = (char *)exp->buf;
         static size_t sim_exp_match_sub_count = 0;
@@ -13259,27 +13260,38 @@ for (i=0; i < exp->size; i++) {
                 }
             }
         ++regex_checks;
-        ovector = (int *)malloc (3 * (ep->re_nsub + 1) * sizeof (*ovector));
+        ovector_elts = 3 * (ep->re_nsub + 1);
+        ovector = (int *)calloc ((size_t) ovector_elts, sizeof(*ovector));
         if (sim_deb && exp->dptr && (exp->dptr->dctrl & exp->dbit)) {
             char *estr = sim_encode_quoted_string (exp->buf, exp->buf_ins);
             sim_debug (exp->dbit, exp->dptr, "Checking String: %s\n", estr);
             sim_debug (exp->dbit, exp->dptr, "Against RegEx Match Rule: %s\n", ep->match_pattern);
             free (estr);
             }
-        rc = pcre_exec (ep->regex, NULL, cbuf, exp->buf_ins, 0, PCRE_NOTBOL, ovector, 3 * (ep->re_nsub + 1));
+        rc = pcre_exec (ep->regex, NULL, cbuf, exp->buf_ins, 0, PCRE_NOTBOL, ovector, ovector_elts);
         if (rc >= 0) {
             size_t j;
             char *buf = (char *)malloc (1 + exp->buf_ins);
 
             for (j=0; j < (size_t)rc; j++) {
                 char env_name[32];
+                int end_offs = ovector[2 * j + 1], start_offs = ovector[2 * j];
 
                 sprintf (env_name, "_EXPECT_MATCH_GROUP_%d", (int)j);
-                memcpy (buf, &cbuf[ovector[2 * j]], ovector[2 * j + 1] - ovector[2 * j]);
-                buf[ovector[2 * j + 1] - ovector[2 * j]] = '\0';
-                setenv (env_name, buf, 1);      /* Make the match and substrings available as environment variables */
-                sim_debug (exp->dbit, exp->dptr, "%s=%s\n", env_name, buf);
-                }
+                if (start_offs >= 0 && end_offs >= start_offs) {
+                    memcpy (buf, &cbuf[start_offs], end_offs - start_offs);
+                    buf[end_offs - start_offs] = '\0';
+                    setenv (env_name, buf, 1);      /* Make the match and substrings available as environment variables */
+                    sim_debug (exp->dbit, exp->dptr, "%s=%s\n", env_name, buf);
+                    }
+                else {
+                    /* Substring was not captured by regexp: remove from the environment
+                     * (unsetenv is local static -- doesn't actually remove the variable from
+                     * the environment, sets it to an empty string.) */
+                    sim_debug (exp->dbit, exp->dptr, "unsetenv %s\n", env_name);
+                    unsetenv(env_name);
+                    }
+                } 
             for (; j<sim_exp_match_sub_count; j++) {
                 char env_name[32];
 
@@ -13607,6 +13619,9 @@ if (debug_line_offset + len + 1 > debug_line_bufsize) {
     debug_line_bufsize += MAX(1024, debug_line_offset + len + 1);
     debug_line_buf = (char *)realloc (debug_line_buf, debug_line_bufsize);
     debug_line_buf_last = (char *)realloc (debug_line_buf_last, debug_line_bufsize);
+    /* MSan: Ensure debug_line_buf_last is initialized. If we realloc-ed it,
+     * then we really don't care what its contents are or were. */
+    memset(debug_line_buf_last, 0, debug_line_bufsize);
     }
 memcpy (&debug_line_buf[debug_line_offset], buf, len);
 debug_line_buf[debug_line_offset + len] = '\0';
--- a/sim_video.c
+++ b/sim_video.c
@@ -2235,7 +2235,7 @@ return 0;
 const char *vid_version(void)
 {
 static char SDLVersion[160];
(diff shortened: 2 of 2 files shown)


markpizz:
Several comments about this:

1) The change to _sim_debug_write_flush is wrong and serves no purpose.  The current contents of the memory currently pointed to by debug_line_buf_last is explicitly intended to be preserved (and it is a string which is null terminated).  The realloc() makes room to extend the buffer which will then be able to hold additional data which will be appended to the existing contents.

2) Comments in this PR text and explicitly in the proposed code change relating to unsetenv() are at best snarky and making any comment about "unsetenv() is locally implemented in scp.c, which sets the environment variable to an empty string; it isn't actually removed." how this makes any difference in any context (especially here in the regex match logic which is at best transitory).  An environment variable defined as an empty string will behave (in SCP references to environment variables) precisely equivalent to an undefined environment variable since both will expand to an empty string.

pkoning2:
I would think that pointing out that `unsetenv` which is a standard library function is here overridden to do something nonstandard is in fact useful.
Agreed on the realloc.  Scott, did you think realloc creates a new buffer with undefined contents?  It's documented as preserving the contents.

pkoning2:
I merged this, minus the realloc change.

markpizz:
> -- If a substring wasn't captured, the offset pair is { -1, -1 }. (Of course, this would never happen in SIMH. :-)

What tool, with what input, inspired the change to the results from pcre_exec?

The simh use of pcre_exec() requires all groups in the regex to match and DOES not allow for partial matching.  Partial matching would only happen if pcre_exec() were invoked with option flags that allowed partial matches and it is NOT called with such flags!

So, this change specifically protects against some theoretical bug in the pcre API.  Using this kind of thinking, all library calls should use imaginative guessing to see if the library call didn't do what it was supposed to do.  Such imagination would therefore require the calling code to implement the function of the library so that it could validate that the called library function actually did the right thing.  Imagine how big the program would get if this thinking was followed consistently to validate that each API called actually produced the right results.

bscottm:
@pkoning2: `realloc` copies the original content to the larger buffer, if the original pointer isn't `NULL`; otherwise, `realloc` acts like a glorified call to `malloc` when the original pointer is `NULL`. The Clang and gcc memory sanitizers detect that the region beyond the original content is not initialized, which is what's causing the the MSan fail.

@markpizz: It's more readable code inspired by reading the PCRE documentation (which explicitly explains the return offset array and values, of which `-1` is explicitly stated as a possible return value, BTW.) For code that's lightly commented, it might have been a good idea to add a comment explaining that the check for `-1` is unnecessary.

pkoning2:
You're right on how realloc works, but the point that the sanitizers missed is that, if the previous contents of the buffer are zero-terminated, so are the contents of the newly enlarged buffer.

bscottm:
@pkonig: The region past the copied original content isn't initialized. That's what MSan caught. Fairly petty on MSan's part, although a common cause of interesting bugs (even though the original contents are zero terminated, but MSan can't read SIMH's mind.)

markpizz:
> @bscottm The region past the copied original content isn't initialized. That's what MSan caught. Fairly petty on MSan's part, although a common cause of interesting bugs (even though the original contents are zero terminated, but MSan can't read SIMH's mind.)

Absolutely true, so it is YOUR REPONSIBILITY to carefully review the results of the "caught" problems and determine if what has been caught is a false positive rather than introducing bugs which merely satisfy the tool's report!

bscottm:
@markpizz: False positive failures won't make the test pass during CI/CD unless one is a fan of persistent CI/CD failures. So, they need to be fixed.

You can tell me that the failure is Bravo Sierra as much as you want. It simply doesn't address the failure. They were relatively easy fixes and closed a path leading to common memory allocation (mis|ab)use. What's your issue with making the code more robust or more maintainable?

markpizz:
@bscott If a fault detection tool produces false positives, the tool absolutely needs to be fixed OR there should be a way to ignore the known false positive problem(s).

What's your problem with making such tools more robust so they don't produce false positives?  False positives are failures in the tool not in the code that the tool analyzed.

The specific false positive cases in this Msan run that drove the PR, are: 
1) The potential return values for pcre match group entries some how magically having a broken pcre_exec that doesn't honor it's option flags.
2) The problem that said buffers which are realloc() need to be zeroed.

The potentially uninitialized code in vid_version() was an accurate claim!

If the tool in question produces false positives and has no way to specifically ignore know false positives, then it should be run manually every once in a while until the tool is fixed.  Note that this massive Msan error list reflects all of the errors in the codebase since the project started.  To some folks, running such a tool on every pushed commit might seem not particularly valuable.

Your suggestion that the pcre group entry code could use more comments explaining what was expected to happen wouldn't have caused the tool to not produce a false positive, right?

The PR was merged without the changing the false positive relating to the realloc issue.  Isn't that sufficient to cause repeated failures of the tool?

pkoning2:
Is there a way to tell the tool that something is a false positive?  Say, a comment that it recognizes (in the fashion of old style comments that "lint" would understand)?
If it insists on marking this as an error and can't be told to stop doing so, the answer is to stop using the tool.  It's nice to have checkers, but not if they give wrong answers and can't be told to stop doing so.

bscottm:
@pkoning2 Short answer is, "Not really."  The sanitizers stop at the first detected issue. All of the Clang and GCC sanitizers are a combination of compiler instrumentation and runtime support. It's not at all like `lint` -- the detected errors are actual runtime errors vs. hypothetical static analyzer hints.

Fixing the so-called "false positive" is necessary so that later memory issues can be uncovered. Not complicated.

Clang does support a special case list. In this instance, the juice isn't worth the squeeze -- zapping `realloc`-ed memory is really low overhead compared to maintaining a special case file.

pkoning2:
My point is that this**is** a false positive, caused by inadequate analysis done by the tool.  I wonder if Coverity gets this right, though it's not foolproof either.
Checkers are all well and good, but when they get the wrong answer they should be fixed; it doesn't make sense for code to do extra stuff to work around checker complaints that do not correspond to real defects.

bscottm:
@pkoning2 The sanitizer detected reads to uninitialized memory at runtime. So, let's make sure it's initialized so that we can find other uninitialized memory issues. Why is doing the right thing controversial?

* Zeroing out the log de-duplicator's `realloc`-ed memory is slightly lower overhead than zeroing out the incremental space added. Since it's going to be completely overwritten, there's really no point to computing the increment that needs to be zeroed. No penalty for doing the right thing.
* `calloc` the scratch space used by PCRE. No penalty for doing the right thing.

I would have given Mark's `prce` argument a bit more credibility had he merely pointed out that `-1 - (-1) = 0` (both offsets are `-1` for an uncaptured pattern), so that at worst case, simh is allocating a 1-byte string. Instead, he launched into a completely unrelated polemic.

I submit that the code is easier to read vs. being clever. Someday, someone else will have to maintain this codebase. Might as well try to leave them something that's reasonably easy to maintain.

markpizz:
> @pkoning2 The sanitizer detected reads to uninitialized memory at runtime. So, let's make sure it's initialized so that we can find other uninitialized memory issues. Why is doing the right thing controversial?

The code DOES NOT read from uninitialized memory and it never has.  The tool detected a false positive!  If you can identify that fact in some way that that the tools knows about then you can probably use it.

> I would have given Mark's prce argument a bit more credibility had he merely pointed out that -1 - (-1) = 0 (both offsets are -1 for an uncaptured pattern), so that at worst case, simh is allocating a 1-byte string. Instead, he launched into a completely unrelated polemic.

It's your job, when reporting problems indicated by an omnipotent tool to confirm the details!

> I submit that the code is easier to read vs. being clever. Someday, someone else will have to maintain this codebase. Might as well try to leave them something that's reasonably easy to maintain.

This is a really great argument to justify the tools report of a false positive and its apparent inability for the user to indicate the fact that it is a false positive.

> @pkoning2 said: My point is that this is a false positive, caused by inadequate analysis done by the tool. I wonder if Coverity gets this right, though it's not foolproof either.

Coverity absolutely has a way for the reviewer to declare particular flagged defects as false positives.

> Checkers are all well and good, but when they get the wrong answer they should be fixed; it doesn't make sense for code to do extra stuff to work around checker complaints that do not correspond to real defects.

Agreed 100%.

Issue #12, opened 2022-06-15
Segmentation fault after changing floppy images with microvax1 and microvax2 simulators


vaxorcist:
Every time I try to change floppy images (for creating a set of VMS Standalone Backup floppies), I get a "Segmentation fault", both with the microvax1 and the microvax2 simulators. I had the same problem with changing disk images before. This is a severe bug as it prevents changing media with the simulator running.

- #### Context
MicroVMS V4.6
$ @SYS$UPDATE:STABACKIT

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
microvax2:
sim> sh ver
MicroVAX II (KA630) simulator V4.0-0 Current
    Simulator Framework Capabilities:
        64b data
        64b addresses
        Threaded Ethernet Packet transports:PCAP:TAP:NAT:UDP
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        RAW disk and CD/DVD ROM support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC 8.3.0
        Simulator Compiled as C arch: x64 (Release Build) on Jun 14 2022 at 22:55:04
        Build Tool: simh-makefile
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: SDL Version 2.0.9, PNG Version 1.6.36, zlib: 1.2.11
        PCRE RegEx (Version 8.39 2016-06-14) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        Ethernet packet info: libpcap version 1.9.0-PRE-GIT (with TPACKET_V3)
        OS: Linux think 4.19.0-12-amd64 #1 SMP Debian 4.19.152-1 (2020-10-18) x86_64 GNU/Linux
        Processor Name: Intel(R) Core(TM) i5-2520M CPU @ 2.50GHz
        tar tool: tar (GNU tar) 1.30
        curl tool: curl 7.64.0 (x86_64-pc-linux-gnu) libcurl/7.64.0 OpenSSL/1.1.1n zlib/1.2.11 libidn2/2.0.5 libpsl/0.20.2 (+libidn2/2.0.5) libssh2/1.8.0 nghttp2/1.36.0 librtmp/2.3
        simh git commit id: 5a7a4c77
[...]

PR #275, opened 2023-07-15
sim_disk.c: only free filebuf if it was allocated here too.

Rhialto:
PDP11/pdp11_rq.c re-uses ->filebuf as a sector buffer (under the #defined name rqxb) and allocates it as such. If an RQ disk is detached and another attached, this buffer would be lost and the pointer reset to NULL. sim_disk.c would only allocate the buffer if UNIT_BUFABLE is set, which means to buffer the whole disk. Since this rightly isn't set on RQ disks, the pointer would remain NULL and segfaults would ensue.  See open-simh issue 274.

### DIFF
--- a/PDP11/pdp11_rq.c
+++ b/PDP11/pdp11_rq.c
@@ -172,6 +172,7 @@ extern int32 MMR2;
 #define unit_plug       u4                              /* drive unit plug value */
 #define io_status       u5                              /* io status from callback */
 #define io_complete     u6                              /* io completion flag */
+/* we can re-use filebuf because we don't set UNIT_BUFABLE in flags */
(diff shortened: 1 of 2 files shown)


markpizz:
The commits in this PR should be squashed.

Conventions referring to changes to sim_disk behavior have always had commit messages starting with DISK:.  Similarly, changes to sim_ether have started with ETHER: and changes to sim_tape have started with TAPE:, etc.

Rhialto:
@markpizz Given that you took issue #12 as a prompt to fix it in your own non-free-software version only, and didn't contribute it here, I will value advice from other people higher.

pkoning2:
I merged this.

PR #276, opened 2023-07-17
AltairZ80: Updated documentation

psco:


pkoning2:
I hadn't noticed this before, but Altair is the only emulator that has documentation that no one else can modify.  All the others have .doc files.  Please provide a .doc file, or whatever other format is the source of that PDF file.  It can be instead of the PDF or in addition to, but one or the other is needed.

psco:
I will create another pull request to address this.

pkoning2:
Thanks Peter!

Issue #277, opened 2023-07-17
Adding RQDX1


Rhialto:
This is a continuation of #215, which was "closed as fixed" by @vaxorcist, but not actually fixed in this repo, and I couldn't reopen the issue myself.

To really fix the issue in open-simh, I added a RQDX1 flavor to our RQ device family. As discussed, I also made that version the default in `microvax1`.

The model ID for RQDX1 I obtained from the list in appendix C of http://www.bitsavers.org/pdf/dec/dsa/mscp/Mass_Storage_Control_Protocol_Ver_2.4.0_Jun92.txt . While having that list, I checked it against our implementation and I noticed some discrepancies. I added some comments in the code, with the goal to discuss them, and resolve that before committing the final version. Here is a copy:
```
// AFAIK the UNIBUS KLESI and QBUS KLESI used the same controller type ...
#define KLESI_CTYPE     1               // RC25 controller (UNIBUS and QBUS both)
#define KLESI_UQPM      1               // Mass_Storage_Control_Protocol_Ver_2.4.0_Jun92.txt Appendix C:
#define KLESI_MODEL     1               // 1 | HSC50; 3 | RC25 integrated controller
```
I suppose we should change to 3 here since the model value we have now is for HSC50.
```
#define UDA50_CTYPE     3               // UNIBUS SDI (RAxx) controller
#define UDA50_UQPM      6               // really type of UDA50A; UDA50 is 2
#define UDA50_MODEL     6
```
The model value we are using is really for UDA50A, not UDA50. I changed the name in one location in the text as a place holder. Maybe we should change it everywhere. Or maybe the difference is too small to bother.
```
#define KDA50_CTYPE     6               // QBUS SDI (RAxx) controller
#define KDA50_UQPM      13              // KDA50-Q
#define KDA50_MODEL     13
```
Here the given name is just slightly different. Probably doesn't matter.
```
#define KRU50_CTYPE     8               // UNIBUS RRD40/50 CDROM controller
#define KRU50_UQPM      26              // unassigned in appendix C
#define KRU50_MODEL     26
```
This one doesn't occur in the list at all.

As we have seen in practice, these model numbers don't matter in the vast majority of cases, so we could decide to do nothing about the discrepant cases. Or add a few more models so that we have both versions. Or we just add a pseudo register to access the model number, so that you can set any arbitrary value.

[...]

Issue #271, opened 2023-07-10
The MicroVAX I simulator microvax1 crashes when using MicroVMS Standalone Backup V4.0.


vaxorcist:
The MicroVAX I simulator microvax1 crashes when using MicroVMS Standalone Backup V4.0. (using e.g. MicroVMS Standalone Backup V4.7 it works):

```
./microvax1.ini-14> ATT RQ1 STABA040-1.dsk
%SIM-INFO: RQ1: 'STABA040-1.dsk' Contains ODS2 File system
%SIM-INFO: RQ1: Volume Name: SYSTEM_1     Format: DECFILE11B   Sectors In Volume: 800
./microvax1.ini-25> BOOT RQ1
%SIM-INFO: Loading boot code from internal ka610.bin
.2..1..0.
  

  MicroVMS Version V4.0 15-SEP-1984 22:29
  


PLEASE ENTER DATE AND TIME (DD-MMM-YYYY  HH:MM)  10-JUL-2023 12:16

Please remove the volume "SYSTEM_1" from the console device.

Insert the standalone application volume and enter "YES" when ready: 
Simulation stopped, PC: 80008B1F (BRB 80008B1F)
sim> ATT RQ1 STABA040-2.dsk
%SIM-INFO: RQ1: 'STABA040-2.dsk' Contains ODS2 File system
%SIM-INFO: RQ1: Volume Name: BACKUP       Format: DECFILE11B   Sectors In Volume: 800
sim> c


Insert the standalone application volume and enter "YES" when ready: yes

Resuming load operation on volume "BACKUP", please stand by . . .


%BACKUP-I-IDENT, Stand-alone BACKUP V4.0; the date is 10-JUL-2023 12:16:18.55
Simulation stopped, PC: 80008B1F (BRB 80008B1F)
sim> ATT RQ1 BL-B001B-BE.dsk
%SIM-INFO: RQ1: 'BL-B001B-BE.dsk' Contains ODS2 File system
%SIM-INFO: RQ1: Volume Name: MICROVMS     Format: DECFILE11B   Sectors In Volume: 800
sim> c

$ backup/log/init dua1:MICROVMS/SAVE dua0:

Exception in interrupt or exception, PC: 8009DA60 (MOVQ R0,-(SP))
sim> sh ver
MicroVAX I (KA610) simulator V4.0-0 Current
    Simulator Framework Capabilities:
        64b data
        64b addresses
[...]

Issue #215, opened 2023-04-11
MicroVAX I simulator microvax1 not booting VMS Standalone Backup


vaxorcist:
The MicroVAX I simulator microvax1 does not boot VMS Standalone Backup.

This was tested with VMS Standalone Backup versions V4.7 and V5.5-2 on RX50 images.
See below for V4.7 (same result for V5.5-2):
```
Logging to file "STABA047-MV1.log" at Tue Apr 11 08:36:51 2023
MicroVAX I (KA610) simulator Open SIMH V4.1-0 Current
    Simulator Framework Capabilities:
        64b data
        64b addresses
        Threaded Ethernet Packet transports:PCAP:TAP:NAT:UDP
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        RAW disk and CD/DVD ROM support
        Asynchronous I/O support (Lock free asynchronous event queue)
        Asynchronous Clock support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC 10.2.1 20210110
        Simulator Compiled as C arch: x64 (Release Build) on Mar 20 2023 at 08:30:39
        Build Tool: simh-makefile
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: SDL Version 2.0.14, PNG Version 1.6.37, zlib: 1.2.11
        PCRE RegEx (Version 8.39 2016-06-14) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 1ms
        Ethernet packet info: libpcap version 1.9.0-PRE-GIT (with TPACKET_V3)
        OS: Linux think 4.19.0-12-amd64 #1 SMP Debian 4.19.152-1 (2020-10-18) x86_64 GNU/Linux
        Processor Name: Intel(R) Core(TM) i5-2520M CPU @ 2.50GHz
        tar tool: tar (GNU tar) 1.34
        curl tool: curl 7.74.0 (x86_64-pc-linux-gnu) libcurl/7.74.0 OpenSSL/1.1.1n zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3
        git commit id: 86b19c76+uncommitted-changes
        git commit time: 2023-02-26T19:53:06-0500
[...]

PR #278, opened 2023-07-17
DISK: add RQDX1 model and make it default for MV1.

Rhialto:
Now MV1 can boot Standalone Backup 4.0 from RX50.
See issue #215, #271, #277.

### DIFF
--- a/PDP11/pdp11_rq.c
+++ b/PDP11/pdp11_rq.c
@@ -122,6 +122,12 @@ extern int32 MMR2;
 #error "Assertion failure: RQ_NUMCT exceeds 4"
 #endif
 
+#if defined (VAX_610)
+#define MICROVAX1       1
+#else
+#define MICROVAX1       0
+#endif
+
 #include "pdp11_uqssp.h"
 #include "pdp11_mscp.h"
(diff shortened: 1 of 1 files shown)


PDP11/pdp11_rq.c
```diff
@@ -753,31 +753,35 @@ x  RA73 70(+1)  21      2667+   21      1       ?       3920490
 
 // AFAIK the UNIBUS KLESI and QBUS KLESI used the same controller type ...
 #define KLESI_CTYPE     1               // RC25 controller (UNIBUS and QBUS both)
-#define KLESI_UQPM      1
-#define KLESI_MODEL     1
+#define KLESI_UQPM      1               // Mass_Storage_Control_Protocol_Ver_2.4.0_Jun92.txt Appendix C:
+#define KLESI_MODEL     1               // 1 | HSC50; 3 | RC25 integrated controller
```
pkoning2, line 757: That is (and was) incorrect: model 1 is HSC50, for RC25 you need 3.  If I change it to 3 RSTS reports it as RC25 (rather than "unknown" which is how it reports things that aren't plausible PDP-11 peripherals).

PR #280, opened 2023-07-20
FOR REFERENCE ONLY: Add pseudo registers MODEL and PORT to RQ and TQ controllers.

Rhialto:
EXAMINE RQ MODEL
DEPOSIT RQ MODEL 42
EXAMINE RQ STATE

### DIFF
--- a/PDP11/pdp11_rq.c
+++ b/PDP11/pdp11_rq.c
@@ -757,7 +757,7 @@ x  RA73 70(+1)  21      2667+   21      1       ?       3920490
 #define KLESI_MODEL     3
 
 #define RUX50_CTYPE     2               // UNIBUS RX50-only controller
-#define RUX50_UQPM      10		// this should be 10 according to the MSCP spec
+#define RUX50_UQPM      10
 #define RUX50_MODEL     10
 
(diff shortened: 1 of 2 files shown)


pmetzger:
@Rhialto Do you want something like this committed? I don't understand the FOR REFERENCE ONLY thing?

Rhialto:
Sounds good to me. If I recall correctly, the "for reference only" is because some people didn't seem to think this was a desirable feature. But I had already done the work and I wanted to "preserve the code for history".

#392 is kind of in the same category. You could consider using that one, or something like it, too.

pmetzger:
@Rhialto You can do pull requests to ZIMH for this stuff after I've got it building on NetBSD (hopefully later today.)

markpizz:
Having simh REGister variables to reflect the details of the device settings harms nothing and really doesn't add much for these devices since the primary user interface is the SET RQ|TQ {RQDX1|RQDX2|...} and SHOW RQ|TQ commands which display the desired details.  These REGisters are only visible to SCP EXAMINE & DEPOSIT commands and are not available (as such) to anything running in a simulator.

PR #283, opened 2023-07-30
AltairZ80: Add CTS command to M2SIO device

deltecent:
Adds CTS command to force CTS to an active low state which will allow the transmission of data without CTS being present. Corrects the documentation to state that CTS controls the 6850 UART's transmitter, not RTS. From the MITS 88-2SIO documentation:

> If the Data Carrier Detect and Clear to Send inputs are not connected, they must be jumpered to ground.


psco:
Looks fine and can be merged.

deltecent:
Working with Walt to bring CP/M terminal programs back to life brought this to light.

PR #284, opened 2023-07-31
Remove windows_build.yml

bscottm:
Remove windows_build.yml due to a MS compiler enforcement change. Build fails because libpng16.lib was compiled with a different compiler version (VS 2008?) relative to the current VS 2022 installed in the Github windows-latest runner.

MS decided that libraries must be compiled with a compatible compiler, preferably the compiler-in-use. While not conclusively traced to a MS trouble ticket, it seems reasonable to assume there exists a trouble ticket wherein an older MS compiler generated code that was incompatible with VS 2022, thereby necessitating the enforcement (speculation: LTO instrumentation, for example?)


markpizz:
> MS decided that libraries must be compiled with a compatible compiler, preferably the compiler-in-use. While not conclusively traced to a MS trouble ticket, it seems reasonable to assume there exists a trouble ticket wherein an older MS compiler generated code that was incompatible with VS 2022, thereby necessitating the enforcement (speculation: LTO instrumentation, for example?)

If you're mentioning a MS trouble ticket, then provide pointers otherwise it is a rumor created by you.

Meanwhile, the fact is that some static libraries compiled with the current (VS 2022 in this most recent example) compiled in release mode (the debug mode compiles do not have an issue) are not compatible with release compiled code that references these libraries when compiled with an update release of that compiler (again VS 2022).  Other static compiled libraries work without issue even if they're compiled with various often much older VS compiler versions going back all the way to VS 2008.

The offending static library is libpng16 which the release build somehow dislikes.  

------
As for the change in this PR, your deleting the one file (windows_build.yml) that @tlhackque seems to have put some thought into since it seemed to potentially target both the master branch and the Supnik-Current branches.  I have no problem with deleting this, BUT if this goes, then these files:
1) "Visual Studio Projects\*.vcxproj"
2) "Visual Studio Projects\Simh.ci.sln"

all of which were merely generated by VS 2022 when @tlhackque opened "Visual Studio Projects\Simh.sln" with VS 2022.
There is 0 value to these extra generated pieces and there is negative value since having them there means that for them to be theoretically useful, folks adding modules to any simulator will have to manually update these files, this along with the build library compatibility problem motivating the removal of windows_build.yml in the first place.

markpizz:
> Build fails because libpng16.lib was compiled with a different compiler version (VS 2008?) relative to the current VS 2022 installed in the Github windows-latest runner.

I missed this comment earlier, but FACTS WRONG AGAIN!  The involved libpng16.lib was compiled with a slightly earlier VS 2022 compiler NOT VS 2008.  EACH TIME MS releases an update to the previously released "current" VS Version, this library (compiled with the same VS number, but an earlier update) no longer is deemed compatible.

bscottm:
@markpizz: And the world is going to come to an abrupt end! OMFG!

bscottm:
@markpizz: I admitted speculation on my part in the explanation. Seriously, stop halting and catching fire. It's not the end of the world, and it is possible to politely ask me to change/update the explanation without hysterics.

If you actually look up [LINK error C1047](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1047?view=msvc-170), it's specifically related to my parenthesized speculation -- whole program optimization (/GL) and link-time code generation (/LTCG).

markpizz:
>Seriously, stop halting and catching fire.

I tend to catch fire when I have to repeat the absolutely same thing well more than twice.

Feel free to leave any amount of turds in this repo.  I can only give advice.

pkoning2:
It looks pretty simple to me.
We have two CI streams for MS, one that deals with the MS bizarre lack of compatibility, and one that doesn't.  Since it isn't necessary to have two, it suffices to remove the one that doesn't handle MS, rather than attempt to repair it.

markpizz:
> @pkoning2 said: It looks pretty simple to me.
We have two CI streams for MS, one that deals with the MS bizarre lack of compatibility, and one that doesn't. Since it isn't necessary to have two, it suffices to remove the one that doesn't handle MS, rather than attempt to repair it.

I agree completely, so unless you want irrelevant, useless, generated files in the repo, the other files should go as well, or like I said you can leave any amount of these irrelevant, useless, generated files laying around.  If these should stay, then you should start keeping all sorts of .o, .obj and generated binaries in the repo along with other irrelevant, useless, generated files.

PR #286, opened 2023-08-18
CMake: Win XP (v141_xp) VS install script

bscottm:
Move the PowerShell code from .github/workflow/cmake-builds.yml into its own script, cmake/github_v141_xp.ps1, to keep cmake-buils.yml readable. The script also facilitates synchronizing with the Chocolatey installer's Wait-VSIstallerProcesses function easier (from which this this code is derived.)

cmake-builds.yml: Check the output from Get-VSSetupInstance and Set-VSSetupInstance. Empty (or null) output indicates that the v141_xp install did not complete successfully. Build process will bail when that happens.

cmake/v141_xp_install.ps1: Unused and now unnecessary script.


XVilka (approved):

pkoning2:
Merged.

PR #291, opened 2023-09-08
SEL32: Do general code cleanup to remove unused code.

AZBevier:
SEL32: Update Ping and ICMP support code to use correct packet size.
SEL32: Update SetupNet script to support latest Fedora release.
SEL32: Improve disk write speed.
SEL32: Add .tap file reassignent support in sel32_mt.c.


pkoning2:
merged.

Issue #294, opened 2023-09-15
apple silicon build problem(s?)


at7heb:
- #### Context

I'm trying to build simh on a MacBook Pro with the M1 chip, Ventura 13.5.2. I use home-brew. There are build problems.

- #### how you built the simulator or that you're using prebuilt binaries
First problem is accessing the GNU getopt. I get the following output (I've not included the output from the brew installations and updates; gnu-getopt is installed)

> heb@howards-mbp simh % cmake/cmake-builder.sh   
> cmake/cmake-builder.sh: GNU getopt needed for this script to function properly.
> cmake/cmake-builder.sh: Specifically, a 'getopt' that supports the '-T' flag (enhanced getopt)
> 

That is an easy fix in `cmake/cmake-builder.sh` -- add this line

> [[ -d /opt/homebrew/opt/gnu-getopt/bin ]] && PATH="/opt/homebrew/opt/gnu-getopt/bin:$PATH"

The next problem is 

> heb@howards-mbp simh % cmake/cmake-builder.sh
> ** cmake version 3.27.5
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake).
> realpath: /Users/heb/Documents/dev/simh/cmake/build-unix: No such file or directory
> cmake/cmake-builder.sh: SIMH top-evel directory: /Users/heb/Documents/dev/simh
> cmake/cmake-builder.sh: Build directory:         
> usage: mkdir [-pv] [-m mode] directory_name ...
> /opt/homebrew/bin/cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -S /Users/heb/Documents/dev/simh -B  
> CMake Error: No build directory specified for -B
> CMake Error: Run 'cmake --help' for all supported options.
> *** cmake/cmake-builder.sh: Errors detected during environment generation. Exiting.
> heb@howards-mbp simh % 

after running make with --help, I tried building with the --flavor xcode and ran into the next problem

> heb@howards-mbp simh % cmake/cmake-builder.sh --flavor xcode    
> ** cmake version 3.27.5
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake).
> realpath: /Users/heb/Documents/dev/simh/cmake/build-xcode: No such file or directory
> cmake/cmake-builder.sh: SIMH top-evel directory: /Users/heb/Documents/dev/simh
> cmake/cmake-builder.sh: Build directory:         
> usage: mkdir [-pv] [-m mode] directory_name ...
[...]

PR #295, opened 2023-09-16
Fix apple silicon build problems Issue #294

at7heb:
- find gnu-getopt when installed by homebrew in /opt
- invoke realpath only on known path components (needed for some realpath implementation(s)) (relevant when build-flavor directory does not yet exist)


bscottm:
@at7heb: thanks for that!

pkoning2:
Merged.

PR #297, opened 2023-09-23
PDP11: RF11: Fix DAE computation in service routine

al20878:
This is a re-do of RP #241 as confirmed by Bob Supnik (and without the "window dressing")
Please apply

### DIFF
--- a/PDP11/pdp11_rf.c
+++ b/PDP11/pdp11_rf.c
@@ -382,7 +382,7 @@ do {
     } while ((rf_wc != 0) && (rf_burst != 0));          /* brk if wc, no brst */
 
 rf_da = da & DMASK;                                     /* split da */
-rf_dae = (rf_dae & ~RFDAE_DAE) | ((rf_da >> 16) & RFDAE_DAE);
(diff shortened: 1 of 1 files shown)


al20878:
Can this please be pushed?  Bob Supnik told me he had applied this patch to his V3 code (back on Jun 26, in a private email)

al20878:
May I ask what is preventing this patch to get merged? (and #302 and #304, for that matter?)

pkoning2:
Merged.

al20878:
@pkoning2 :

Thanks, but something is not going right with all these merges...  This is what I see:

ClosedWithUnmergedCommits

"Closed with unmerged commits" and also the icon is a broken merge.

There's still #304 pending with an obvious fix.  Could you please push it as well?

pkoning2:
You're looking at a github bug.  Rather than merge the changes one at a time (which triggers a build each time) I rebase and merge them in a local workspace and then push them all.  Github can't understand that process, and this is what you see as a result.  Look in the resulting code, you should find the correct content.

PR #298, opened 2023-09-23
TX-0: Execute pad and cry separately

dpeschel:
Before, when the simulator decoded pad, it handled pad without cry and pad with cry. When it decoded cry, it printed a "TO DO" message.

http://bitsavers.org/pdf/mit/tx-0/memos/M-5001-19_macros_Apr60.pdf derives a simple algorithm for cry. Implement that, remove the test for cry from the pad code, and reformat the debug messages.

The same change appears in sim_instr and sim_opr_orig.


hharte (approved):
Thanks @dpeschel !

hharte:
@pkoning2 This looks good and can be merged.

pkoning2:
Merged.

PR #299, opened 2023-09-23
TX-0: Stop ptp_svc from setting the ios bit; sim_instr never clears it and simulation loops.

dpeschel:
The CPU-to-device routines and device service routines need a redesign. In the meantime, this quick patch makes tape punching possible.


hharte (approved):
Thanks @dpeschel .  @pkoning2 this looks good and can be merged.

pkoning2:
Merged.

PR #301, opened 2023-09-26
Fixed to PDP10 KL and Sim-Card.

rcornwell:
This is a few patches to fix some issues. These are independent of each other so they should not be merged into one commit.


pkoning2:
merged.

PR #302, opened 2023-10-02
SCP: Fix correctly loading simh.ini when user HOME not defined

al20878:
When loading `simh.ini` if none of the HOME directory environment is set, a stale `stat` value is used in decision whether to load `simh.ini` from the current directory.

This patch fixes this issue (so the local `simh.ini` will be tried).


al20878:
I see this has been applied to the proprietary `simh`...  Can this please be also done with `open-simh`?

pkoning2:
Merged.

PR #303, opened 2023-10-05
PDP11: RP11: Major update after XXDP

al20878:
I have run the code thru XXDP and some other OS's and scenarios rigorously, and found discrepancies, which need to be addressed by this rather extensive patch.

1. Each drive must implement its own "drive status" register, to be able to track per-drive errors / conditions correctly;
2. Fixed `INT_SET()` /  `INT_CLR()` in `RPCS` write function (wrong order of the "`if`" conditions);
3. Some behavior was implemented not exactly how it was expected from the real hardware, such as:
  a. Post-I/O register values in `RPDA` and `RPCA` (including the corner case of pack overflow);
  b. I/O stacking, which wasn't mentioned in any available documentation, but only XXDP listings;
  c. `RESET`/`IDLE` function must be accepted for a "busy" controller;
  d. `HOME` function must always execute, even when "device ready" is not set (e.g. when `SEEK` error detected);
  e. `SEEK` incomplete should not respond with "device ready" (however, the condition can be cleared by `HOME`, d.);
  f. `WLOA`-induced write-lock violation wasn't reflected in "device status".
4. Some timing was off so that the device worked "too fast" -- this was fixed (except when the races are in the actual test code, and cannot be logically fixed);
5. `WLOA` setup command bug was fixed;
6. Added more code comments found per the above peculiarities.

There is a DEC/X11 `IOMODX` test available (`XRPAM0`) but I was still unable to build it successfully [somehow, its linker zeroes the entire XXDP+ drive when creating a file on a DU: (RQ) device -- so the XXDP+ disk image can simply be just corrupt].  But I still have a couple of ideas about it, which I want to explore.  If the test builds and shows some additional discrepancies (which I do not anticipate from just looking through the listing), I'll post another patch.

### DIFF
--- a/PDP11/pdp11_rr.c
+++ b/PDP11/pdp11_rr.c
@@ -68,9 +68,10 @@
 
 /* Parameters in the unit descriptor */
 #define CYL             u3                              /* current cylinder */
-#define FUNC            u4                              /* function */
-#define HEAD            u5                              /* current track */
-#define SEEKING         u6                              /* unit performing a seek */
+#define HEAD            u4                              /* current track */
+#define FUNC            u5                              /* current func */
+#define STATUS          u6                              /* RPDS bits for the unit */
+#define SEEKING         us9                             /* seek cmd underway */
 
 /* 4 dummy UNIBUS registers followed by... */
 #define RP_IOFF         4
@@ -114,7 +115,9 @@ static BITFIELD rp_ds_bits[] = {
     BIT(RDY),
     ENDBITS
 };
-#define RPDS_DKER(x)    ((x) & (RPDS_HNF | RPDS_INC) ? RPER_DRE : 0)
+#define RPDS_REAL       0017400                         /* bits stored */
+#define RPDS_DKER       (RPDS_HNF | RPDS_INC)           /* drive error */
+#define RPER_DKER(x)    ((x) & RPDS_DKER ? RPER_DRE : 0)/* DRE for RPER */
 
 /* RPER 776712, error register, read-only */
 static BITFIELD rp_er_bits[] = {
@@ -269,11 +272,11 @@ static BITFIELD rp_suca_bits[] = {
 /* Maintenance Write Lockout Address (LOA) (the switches on the maint. panel) */
 static const char* offon[] = { "OFF", "ON" };
 static BITFIELD rp_wloa_bits[] = {
-#define RPWLOA_IMPL     01777
-#define RPWLOA_CYL2     0377                            /* cyls locked (x2) */
+#define RPWLOA_IMPL     03777
+#define RPWLOA_CYL2     0377                            /* cyls locked (x2+1) */
     BITFFMT(CYL2,8,%u),
 #define RPWLOA_V_DRV    8
-#define RPWLOA_M_DRV    3
+#define RPWLOA_M_DRV    7
 #define RPWLOA_DRV      (RPWLOA_M_DRV << RPWLOA_V_DRV)  /* drive(s) locked */
     BITFFMT(DRV,3,%u),
 #define GET_WLOACYL(x)  ((((x) & RPWLOA_CYL2) << 1) | 1)
@@ -343,7 +346,7 @@ static t_stat rr_wr (int32 data, int32 PA, int32 access);
 static int32  rr_inta (void);
 static t_stat rr_svc (UNIT *uptr);
 static t_stat rr_reset (DEVICE *dptr);
-static void   rr_go (void);
+static void   rr_go (int32 func);
 static void   rr_set_done (int32 error);
 static void   rr_clr_done (void);
 static t_stat rr_boot (int32 unitno, DEVICE *dptr);
@@ -358,6 +361,7 @@ static const char *rr_description (DEVICE *dptr);
 
 /* RP11 data structures
 
+   rr_dib       RR device context for PDP-11
    rr_reg       RR register list
    rr_unit      RR unit list
    rr_mod       RR modifier list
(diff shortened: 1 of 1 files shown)


markpizz:
You might want to fix an unusual spelling of "suppoerted".

In general, the PR creation comments should be in the commit message so that the thinking behind the changes travels with the repo rather than depending on a particular repo instance/location.  Commit messages can be of arbitrary length.  Start extra info after a blank line following the summary message and wrap long lines less than 80 characters.

al20878:
> You might want to fix an unusual spelling of "suppoerted".

Thanks.  Fixed.  Although it's not the first typo in the simh code and def is not going to be the last one.  Besides, it was inside a comment so not even exposed.

> PR creation comments should be in the commit message

Redid the commit message to re-state all changes detailed in the PR description.

Both changes squashed into a single commit and re-pushed.

al20878:
UPD:  DEC/X11 `XRPAM0` test has also passed for this implementation:
```
BSY>
RPAM0 END PASS #00135. RUNTIME: 000:00:00 PSTIME: 000:00:00

RPAM0 END PASS #00136. RUNTIME: 000:00:00 PSTIME: 000:00:00

RPAM0 END PASS #00137. RUNTIME: 000:00:00 PSTIME: 000:00:00

RPAM0 END PASS #00138. RUNTIME: 000:00:00 PSTIME: 000:00:00

RPAM0 END PASS #00139. RUNTIME: 000:00:00 PSTIME: 000:00:00

RPAM0 END PASS #00140. RUNTIME: 000:00:00 PSTIME: 000:00:00

...

SUMMARY AT RUNTIME: 000:00:00
RPAM0 AT VA: 066370 STAT 154015 PASS #00177 HRDERRS 00000 SFTERRS 00000
```
No errors detected.

al20878:
I rebased this commit on top of the latest commits in open-simh master...  Can this please be applied?

PDP11/pdp11_rr.c
```diff
@@ -488,7 +492,7 @@ DEVICE rr_dev = {
 
 /* I/O dispatch routine, I/O addresses 17776710 - 17776736
 
-   17776710     RPDS    read-only except for attention bits
+   17776710     RPDS    read-only except for attention bits (write-clear)
```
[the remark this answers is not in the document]
al20878: Was that the only thing that prevented the patch to be applied?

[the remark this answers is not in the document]
pkoning2: No, it's something I noticed while reading the patch (which I haven't finished yet).  And the issue isn't the clarification but rather that the RPDS write handler seems to clear attention for any write, rather than clearing only attention bits corresponding to 1 bits in the data.

And yes, BIC RPDS,RPDS is obviously wrong; MOV RPDS,RPDS is the common and correct idiom for "clear all the write-1-to-clear bits"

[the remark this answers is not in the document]
al20878: >  And the issue isn't the clarification but rather that the RPDS write handler seems to clear attention for any write,

I don't understand where you see that.

The relevant code is this (line 614):
```
           rpds &= ~(data & RPDS_ATTN);                /* clr attention bits */
```
which does the bit clear.

I am submitting this patch after XXDP'ing and DEC/X11'ing the device implementation, so if this basic thing wasn't handled correctly, the tests would not have passed.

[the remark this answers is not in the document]
pkoning2: You're right, I misread it.  Thanks.

PDP11/pdp11_rr.c
```diff
@@ -1083,7 +1178,7 @@ static t_stat rr_reset (DEVICE *dptr)
 {
     int32 i;
 
-    /* some sanity check first */
+    /* compile-time sanity check first */
```
pkoning2, line 1181: That doesn't seem to be a compile time test.  Do we have STATIC_ASSERT or something like that?  It would be nice to use that when possible.

PDP11/pdp11_rr.c
```diff
@@ -1200,21 +1315,21 @@ static const uint16 rr_boot_rom[] = {
...
 /* 002002 */ 0012706, BOOT_ENTRY,     /* BOOT:  MOV     #BOOT, SP       ; ENTRY POINT PC               */
 /* 002006 */ 0112700, 0000000,        /*        MOVB    #0, R0          ; UNIT NUMBER                  */
 /* 002012 */ 0012701, 0176726,        /*        MOV     #176726, R1     ; RPCS + 12                    */
-/* 002016 */ 0012704, BOOT_START+020, /*        MOV     #<START+20>, R4 ; BACKLINK TO ROM W/OFFSET 20  */
+/* 002016 */ 0012704, BOOT_START+020, /*        MOV     #<START+20>, R4 ; BACKLINK TO PROM W/OFFSET 20 */
 /* 002022 */ 0005041,                 /*        CLR     -(R1)           ; DISK ADDRESS                 */
 /* 002024 */ 0005041,                 /*        CLR     -(R1)           ; CYLINDER ADDRESS             */
 /* 002026 */ 0005041,                 /*        CLR     -(R1)           ; MEMORY ADDRESS               */
-/* 002030 */ 0012741, 0177000,        /*        MOV     #-512., -(R1)   ; WORD COUNT                   */
+/* 002030 */ 0012741, 0177000,        /*        MOV     #-512., -(R1)   ; WORD CT (ALWAYS 2 FULL SECTS)*/
```
pkoning2, line 1332: Two sectors?  That's odd, I thought that bootstraps always read 512 bytes (one sector, in most disks).

al20878: Paul, please read listings for M9312 -- for disk devices the word count is always 2 full sectors.

al20878: May I ask why are you scrutinizing my code comments?

PDP11/pdp11_rr.c
al20878: > That doesn't seem to be a compile time test.

It is.  The condition is a constant expression known at the time of compilation.  So `assert()` would have been evaluated by the compiler, and there would (most likely) be an unconditional call for the assert handler if the condition was false.  That's what I mean.  At any rate, `assert()` is not compiled at all when the simulators are built without `DEBUG=1` as make's argument.

pkoning2: Sure, so it's a run time call to assert with a constant (evaluated at compile time) argument.  
There is such a thing as "compile time assert" but it's a different thing entirely, and very useful when available: it's a macro, or compiler built-in operation, that is completely handled at compile time.  If the check succeeds, no code is generated at all, ever.  If the check fails, you get a compile error.  So the failure shows up at build time, not run time as an assert does.
There are various ways of doing this; one relies on the fact that arrays with negative bounds are invalid.  So something like
```
#define static_assert(cond) { char assert ## __LINE__[-(!(cond))]; }
```
give or take a few details.

PDP11/pdp11_rr.c
pkoning2: In this particular case, while the question was prompted by what I read in the comments it is actually about the substance of the code.
I'll definitely have to take a look at the M9312, I wonder if that's a mistake or an intentional change introduced in that device.  In the older M792 boot card (the one with discrete diodes), the word count is definitely -256, not -512.  In the MR-11B boot it's -512, though.  I wonder if that's because it added support for the TM11, which has a byte count rather than a word count.  So -512 there means 512 bytes, the standard boot block size.

al20878: > I'll definitely have to take a look at the M9312

To make your investigation easier, you can take a look at M9312 listings here (among other places, obviously):

http://www.ak6dn.com/PDP-11/M9312/

RP11 bootstrap, and RK11 (and most everything else what I checked in there for disks) are doing `-512.` for the WC registers.  Whether was it a bug or not, is no longer relevant now, I suppose, because it has been like that for all this time.  IIRC, RSTS keeps time in the _last_ 3 tops words of the "first sector" (loc octal 772, 774, and 776), when it is doing a software bootstrap (or reboot).  In case of hardware bootstrap (that we're talking about here), those locations are obviously overwritten by the boot sector (regardless of whether only one sector is getting loaded, or two).

pkoning2: Yes, good point.  Thanks for the link.

PR #304, opened 2023-10-06
PDP11: TS11: Fix help output for BOOT command

al20878:
The BOOT command for this device was not correctly documented in HELP (appearing both supported and not) for PDP-11.
This patch fixes the issue and syncs HELP output with the actual code (by using the same `#if` conditional).


al20878:
@pkoning2 can you please explain what's holding this one, and why it was skipped last night when you applied a bunch of other PRs?

pkoning2:
Yes.... I got to a pile of the others but not yet this one.

PR #312, opened 2023-10-13
SCP: SOCK: Increase backlog in the listen() syscall

al20878:
The currently used backlog of (1) effectively disables any two or more simultanueously incoming connections at the kernel level, because the backlog parameter only allows 1 such connection.  Yet since terminal multiplexers can handle tens of clients, it's not technically impossible to have more than one incoming request to be received at any particular moment (for two different "lines").  A reasonable backlog of more than 1 safeguards that the "extra" connections won't be refused outright, and postpones that decision to be made by simh (not the kernel) -- for example, when, say, all multiplexer lines are busy, with an explanation (which states so) rather than just the infamous "Connection reset by peer" (effected by the kernel reset).

This patch increases the backlog to 64.

### DIFF
--- a/sim_sock.c
+++ b/sim_sock.c
@@ -1088,7 +1088,7 @@ if (!(opt_flags & SIM_SOCK_OPT_BLOCKING)) {
     if (sta == SOCKET_ERROR)                            /* fcntl error? */
         return sim_err_sock (newsock, "setnonblock");
     }
-sta = listen (newsock, 1);                              /* listen on socket */
+sta = listen (newsock, 64);                             /* listen on socket */
 if (sta == SOCKET_ERROR)                                /* listen error? */
     return sim_err_sock (newsock, "listen");
 return newsock;                                         /* got it! */


markpizz:
This suggestion came along with discussion here: https://github.com/simh/simh/issues/1077

al20878:
Good memory @markpizz 

> This suggestion came along with discussion here: https://github.com/simh/simh/issues/1077

Indeed it did, and you seem to have implemented an extremely over-complicated solution for that in *your* simulator:

https://github.com/simh/simh/commit/49902b830e29ef987454a925694e653f42fd6370

The thing is that `listen(1)` is *technically* **a race condition**.  And as you correctly noted in _your_ commit message, most of the time, even "1" just works (because the simulator is able to wrap around to check for incoming connections quite often).  Meaning, that just setting a high enough number (I should have used `SOMAXCONN` if I was sure it existed on all platforms), is already "good enough".  Introducing a yet another command parameter, and inflating an already huge HELP with yet another "detail" is an overkill.

For "most cases", it's irrelevant which parameter was used for `listen`, 1, or 64, or 128...  The aim for using a higher number is to lessen the chance of experiencing an unexplained connection reset.  No need for the user to be concerned / involved in the process, which means that "`backlog=n`" is not a command option that would be needed.

The logic in simh code does not change *a bit* with using any different backlog value -- because it can't know why the connection was available for `accept()`, was it picked up from the queue, or was it freshly incoming -- as long as the listening port is reporting there's a client, simh is going to `accept()` and then decide if the connection can proceed, or the capacity (of a multiplexer, for example) has been exceeded, and it has to be dropped.

P.S. The only valid reason to use `listen(1)`, and actually the only one I saw in code elsewhere, was in the FTP server code that expects the FTP client to pick up a data connection, when the server is switched to passive mode.

pkoning2:
Merged.

PR #316, opened 2023-10-30
CMake: Multiple targets, clean cache

bscottm:
Two updates suggested by Paul Koning:

- Enable building multiple simulator targets using a comma-separated list, e.g.: pdp8,pdp11,3b2

- Unconditionally clean CMake's cache each time the script is invoked. This eliminates confusing CMake configuration messages when CMake changes or updates, and CMake policies change.

  The most recent policy change was IMPORT_LOCATION, which is now mandatory for interface libraries. An old CMake configuration cache doesn't reflect the import library IMPORT_LOCATION property, which caused confusing warning messages. The cure is removing the old cache and reconfiguring.

### DIFF
--- a/README-CMake.md
+++ b/README-CMake.md
@@ -90,7 +90,7 @@ Before you begin building the simulators, you need the following:
     `pacman`, `rpm`, ...)
 
       apt: `sudo apt install cmake cmake-data`
-      
+
       pacman: `sudo pacman install cmake`
 
   - macOS: Install `cmake` using your preferred external package management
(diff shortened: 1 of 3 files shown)


bscottm:
The README-CMake.md patch is larger than expected because I edited it in `vim`, which I have configured to trim trailing spaces. Also added quick fix for the "-c" flag to `cmake-builder.sh` -- it takes an argument, but its spec to `getopt` didn't specify the requisite ":".

pkoning2:
Merged.

bscottm:
Thank you.

PR #319, opened 2023-11-08
PDP11: TA11/TU60 fix identification

al20878:
The pdp11_ta.c module referred to TU60 incorrectly as TA60, and this patch fixes the issue.
FTR, SimH 3.x current uses the correct notation for the TU60 cassette unit and the TA11 tape controller.


pkoning2:
Merged.

PR #320, opened 2023-11-12
Additional SDI drives for MSCP emulation, minor RQ device bugfix

cmgauger:
Added additional SDI MSCP devices (the ESE20 solid state disk, and ESE50 series of solid state disks).

Added RF35 to the drv_tab and drv_types tables, fixing an off-by-one error.

### DIFF
--- a/PDP11/pdp11_rq.c
+++ b/PDP11/pdp11_rq.c
@@ -158,7 +158,7 @@ extern int32 MMR2;
 #define UNIT_V_ONL      (DKUF_V_UF + 0)                 /* online */
 #define UNIT_V_ATP      (UNIT_V_ONL + 1)                /* attn pending */
 #define UNIT_V_DTYPE    (UNIT_V_ATP + 1)                /* drive type */
-#define UNIT_W_DTYPE    5                               /* 5b drive type encode */
+#define UNIT_W_DTYPE    6                               /* 6b drive type encode */
 #define UNIT_M_DTYPE    ((1u << UNIT_W_DTYPE) - 1)
 #define UNIT_ONL        (1 << UNIT_V_ONL)
 #define UNIT_ATP        (1 << UNIT_V_ATP)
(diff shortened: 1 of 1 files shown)


cmgauger:
Mark:

To address your points:

1. I hadn't wanted to mess with with the definition of the `DTYPE` field width, as I've not yet had a chance to look at the definition of the `UNIT` structure. But a very quick and dirty test of setting the field with to six bits does appear to work (enough so that VMS 6.2 will install to the RD33 and not break, apparently). I have committed that change and the addition of the RD33 in a second commit.

2. I have amended the original commit message to follow the SIMH style guide.

Regards,
Christian

markpizz:
Hi Christian,

Commits look good, 

The only reason the DTYPE width is changing is to add more/new disk types, so they really should be one commit with the extended commit message summarizing everything.

- Mark

cmgauger:
Mark,

It's two commits as the first just adds the ESE20 and three ESE50 models and fixes the RF35, whereas the second commit actually modifies `DTYPE` and adds in the RD33.

This way if the change to the width of `DTYPE` has silently broken something down the line, it can be backed out/reversed without having to remove the ESE20 and ESE50s (or the fix for the RF35).

Regards,
Christian

markpizz:
> This way if the change to the width of DTYPE has silently broken something down the line, it can be backed out/reversed without having to remove the ESE20 and ESE50s (or the fix for the RF35).

As you observed, it didn't break anything.  The UNIT flag word has 15 bits reserved for simulated device specific purposes.  3 of those bits are used by the sim_disk API's and thus there are 12 bits left for device specific purposes.

You added the extra bit to the DTYPE field which was at the end of the bits that the RQ device was using, so now 8 bits are being used for device specific purposes vs the prior 7.  There are still 4 unused bits.

There is no need for a separate commit since the need to add it was solely due to the set of drives you came along with info about and have added, so one commit solves the whole problem of adding these drive types.

cmgauger:
I have compacted the commits down to a single commit now; I also took the liberty to tidy up a few things, add some (dubiously?) useful information to the disk unit summary table up near the head of the file, and add the RX18 before I forget and have to make a new pull request.

PDP11/pdp11_rq.c
```diff
@@ -248,36 +248,60 @@ struct rqpkt {
...
-   RD53 17      8       1024    8       1       5*8     138672
-   RD54 17      15      1225    15      1       7*8     311200
+   type  sec  surf  cyl   tpg  gpc  RCT   LBNs    notes
+   RX18  9    1     40    9    1    -     360
+   RX50  10   1     80    5    16   -     800
+   RX33  15   2     80    2    1    -     2400
+x  RD50  17   4     153   ?    ?    ?     ?       Seagate ST-506
+   RD51  18   4     306   4    1    36*4  21600   Seagate ST-412
+   RD31  17   4     615   4    1    3*8   41560   Seagate ST-225
+   RD52  17   8     512   8    1    4*8   60480   Quantum 540, ATASI 3046, Evotek ET3540
```
pkoning2, line 258: I didn't know the RD50 is supported by MSCP.  It's a Pro device, yes, though there it has 16 sectors per track (same goes for the other RD series drives the Pro supports, because its controller is limited to a 4-bit sector number).
The parameters for RD52 are the ones for the Quantum drive; the Atasi has a different geometry (7 surfaces, 644 cylinders, don't know the sector/track count on MSCP).  And I've heard of a CDC drive DEC sold as RD52, or perhaps only planned to, that one has 5 surfaces).

cmgauger: As far as I can tell, the RD50/Seagate ST-506 is not implemented in MSCP spec (it doesn't appear in the Disk Drive/Media Identifier table or the Unannounced Disk Drive/Media Identifier table). I only include it in the reference table since it is an RD-series disk though not MSCP attached (theoretically if one wanted to add it, one could; just being mindful that the model byte and media identifier code would be fictitious, likely similar to how a third-party MSCP-based controller would report its attached drives).

With regards to the RD52, I have seen the RD52 be referred to as one of the three drives listed, and the ZRQCH0 formatting program in XXDP does explicitly list two different configurations of RD52:
- RD52 part # 30-21721-02 (1 light on front panel)
- RD52 part # 30-23227-02 (2 lights on front panel)
It's possible one of those refers to the Quantum drive, and the other to the ATASI (or Evotek) drive.

PR #321, opened 2023-11-15
PDP11, VAX, PDP10: Updates for CH11 Chaosnet device.

larsbrinkhoff:
Two updates for the CH11 Unibus Chaosnet interface:
- Make the device work for the PDP10/KS10 emulator.  It was previously untested, but now I have tested it and made fixes.
- Speed up polling when there is active traffic.

Tested with:
- PDP10 (KS10) running ITS.
- PDP11 running MINITS.


pkoning2:
Merged.

PR #323, opened 2023-11-17
PDP11: RP11: Remove duplicate checks (now only done in svc routine)

al20878:
Simplify RR device logic for the go() routine (checks were duplicated by the service routine).

Also (minor), replace a "magic" constant with the proper macro in the read and write register routines.


pkoning2:
Merged.

gtirloni:
@pkoning2 why do you say this was merged if the PR is not merged?

pkoning2:
The code is merged, but via the git command line.  Unfortunately github doesn't recognize that as a merge so I have to say it's merged and close the PR manually.  Check the code, you should find the changes are in the main branch.

PR #324, opened 2023-11-22
CMAKE: Python distutils obsoleted.

bscottm:
Python 3.12 will not have the distutils package in the standard library. The TextFile class is particularly useful when reading [Mm]akefiles, and recreating its functionality would be painful.

Stash a local copy of the last version of distutils.text_file.py and use it. Do not rely on distutils being present or installed.


pkoning2:
Merged.

bscottm:
Thank you.

PR #328, opened 2023-12-04
CMake: Updates

bscottm:
"--flavor"/"-f" flag not specified on the command line: Complain loudly, print help and exit. The script used to default to "Unix Makefiles".

Add missing "-DHAVE_LIBPNG" compiler command line define when the PNG library is detected/present for screen capture support.


bscottm:
@pkoning2 (@sethm, @XVilka) This is now a raftload of CMake-related updates:

Issue https://github.com/open-simh/simh/issues/294: "apple silicon build problem(s?)": If the "--flavor/-f" flag is not specified on the command line, then complain loudly, print help and exit. The script used to default to "Unix Makefiles".

Updates:

- Add missing "-DHAVE_LIBPNG" compiler command line define when the PNG library is detected/present for screen capture support.

- Add "clang64" to the list of MinGW64 platforms for which the  .travis/deps.sh script can install build dependencies.

- Add PThread4W_FOUND to the condition that sets async I/O for Win32 when using vcpkg for build dependencies.

- Add vs2022-x64, vs2019-x64 and vs2017-x64 build environments to build 64-bit Windows executables.

- Use simulator AIO only where needed by the simulator (i.e., the simulator calls/uses AIO_CHECK_EVENT in sim_instr())

  - Add "USES_AIO" flag to add_simulator() to mark a simulator that actually uses asynchronous I/O.

  - Build "_aio" SIMH core library variants that have AIO turned on, link with the "_aio" variant when a simulator sets USES_AIO.

  - Emit a warning message when WITH_ASYNC is False (CMake configuration option) to notify the user/developer that some functionality will be crippled.

  Affected simulator builds: 3b2 family, PDP-6, PDP-11, VAX family, IMLAC and TT2500. The makefile and cmake/generate.py also updated to remain in sync with CMake.

  N.B.: Simulators still link with the underlying platform's threading library. SEL32 requires pthreads or equivalent threading library, independent of AIO.

CMake 3.28.1 INTERFACE_LINK_LIBRARIES behavior change: The file name must now be an absolute path. Relative paths no longer accepted. Internally, SIMH enforces this if CMAKE_VERSION >= 3.19, when REAL_PATH was first implemented.

bscottm:
@sethm: "Affected" means "really needs AIO". There are only a handful of simulators that really need AIO.

PR #331, opened 2023-12-18
PDP11: fix disassembly of reg,src instructions

pkoning2:
This corrects the disassembly of MUL, DIV, ASH, and ASHC so it works correctly when the register is not R0.


al20878:
Hi Paul,
I'm trying to figure out your change, but I must be missing something...
The `masks` array is used only in instruction disassembling, and is indexed with the instruction class number.
The class number is from the range 0..19 (20 possible values), but the `masks` array has 4x6 = 24 entries.
Why is the discrepancy?
Thanks!

pkoning2:
How did you get 24 entries?  I count 20.  Remember that when you look at the "changed files" display it shows you the difference, so the red line is a deleted line and the green line is its replacement.

al20878:
> How did you get 24 entries?
```
static const int32 masks[] = {
0177777, 0177770, 0177700, 0177770,
0177700+I_D, 0177400+I_D, 0177700, 0177400,
0177400, 0177000, 0177000, 0177400,
0177400+I_D+I_L, 0170000, 0177777, 0177777,
0177000+I_D, 0177400+I_D, 0177700, 0177400,
0177000, 0177700+I_D, 0177400, 0177400+I_D+I_L
};
```
6 lines with 4 entries each, makes 24.

pkoning2:
Ok, I misread the "changed file" display as if it were the whole table.
The answer is "I have no idea".  It was that way before; the fix doesn't change that.  Perhaps @rms47 can tell us what happened.

rms47:
Table had duplicate entries for more than a year. It should be:

static const int32 masks[] = {
0177777, 0177770, 0177700, 0177770,
0177700+I_D, 0177400+I_D, 0177700, 0177400,
0177400, 0177000, 0177000, 0177400,
0177400+I_D+I_L, 0170000, 0177777, 0177777,
0177000, 0177400+I_D, 0177400, 0177400+I_D+I_L
};

The 24-entry table and this one are the same up until entry 15 (cond code set). The last four should be:

16. SOPR - single operand, register - XOR
17. FOPA - single flt single/double operand, faccum - fadd etc - moded for floating length
18. SOPA - single operand, faccum - LDEXP
19. SMDA - single int/long operand, faccum - moded for both integer length and floating length

I broke it at the end of 2022, when I added FOPA, SOPA, SMDA to account for all the src -> dst floating point cases:

   31-Dec-22    RMS     Floating loads are src,dst (nickd4)

and it got copied, verbatim, into V4.

al20878:
Thanks, @rms47!

@pkoning2 , it looks like the version of the array that Bob says is correct does not agree with your patch...

Can you please review and re-patch?

rms47:
> Thanks, @rms47!
> 
> > SOPR - single operand, register - XOR
> 
> Although, XOR is a RSOP operation (as well as JSR), 10. SOPR is for other arithmetic EIS ops, like MUL, DIV, ASH(C).
> 
True that. I checked that it worked for the EIS instructions and the fl ops that were added.

al20878:
@pkoning2 should I open another issue for this to be fixed?

Thanks!

al20878:
@rms47 : I checked http://simh.trailing-edge.com/sources/current/PDP11/ and pdp11_sys.c is still not fixed there

```
static const int32 masks[] = {
0177777, 0177770, 0177700, 0177770,
0177700+I_D, 0177400+I_D, 0177700, 0177400,
0177400, 0177000, 0177000, 0177400,
0177400+I_D+I_L, 0170000, 0177777, 0177777,
0177700+I_D, 0177400+I_D, 0177700, 0177400,           // <== This line should be deleted, erroneous copy of line 2
0177000, 0177700+I_D, 0177400, 0177400+I_D+I_L
};
```

PR #336, opened 2023-12-30
SCP: Adjust timer calibrations if SET THROTTLE x/t type is enabled

deltecent:
SCP doesn't initially adjust timer calibrations if SET THROTTLE x/t type is enabled causing large delays in event dispatches. This PR runs the timers through the same calibration adjustment loop as other throttle types.


deltecent:
This seems to fix a problem I was noticing with AltairZ80 simulator when using `SET THROTTLE x/y`. If anyone else uses this feature with other simulators, please test.

deltecent:
@markpizz says he is working on a fix for this.

PR #337, opened 2023-12-30
ALTAIRZ80: Adds JAIR devices to simulator

deltecent:
Josh's Altair/IMSAI Replacement (JAIR) is a single board computer (SBC) for the S100 bus. I created this device to assist a friend with getting IMP and CBBS working on his JAIR. Without having a JAIR myself, AltairZ80 to the rescue!

The following devices are added:

JAIR   - The main JAIR board I/O and ROM
JAIRS0 - Serial Port 0 (COM1)
JAIRS1 - Serial Port 1 (COM2)
JAIRP  - Parallel Port

This was a bit challenging because the JAIR uses an on-board SD card with FAT file system to hold its BIOS and CP/M disk images. The ATTACH command is used to mount SD card images to the simulator. The simulator emulates the SD card interface for read/writing SD card sectors. These images are easily mounted on a Mac making moving files around easy. I do not know about Windows. These same images can be written to an SD card and used with real JAIR hardware, which may be useful for JAIR owners.

The serial and parallel ports fully support TMXR. Host serial ports and sockets may be attached to these devices.

I am working on a GitHub repository containing an SD card image and init script that will boot CP/M 2.2 on the JAIR simulator.

Is this device one that should be added to AltairZ80?


deltecent:
I've added CP/M files to GitHub at [https://github.com/deltecent/jair](url)

psco:
Looks good and should definitely be added. Thank you @deltecent !

PR #340, opened 2024-01-05
ETHER: Make sure SET NOASYNC is effective for Ethernet devices

markpizz:
SET NOASYNC is intended to disable asynchronous threads from interacting with the simulator event system.

The only cases where there asynchronous threads are:
1) Disks which are designed for asynchronous support using sim_disk.  The only case is the MSCP controller implemented by pdp11_rq (Device name RQ).
2) Tapes which are designed for asynchronous support using sim_tape.  The only case is the TMSCP controller implemented by pdp11_tq (Device name TQ).
3) Devices which use sim_ether to interface with Ethernet networks.  There are a number of these cases.

Prior to this change, SET NOASYNC only actually affected the RQ and TQ devices and had no effect on Ethernet devices.

### DIFF
--- a/VAX/vax_xs.c
+++ b/VAX/vax_xs.c
@@ -130,7 +130,7 @@ DEVICE xs_dev = {
     1, DEV_RDX, 20, 1, DEV_RDX, 8,
     NULL, NULL, &xs_reset,
     NULL, &xs_attach, &xs_detach,
-    &xs_dib, DEV_DEBUG | XS_FLAGS, 0,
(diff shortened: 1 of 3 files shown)


hharte (approved):
Thanks Mark,

The change looks good, and it makes sense for the Ethernet to behave similarly to the other async (TQ and RQ) devices.

markpizz:
Minimal change to only sim_ether would be effective if it occurred before sim_ether using device code was not yet attached.  This most recent change added an Ethernet device check for the device attached state before changing the asynchronous mode.

PR #344, opened 2024-01-09
Add VAXServer3100 (KA41-E)

kkaempf:
This PR adds support for VAXServer 3100 (similar to VAXStation 3100).

This was tested to boot VMS 5.5-2 from an existing disk image.
Firmware is available here: [even](https://github.com/larsbrinkhoff/DECROMs/blob/master/23-118E8.bin) [odd](https://github.com/larsbrinkhoff/DECROMs/blob/master/23-119E8.bin)


pkoning2:
I noticed you didn't update the Cmake control files for the new target.  Can you do that, or perhaps @bscottm can help.

markpizz:
There is already support for the VAX Server 3100 (aka KA41-E).

This functionality should be available in the existing microvax3100e simulator simply by entering "SET CPU MODEL=VAXSERVER".

This command works, but the results don't actually boot cleanly and hangs before getting to the >>> prompt.

This bug should be fixed rather than adding an additional ROM specifically for the KA41-E.  The ROMs differ by a total of 5 bytes I think.  Most of the various VAX simulators have patches to the ROMs to disable various things which the ROMs test that aren't implemented.  For some reason this ROM doesn't and the fix should probably involve some ROM patching.

pkoning2:
Well, that is unfortunate; I wonder how much time was spent on this item that could have been avoided.  Is there a documentation hole that should be fixed?

kkaempf:
> Well, that is unfortunate; I wonder how much time was spent on this item that could have been avoided. 

No worries. It was a nice opportunity for me to start contributing to simh.

Will look to @markpizz's suggestion now 😉

kkaempf:
@pkoning2 I believe to have fixed the cmakefiles in my original branch 🤞🏻

PR #346, opened 2024-01-14
AltairZ80: Adds value required to SET CPU MEMORY

deltecent:
HELP CPU will show that SET CPU MEMORY requires a value. A helpful message will display if an invalid value is entered.


deltecent:
Going to wrap this PR into a general memory PR.

PR #350, opened 2024-01-31
PDP11: RP11: Make sure to advance DA after every I/O

al20878:
It looks like disk controllers, which automatically update disk address (DA) after completion of I/O, are expected to do so even if there was no data transfer because of I/O errors. I was studying RSX-11's Error Logger documentation and examples are clearly offsetting disk addresses backwards by one when I/O errors are reported by the controller.

Since once the controller has found the DA-specified sector, the I/O begins regardless of the condition of the sector (bad or good data) or ability to transfer the contents between the disk and the memory.  If an error occurs (NXM, for instance) the operation would stop (with the error reported) at the end of the sector.  So if, for example, the bus address register had a bad address from the get-go, no data would be able to transfer at all, yet DA should still be updated with DA + 1 once the controller asserts the DONE bit.

This patch makes sure that DA is always advanced when I/O has actually been commenced.

### DIFF
--- a/PDP11/pdp11_rr.c
+++ b/PDP11/pdp11_rr.c
@@ -927,8 +927,8 @@ static t_stat rr_svc (UNIT *uptr)
     cyl  = uptr->CYL;
 
     if (wr) {
-        if ((wloa & RPWLOA_ON)  &&  !rper/*valid DA*/  &&
-            (n <= GET_WLOADRV(wloa)  ||  cyl <= GET_WLOACYL(wloa))) {
+        if ((wloa & RPWLOA_ON)  &&  !rper/*valid DA*/
+            &&  (n <= GET_WLOADRV(wloa)  ||  cyl <= GET_WLOACYL(wloa))) {
             uptr->STATUS |= RPDS_WLK;                   /* DA write-locked */
             rper |= RPER_WPV;
         } else if (uptr->flags & UNIT_WPRT)             /* write and locked? */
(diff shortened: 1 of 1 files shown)


pkoning2:
It should be possible to confirm your starting statement from either the manuals or the engineering drawings.  I wonder if someone could do that.

Is this issue specific to the RP11 emulation, or does it apply more generally to other DMA devices as well?

al20878:
For RP11 it is definitely the case.  I am sure that is the case with all other drives, which are supposed to track post-I/O DA, just because how it is handled in software (e.g. Error Logger in RSX, where I noticed that)...

The problem, however, for all other disk implementations in the simulator, is that when there was no actual transfer, the registers are not updated, instead of pointing to the next location (sector or word), after encountering an error.

RP11 mimics the error tracking very closely to the real H/W...  There's a lengthy comment in code about how, and why.  Other implementations are fast-tracking errors, meaning their error reporting may be inconsistent with how the real H/W would have handled the situation.  It's not so much about the data errors (as obviously those would be extremely rare in simulation) but other conditions such as buffer overrun (and NXM) or buggy write-checks.

Having said all the above, is there anything that prevents this patch to be pushed?

It has been verified to work with XXDP, X11/RPAM, and DOS/Batch, without showing any new errors.

pkoning2:
No issue other than "I have to get to it" -- I was just asking the question to understand better.

markpizz:
> RP11 mimics the error tracking very closely to the real H/W... There's a lengthy comment in code about how, and why. Other implementations are fast-tracking errors, meaning their error reporting may be inconsistent with how the real H/W would have handled the situation. It's not so much about the data errors (as obviously those would be extremely rare in simulation) but other conditions such as buffer overrun (and NXM) or buggy write-checks.

You've addressed error paths, by digging into how the OS tracks error details and possibly by testing with a contrived some sort of error that really wouldn't happen with normal software in simulation.  The simh goals have normally been to provide functionality for legacy software.  The goals have explicitly not been to provide precise error reporting since actual errors do not occur in simulation, or if real errors somehow do happen in the host OS somewhere below the a given device's activities, what those theoretical errors are, generally won't have precise mapping to simulated device errors and these things will be impossible to test.

> Having said all the above, is there anything that prevents this patch to be pushed?

You wrote the RP11 device simulator, so there should be no reason not to adopt your changes.

PR #351, opened 2024-02-05
Added an option for Arch Linux to .travis/deps.sh

PurpleAce88:
Added an option that uses pacman to install everything, therefore making setup simpler for Arch users.

### DIFF
--- a/.travis/deps.sh
+++ b/.travis/deps.sh
@@ -8,6 +8,14 @@ install_osx() {
     brew install vde
     brew install cmake gnu-getopt coreutils
 }
+install_arch() {
+    sudo pacman -S pkgconf
+    sudo pacman -S pcre libpng libedit
+    sudo pacman -S mesa
+    sudo pacman -S libsm
+    sudo -S cmake
+
+}
 
 install_linux() {
     sudo apt-get update -yqqm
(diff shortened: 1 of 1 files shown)


pkoning2:
I've never heard of ArchLinux.  Is that some obscure packager?
It seems odd for the OS name to be "arch" when there is also "linux".  Can it check the packager name if you say "linux" and do the incompatible magic?  If not, "arch_linux" or "arch-linux" seems like a better name.

PurpleAce88:
Archlinux is a minimalist Linux distribution, its default package manager is pacman. Yes it may be able to check the package manager you're using, but I thought it best to keep things simpler and just require the user to specify Arch. Yes it may be better to name it arch-linux or arch_linux, although I thought it would be distinct enough, as I only know of one operating system that is named Arch.

pkoning2:
The reason I suggested arch-linux is that it makes it clear it's a flavor of Linux, rather than a distinct operating system like MacOS or FreeBSD or VMS.

PurpleAce88:
Good point, I can fix that

markpizz:
One thing about this PR is the personal name on Github of @PurpleAce88 .  It is hard for me to believe that "Ender Wiggin" is actually your name.   

Wikipedia says:  
> Andrew "Ender" Wiggin is a fictional character from Orson Scott Card's 1985 science fiction novel Ender's Game and its sequels, as well as in the first part of the spin-off series, Ender's Shadow. The book series itself is an expansion of Card's 1977 short story "Ender's Game."

Generally, we'd like to see the real name of the folks who submit changes.  Another PR hasn't been merged for this very reason....

PurpleAce88:
Oh. I should probably change that. You are correct my name is not Ender Wiggin. I have changed my profile accordingly to display my correct name.

pkoning2:
Thanks, so could you close this PR and resubmit it with your real name on the commits?

PR #354, opened 2024-02-18
CMake: generate.py 'experimental' rule flexibility

bscottm:
Requested change: Successfully process makefiles without the 'experimental' rule and fewer simulators than the full open-simh packaging.


bscottm:
@pkoning2: Still working with GH re: CMake 3.27.9 issue.

bscottm:
Updated set of commits.

PR #358, opened 2024-02-22
CMake: Make generate.py reusable

bscottm:
Make generate.py resuable outside of open-simh, as suggested and motivated by Richard Cornwell's simulator repository.

- Make the "experimental" rule optional. Do not generate a Python "KeyError" if the rule is missing.

- Add documentation on how to use the CMake infrastructure outside of open-simh: Customize the packaging.py script, season to taste.

- Update the KA10 simulator customization, moving it to its own Python script, simgen/pdp10_simulator.py. Preparatory move that anticipates additional frontpanel and display options.

- generate.py option "--skip-orphans": Skip the orphaned simulator check (i.e., don't cross-reference the simulators in packaging.py with what was scraped from the makefile.)


bscottm:
Moved the `cmake/` subdirectory changes in the `ibm1130-cmake` branch to this PR.

PR #360, opened 2024-02-23
CMake: Reduce excess quoting

bscottm:
cmake/cmake-builder.ps1 added quotes to arguments that contained spaces so that arguments printed correctly for progress output. This introduced excess quotes that caused CMake (and likely other MS apps) confusion or argument misinterpretation.

Instead of CMake seeing a single "Visual Studio 17 2022", CMake was actually seeing "\"Visual Studio 17 2022\"".

This patch only adds the additional quotes when reporting progress. Otherwise, command line arguments are passed unmolested.

### DIFF
--- a/cmake/cmake-builder.ps1
+++ b/cmake/cmake-builder.ps1
@@ -495,14 +495,14 @@ foreach ($phase in $scriptPhases) {
             Write-Host "** ${scriptName}: Configuring and generating"
 
             $phaseCommand = ${cmakeCmd}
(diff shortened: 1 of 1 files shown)


bscottm:
@pkoning2: This is a hotfix for the recent issues with compiling with cmake/cmake-builder.ps1.

bscottm:
@pkoning2: Waiting for macos-11 to complete. VS 2022 CMake builds complete successfully.

PR #362, opened 2024-03-01
Github CI/CD: Bump action version from v3 to v4.

bscottm:
Github upgraded Node.js from version 16 to version 20; Node.js drives the CI/CD pipelines. Github strongly recommends bumping the actions' version number from "@v3" to "@v4" and produces copious warning messages in the CI/CD "Annotations" pane for each submission.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,11 +19,16 @@ jobs:
   makefile:
     runs-on: ${{ matrix.os }}
     strategy:
+      #-
+      # The makefile builds DO NOT produce artifacts (*). Using the "-latest"
+      # runner images is sufficient.
+      #
+      # (*) "artefact" for the rest of the Anglosphere
+      #-
+
       matrix:
-        ## macos-10.15: Runner seems to be officially disabled, results in canceled
-        ## builds. Leaving this "as-is".
-        ##
-        os: [macos-12, ubuntu-20.04]
+        os: [macos-latest, ubuntu-latest]
+
         simulators:
           # These are supposed to match ALL in makefile.
           # Each job builds ~15 simulators.
@@ -34,7 +39,7 @@ jobs:
           - besm6 imlac tt2500 microvax3900 microvax1 rtvax1000 vaxstation3100m76 vaxstation4000m60
           - scelbi 3b2 i701 i704 i7010 i7070 i7080 i7090 sigma uc15 i650 sel32 intel-mds ibm1130
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       ## Workaround for remnant symlinks in /usr/local pointing back to
       ## macOS frameworks.
       ##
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         ## macos-10.15: Runner seems to be officially disabled, results in canceled
```
pkoning2, line 23: Is this comment now obsolete?

.github/workflows/cmake-builds.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04]
```
pkoning2, line 12: Line 26 in build.yml now says `ubuntu-latest`, does that apply here too?

.github/workflows/cmake-builds.yml
```diff
@@ -32,20 +32,20 @@ jobs:
...
 
 
   cmake-macOS:
-    name: macOS
+    name: macOS 12+
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-12, macos-11]
+        os: [macos-12, macos-13, macos-14]
```
pkoning2, line 46: In the other file you have `macos-latest`, would that make sense here?

Also, what does it mean to list several OS versions?  Does it run on all of them, or just on whichever one (if any) is available?

.github/workflows/build.yml
bscottm: I'll take it out. :-)

bscottm:
- `build.yml` is just the `makefile` builds. There are no artifacts produced, so using a `-latest` image suffices.
- `cmake-builds.yml`: This produces a variety of artifacts, so the specific image names specify out the OS es for which binary artifacts are produced.

The two matrices could be congruent, but I don't really see a compelling reason to do so.

.github/workflows/cmake-builds.yml
bscottm: cmake-builds.yml produces artifacts, so the idea here is to list out the operating systems for which artifacts are produced. There are two Ubuntu images in the GH image runner collection (one of which is "ubuntu-latest"), so I explicitly spelled out the versions.

.github/workflows/cmake-builds.yml
bscottm: See previous comment.

bscottm: Artifacts, artifacts, artifacts!!!

bscottm: BTW: The image name, e.g. "macos-14", appears in the artifact name. So, "-latest" might be awkward.

pkoning2: I understand the difference now, that makes sense.  A comment might help.

PS. "Artefact" is the original spelling, still used in Britain and Australia and also how the corresponding word is spelled in several other languages.  :-)

bscottm: Duly noted.

bscottm:
Notes added to build.yml and cmake-builds.yml re: runner image names.

PR #363, opened 2024-03-03
AltairZ80: Enable SIO interrupt service event after BOOT

deltecent:
The `boot` command causes all devices to reset which removes the `SET SIO INTERRUPT` service routine from the SIMH event queue which disables interrupt processing. The user must escape back to the `simh>` prompt and `SET SIO INTERRUPT` again after boot. An example of the problem:

```
at dsk altdos.dsk
set sio interrupt
boot dsk

MEMORY SIZE? << Enter >>
INTERRUPTS? Y << Enter >>
HIGHEST DISK NUMBER? << interrupt never happens, program hangs >>
```

This PR has sio_reset put the SIO interrupt service routine back on the event queue if interrupts have been enabled with `SET SIO INTERRUPT`.

```
at dsk altdos.dsk
set sio interrupt
boot dsk

MEMORY SIZE? 
INTERRUPTS? Y
HIGHEST DISK NUMBER? 1
HOW MANY DISK FILES? 1
HOW MANY RANDOM FILES? 1

056943 BYTES AVAILABLE
DOS MONITOR VER 1.0
COPYRIGHT 1977 BY MITS INC
.
```

### DIFF
--- a/AltairZ80/altairz80_sio.c
+++ b/AltairZ80/altairz80_sio.c
@@ -581,6 +581,8 @@ static t_stat sio_reset(DEVICE *dptr) {
             if (TerminalLines[i].conn)
                 tmxr_reset_ln(&TerminalLines[i]);
     mapAltairPorts();
+    if (sio_unit.flags & UNIT_SIO_INTERRUPT)
+        sim_activate(&sio_unit, sio_unit.wait);             /* activate unit    */
     return SCPE_OK;
 }
 


markpizz:
Hi @deltecent and @psco ,

This thinking here is absolutely on the right track.  I can't look at the code right now (I'm recovering from bilateral knee replacements), but consider these facts while working through this:

1 - how many times, and under what conditions is the SIO device reset routine called.  Off the top of my head, there are probably 2 or maybe 3 times:
2 - one absolute time is when the simulator initially starts (aka power up).  During this call the -P bit in sim_switches is set, which can potentially be used to influence the logic in the reset routine.
3 - another absolute time is as a result of a BOOT command.  This time the -P bit in sim_switches will not be set.
4 - another possible time might be as a consequence of an ATTACH command.
5 - another possible time might be if there is some instruction/sequence which might trigger this by a running simulator program.

Your SET SIO INTERRUPT command might merely record this in a SIO REG variable, and the previously set value can be referenced by the reset routine...

- Mark

deltecent:
@markpizz, If I understand you correctly, I believe that's what the PR does:

```
    if (sio_unit.flags & UNIT_SIO_INTERRUPT)
        sim_activate(&sio_unit, sio_unit.wait);             /* activate unit    */
```

I hope you have a speedy recovery.

PR #365, opened 2024-03-14
IBM 1130: GUI resource file, RegSanityCheck fix

bscottm:
- Add the missing ibm1130.rc GUI resource file to the Windows build so that the GUI renders correctly.

- Add "TEST_ARGS" argument to CMake's add_simulator function so that the IBM 1130 simulator can pass to "-g" on the command line to disable the GUI when running RegisterSanityCheck, i.e.:

    ibm1130 RegisterSanityCheck -g

  This fixes an edge case Heisenbug encountered during Github CI/CD tests where ibm1130 appears to hang indefinitely on the Windows runners.

  The cause is the GUI's Pump() thread function being prematurely terminated before all GUI resources are acquired: `RegisterSanityCheck` completes and starts terminating the simulator, which prejudiciously terminates the Pump() thread. The Pump() thread, in the meantime, can be terminated in the middle of resource acquisition due to thread scheduling latency. The net result is an infinite loop in the MS C runtime trying to exit the process with unstable internal state. (Separate patch: synchronization across main and Pump() threads to ensure resource acquisition completes.)

  This issue never shows up (cannot show up) on non-Windows platforms or the SIMH makefile.

- cmake/generator.py, cmake/simgen: Add a "test_args" keyword argument to the BasicSimulator constructor that holds the tests argument parameter emitted as the "TEST_ARGS" argument to a simulator's add_simulator(). Ensure that the IBM 1130 emits 'TEST_ARG "-g"' in its add_simulator().

- scp.c: reset_all_p() adds 'P' to the existing switches, versus saving sim_switches and ONLY setting the 'P' power-up reset switch. Net effect is that the IBM 1130 simulator actually sees the 'G' flag that inhibits the GUI during the console device reset.

- Ibm1130/ibm1130_cr.c

  - Fix printf() warnings (format should be long, not int)
  - Signed/unsigned mismatch, size_t for array indexing
  - Comment out the unused trim() function.

- Ibm1130/ibm1130_cpu.c, ibm1130_gui.c: Remove undefined static functions.

### DIFF
--- a/Ibm1130/ibm1130_cpu.c
+++ b/Ibm1130/ibm1130_cpu.c
@@ -246,8 +246,6 @@ extern UNIT cr_unit, prt_unit[];
 #  define ARFSET(v)                         /* without GUI, no need for setting ARF */
 #endif
 
-static void   init_console_window (void);
-static void   destroy_console_window (void);
 static t_stat view_cmd (int32 flag, CONST char *cptr);
 static t_stat cpu_attach (UNIT *uptr, CONST char *cptr);
 static t_bool bsctest (int32 DSPLC, t_bool reset_V);
--- a/Ibm1130/ibm1130_cr.c
+++ b/Ibm1130/ibm1130_cr.c
@@ -728,7 +728,7 @@ extern int cgi;
 static int16 ascii_to_card[256];
 
 static CPCODE *cardcode;
-static int ncardcode;
+static size_t ncardcode;
 static FILE *deckfile = NULL;
 static char tempfile[128];
 static int any_punched = 0;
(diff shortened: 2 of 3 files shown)


bscottm:
@pkoning2: Not an urgent PR, but it does solve the occasional Windows builder test failures.

pkoning2:
Does the generator.py change depend on the changes in the earlier PR https://github.com/open-simh/simh/pull/358 ?

bscottm:
> Does the generator.py change depend on the changes in the earlier PR #358 ?

I don't think so, but I'll check it out. If so, I'll revise PR #358 accordingly.

bscottm:
@pkoning2: Revised the two PRs to make them more independent.

PR #370, opened 2024-03-20
VIDEO: open latency, spurious unrecognized events

bscottm:
Open latency: Encountered in Windows Subsystem for Linux (WSL). WSL graphics initialization takes more than 20x100ms (2 seconds).  It's a non-deterministic delay accrued across OS thread scheduling, WSL's window emulation system and the underlying video event processing loop's execution.

Accomodate the non-determinism via a condition variable and associated mutex to signal EVENT_OPEN's completion.

  - The EVENT_OPEN-generating thread initializes a startup structure that contains the condition variable and mutex, which is passed to the EVENT_OPEN event. The thread then waits for the condition to signal via SDL_CondWait().

  - The EVENT_OPEN event processing queues EVENT_OPENCOMPLETE before returning.

  - EVENT_OPENCOMPLETE signals the condition variable and the SDL_CondWait()-ing thread resumes.

Unrecognized events: PDP-11's VT demo mode ("lunar lander") generates messages when emitting the bell (beep).  These messages are attributed to a SDL issue (presumably prior to SDL2 2.0.6.)

EVENT_REDRAW optimization generates these spurious messages via the vid_get_event_window() function, which reports an unrecognized SDL event if it cannot associate a VID_DISPLAY with an event's window identifier. Not all SIMH video events require a non-NULL VID_DISPLAY pointer.

  - Separate unrecognized event reporting from vid_get_event_window(). If a function expects a valid VID_DISPLAY ptr and gets NULL, then it should report the error via reportUnrecognizedEvent().

  - vid_video_events(): Simplified -- process one event at a time. A function dispatch table drives SIMH video event processing.

  - Separate redraws from SIMH events so that SDL_PeepEvents() filters ONLY on redraw events. Simplifies coalescing duplicate redraw events for the same window.

  - Redraw frame rate: Coalesce redraw requests if they arrive faster than human perception (30 fps) into a deferred redraw event queued by a SDL_AddTimer() callback. Skip window refresh requests if a deferred redraw request is pending.

Functional changes:

- vid_active: Changed to SDL_atomic_t (from int) to ensure correct cross-thread writes. Actually encountered on AARCH64.

- vid_is_active(): vid_active accessor function.  Updated references in VAX/vax4xx_{va,vc,ve}.c, VAX/vax_{va,vc}.c.

- VID_DISPLAY::vid_ready: Changed to SDL_atomic_t (from t_bool) to ensure correct cross-thread writes. Actually encountered on AARCH64.

- sim_video_system_init(): Consolidated SDL2 initialiation function. Allocates the simh_user_event and simh_redraw_event identifiers, initializes the event and window event debug strings, calls SDL_Init().

- Use the newer audio API when SDL version >= 2.0.6, as the SDL documentation strongly suggests and future-proof the code.

- vid_beep_cond, vid_beep_mutex: SIMH is supposed to wait until the entirety of a audio bell plays out. Instead of using SDL_Delay(), use condition variable and mutex to wait exactly the duration needed.

Cosmetic changes:

- Window event, event and user event debugging functions: Initialization code moved out of vid_video_events(). reportUnrecognizedEvent() reports developer-friendly output.

- simh_user_event, simh_redraw_event: SDL_RegisterEvents()-generated event identifiers for SIMH events, vice using the SDL_USEREVENT constant. SDL_RegisterEvents() is the preferred method for defining application-specific events (RTFM SDL_PushEvent().)

### DIFF
--- a/VAX/vax4xx_va.c
+++ b/VAX/vax4xx_va.c
@@ -1285,7 +1285,7 @@ if (dptr->flags & DEV_DIS) {
         return SCPE_OK;
     }
 
-if (!vid_active)  {
+if (!vid_is_active())  {
     r = vid_open (dptr, NULL, VA_XSIZE, VA_YSIZE, va_input_captured ? SIM_VID_INPUTCAPTURED : 0);/* display size & capture mode */
     if (r != SCPE_OK)
         return r;
@@ -1382,7 +1382,7 @@ return cpu_set_model (NULL, 0, (val ? "VAXSTATIONGPX" : "MICROVAX"), NULL);
 
 t_stat va_set_capture (UNIT *uptr, int32 val, CONST char *cptr, void *desc)
 {
-if (vid_active)
+if (vid_is_active())
     return sim_messagef (SCPE_ALATT, "Capture Mode Can't be changed with device enabled\n");
 va_input_captured = val;
 return SCPE_OK;
--- a/VAX/vax4xx_vc.c
+++ b/VAX/vax4xx_vc.c
@@ -497,7 +497,7 @@ if (dptr->flags & DEV_DIS) {
         return SCPE_OK;
     }
 
-if (!vid_active && !vc_active)  {
+if (!vid_is_active() && !vc_active)  {
     r = vid_open (dptr, NULL, VC_XSIZE, VC_YSIZE, vc_input_captured ? SIM_VID_INPUTCAPTURED : 0); /* display size */
     if (r != SCPE_OK)
         return r;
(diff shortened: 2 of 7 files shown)


bscottm:
@pkoning2: Apologies for the larger-than-normal merge request. One thing led to another -- WSL window latency causing seg faults and tracking down the spurious unrecognized event messages. It appears larger due to code shifting, primarily the message array initialization.

bscottm:
@pkoning2: One thing that still bugs me a bit -- the 400ms (give or take) wait for the "beep" to complete. I've preserved that behavior, but it just doesn't seem correct behavior. The "beep" playout should be smarter to keep track of pending beeps, introduce a small pause between pending beeps and let the rest of the video support do its thing (like painting the lunar explosion.)

bscottm:
Pulling this back to address thread data races identified by the Clang sanitizer. SDL is highly threaded underneath its implementation and there is a lot of cross-thread data sharing inside of SIMH as a result.

Issue #382, opened 2024-06-02
Stack overflow in i650 emulator


Rhialto:
- #### Context
It has been a while since I looked at the results from the self-tests / diagnostics (it seems that the makefile just continues even after a failure, which makes failing tests somewhat unapparent). Since the last time I did, I upgraded to the new NetBSD 10, which also comes with a newer gcc of course. But today I noticed that the i650 emulator segfaults in its selftest.

Looking at it with gdb, the cause seems weird, since a function is called with NULL, but in the caller this variable looks totally valid and not NULL.

I was wondering if this could be be because I now use gcc 10, which I usually didn't before.
This is on NetBSD 10 for amd64. gcc version 10.5.0 (nb3 20231008).

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue
I built myself from commit 43963943132d393b074ca4f68c0dce707f194b32 but with some small changes to the makefile, which are reflected in the listed build command (add /usr/X11R7/lib and include if present, use libpng16, use -Wl,-R instead of -R).
 ```
murthe.8:.../cvs/other/open-simh$ BIN/i650 

IBM 650 simulator Open SIMH V4.1-0 Current        git commit id: 2e301487
sim> show version
IBM 650 simulator Open SIMH V4.1-0 Current
    Simulator Framework Capabilities:
        64b data
        32b addresses
        no Ethernet
        Idle/Throttling support is available
        Virtual Hard Disk (VHD) support
        FrontPanel API Version 12
    Host Platform:
        Compiler: GCC 10.5.0
        Simulator Compiled as C arch: x64 (Debug Build) on Jun  2 2024 at 21:23:09
        Build Tool: simh-makefile
        Memory Access: Little Endian
        Memory Pointer Size: 64 bits
        Large File (>2GB) support
        SDL Video support: No Video Support
        PCRE RegEx (Version 8.45 2021-06-15) support for EXPECT commands
        OS clock resolution: 1ms
        Time taken by msleep(1): 20ms
        OS: NetBSD murthe.falu.nl 10.0 NetBSD 10.0 (GENERIC) #0: Sat May 18 22:55:37 CEST 2024  rhialto@murthe.falu.nl:/mnt/scratch/scratch/tmp/xcrash/usr/src/sys/arch/amd64/compile/GENERIC amd64
        tar tool: bsdtar 3.4.0 - libarchive 3.4.0 zlib/1.2.13 liblzma/5.2.4 bz2lib/1.0.8
[...]

PR #383, opened 2024-06-03
Reduce stack usage

Rhialto:
See https://github.com/open-simh/simh/issues/382

### DIFF
--- a/I650/i650_sys.c
+++ b/I650/i650_sys.c
@@ -843,7 +843,7 @@ void deck_print_echo(uint16 * DeckImage, int nCards, int bPrint, int bEcho)
         sim_trim_endspc(line); 
         // echo on console (add CR LF)
         if (bEcho) {
-            for (i=0;i<(int)strlen(line);i++) sim_putchar(line[i]);     
+            for (i=0;line[i];i++) sim_putchar(line[i]);
             sim_putchar(13);sim_putchar(10);
         }
         // printout will be directed to file attached to CDP0 unit, if any
--- a/I650/i650_sys.c
+++ b/I650/i650_sys.c
@@ -705,24 +705,48 @@ int Shift_Digits(t_int64 * d, int nDigits)
            
 */
 
-// load card file fn and add its cards to 
-// DeckImage array, up to a max of nMaxCards
-// increment nCards with the number of added cards
-// uses cdr0 device/unit
-t_stat deck_load(CONST char *fn, uint16 * DeckImage, int * nCards)
+/*
+ * Allocate space for a card deck of nCards cards.
+ */
+uint16 *deck_alloc(int nCards)
+{
+    return malloc(nCards * 80 * sizeof(uint16));
+}
+
+/*
+ * Free a deck previously obtained from deck_alloc().
+ */
+void deck_free(uint16 *deck)
+{
+    free(deck);
+}
+
+// Load card file fn and add its cards to
+// DeckImage array, up to a max of MAX_CARDS_IN_DECK.
+// If DeckImagePtr is NULL, allocates a deck (for now, fixed at
+// MAX_CARDS_IN_DECK, but there is potential to make this flexible
+// and reduce the size to actually used size).
+// Uses cdr0 device/unit.
+t_stat deck_load(CONST char *fn, uint16 ** DeckImagePtr, int * nCards)
 {
     UNIT *              uptr = &cdr_unit[0];
-    uint16 image[80];    
+    uint16 * DeckImage;
+    uint16 image[80];
     t_stat              r, r2;
     int i;
     uint16 c;
 
     // set flags for read only
-    uptr->flags |= UNIT_RO; 
+    uptr->flags |= UNIT_RO;
 
     // attach file to cdr unit 0
     r = (cdr_dev.attach)(uptr, fn);
     if (r != SCPE_OK) return r;
+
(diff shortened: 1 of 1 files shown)


pkoning2:
I'm curious: why does this simulator want to read an entire card deck into memory?  I thought the usual way to emulate unit record devices is to read or write one record at a time.  Would it make sense to adopt that approach here?

markpizz:
I once had a conversation with the author of this simulator with the suggestion that instead of the observed monkey business, the logic should be revised to leverage the functionality provided by @rcornwell's sim_card API.  He was happy with what he had, so wasn't interested in the effort...

pkoning2:
I can understand not using sim_card; there are other simulators that have their own card machinery (like pdp11), perhaps because they predate the introduction of sim_card.
I don't object to reading cards with custom code rather than using sim_card.  But I do object to reading an entire deck into memory.  That sort of thing doesn't make any sense.  It's one thing to do it for an emulated drum or fixed head disk, like the RF11 emulation in pdp11.  But cards are sequential devices...

rcornwell:
Sim_card reads in all cards in a deck at one time. But it buffers them in chunks of memory. Sim_card also supports stacking of multiple decks at once.

pkoning2:
Silly me for assuming things.  But why????

rcornwell:
Because I need to pre-read things so I can figure out what format the current card is. Just easier to allocate the images into a large array and hand off cards as cards are read. Converting sims to use sim_card is not always easy. For PDP11 Bob reads cards and stores them as ASCII. Sim Card gives you card images that you have to translate into local encoding.

pkoning2:
Hm, ok.  I did an emulator with column binary reader and punch, one card at a time.  (Or two, in the case of the punch, because it has a verify station that's actually visible to the software.... a CDC 405 punch with a rather primitive controller.)

rcornwell:
I have a stand alone version of sim_card. I also have a tape reader/writer that returns/uses 1 frame at a time.

Rhialto:
> I'm curious: why does this simulator want to read an entire card deck into memory?  I thought the usual way to emulate unit record devices is to read or write one record at a time.  Would it make sense to adopt that approach here?

If I'm seeing this correctly, this code implements some commands that
manipulate card decks in files, in particular split join and print them.
For that it should not needed to read the whole deck into memory.
The test runs ./I650/sw/run_fortransit.ini which contains
```
carddeck -q split -1 cdp1 deck_it.dck deck_it_header.dck
carddeck -q join  deck_it_header.dck deck_it.dck as deck_it.dck
```

I have not checked which code is used for actual emulated
card I/O. Probably not this code, since I changed the signature of the
function that reads the deck, and that function is only used in the same
file (I650/i650_sys.c) for this `carddeck` command.

While looking I seem to have overlooked another similar array of
MAX_CARDS_IN_DECK cards (in deck_echolast_cmd()) but it doesn't load the
deck from a file there, so I didn't notice so far. I'll adjust that too.

Initially I was hoping to make allocation fully flexible and avoid the
MAX_CARDS_IN_DECK limit. But when joining decks, the code as it is loads
extra cards into a previously loaded deck. To make that possible would
add more complexity so I skipped that for now.

rcornwell:
It might be worth the while to convert this to use sim_card.

Rhialto:
`deck_load()` uses `sim_read_card()` which is in `sim_card.c` so I suppose it already uses it?

rcornwell:
Interesting, maybe he is not reading  cards individually but in bulk. sim_card should slurp the whole deck in and you can read cards via sim_read_card(). Sim_card will read deck at attach and free memory at detach or when deck is empty.

markpizz:
When I suggested to the original author that he convert to using sim_card, I realized that he already had defined some command syntax for managing possibly multiple decks of cards in various ways.  I WAS NOT suggesting that he should change his command syntax that specified this, but that he merely modify the interpretation of the syntax to conform to the ATTACH syntax of sim_card so that all the buffering activities he was doing wouldn't be needed.  @rhialto, if you're reworking things under the covers, maybe you could look more deeply at this and hopefully simplify things... :-)

Rhialto:
@pkoning2 I think so. I had a quick look and improving it to remove the local buffering is a lot bigger renovation. It will need to keep several decks open at the same time and that needs some extra card readers and punches :-)

I650/i650_sys.c
```diff
@@ -705,25 +705,49 @@ int Shift_Digits(t_int64 * d, int nDigits)
...
+    return malloc(nCards * 80 * sizeof(uint16));
+}
+
+/*
+ * Free a deck previously obtained from deck_alloc().
+ */
+uint16 *deck_free(uint16 *deck)
+{
+    free(deck);
+}
```
pkoning2, line 722: On my Mac, "make i650" fails with an error here, saying that it's a non-void function lacking a return statement.

Presumably the function type should have a void return rather than a pointer.

pkoning2 (changes requested):

I650/i650_sys.c
Rhialto: Yes, you're right... just a second...

pkoning2 (approved):

rcornwell:
I see what he is doing, you could rewrite deck_split and deck_join to process cards one at a time, rather then all at once. He has provided some routines to manage decks of cards so you can split a deck or join two decks together. Perhaps I need to provide similar routines in sim_card.

PR #386, opened 2024-06-08
IBM 1130 Fix address computation overflow.

charlesUnixPro:
In sim_instr(), the effective address is computed; for the case of TAG (index register addressing), the contents of
the specified index register is added to the effective address, but the result is not masked to 16 bits as per the
hardware functionality. 

Adding a a 16 bit mask operation fixes the issue.

### DIFF
--- a/Ibm1130/ibm1130_cpu.c
+++ b/Ibm1130/ibm1130_cpu.c
@@ -653,7 +653,7 @@ t_stat sim_instr (void)
 
             eaddr = word2;                  /* assume standard addressing & compute effective address */
             if (TAG)                        /* if indexed */
-                eaddr += ReadIndex(TAG);    /* add index register value */
+		eaddr = (eaddr + ReadIndex(TAG)) & 0xFFFF;	/* add index register value */
             if (INDIR)                      /* if indirect addressing */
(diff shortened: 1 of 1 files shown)


markpizz:
I have nothing to say about the technical merits of your suggested change.  I'm sure  you are right.

Meanwhile, the project conventions about the format of commit messages is:

affected-simulator-or-component: one-line summary of the change \<= 80 in length
\<blank-line-if-more-commit-message-is-appropriate\>
\<additional-commit-comments-less-than-80\>

The commit message should contain all the relevant explanation for the change.  The Pull Request comments might be the same, but the commit message should contain the full justification.  The reason being is that the commits exist as part of the history in every copy/clone of the repo without any dependence on the specific Github instance.

So, for this Pull Request we'd like to see the commit message be:

IBM1130: Fix address computation overflow

In sim_instr(), the effective address is computed; for the case of TAG 
(index register addressing), the contents of the specified index register 
is added to the effective address, but the result is not masked to 16 
bits as per the hardware functionality.

Adding a a 16 bit mask operation fixes the issue.

---
You can readily back out the most recent commit locally:
     $ git reset --soft HEAD^
and the commit again with the appropriate commit message and then push out to github with the --force flag that will properly update the pull request.

PR #390, opened 2024-06-15
The use of egrep is deprecated

stevefalco:
On Fedora40, I get warnings that egrep is deprecated.  Replacing "egrep" with "grep -E" eliminates the warnings, and is functionally identical.


pkoning2:
Thanks!  I thought this was already done a while ago, but apparently not.

PR #394, opened 2024-06-18
Use of uintxx_t

folkertvanheusden:
Replaced all instances of uint8, Uint8, uint16, Uint16, uint32 and Uint32 by their uintXX_t variants.
This was performed with scripting.
I did a compile run and everything compiled (using the cmake build scripts). Some things did not link but that was caused by 'vid_...'-symbols missing.

Please let me know if this works for you (a commit per directory) or if you would like it differently.


pkoning2:
It would be cleaner as a single commit for the work.  In general, the details of individual steps taken to get to the end result are not interesting and preferably not visible.

pkoning2:
Actually, the failure isn't vid_xxx missing; it's zlib_xyz missing called from vid_xxx.

I'm baffled that the build system doesn't have zlib installed, or that it's installed in a place where the makefile can't find it.  Probably the latter, actually, given that the cmake case does work.

markpizz:
This whole set of actually unneeded changes and the prior PR is due to a problem with the cmake setup missing something.

Shouldn't it be fixed?

rcornwell:
Not sure I agree with this, this is effecting a lot of code.

Rhialto:
Interestingly, AltairZ80 also has an MC68000 emulator in it. It has again it own home-grown names, and to make it worse,  it also has some weird ifdef where 64-bit types could be either 64 or 32 bits....
AltairZ80/m68k/m68kcpu.h:
```
#if M68K_USE_64_BIT
typedef signed   long long sint64_t;
typedef unsigned long long uint64_t;
#else
typedef sint32_t sint64_t;
typedef uint32_t uint64_t;
#endif /* M68K_USE_64_BIT */
```

Rhialto:
Er, I think I pasted a wrong version there. It was more like
```
#if M68K_USE_64_BIT
typedef signed   long long sint64;
typedef unsigned long long uint64;
#else
typedef sint32 sint64;
typedef uint32 uint64;
#endif /* M68K_USE_64_BIT */
```
where its own invented names are sint*N* and uint*N*.

rcornwell:
I am actually wondering about this, changing these means that things will break if stdint.h is not available. Right now the types are defined in sim_defs.h. So everyone should be using uintx and intx. I noticed that uint64 and int64 are not defined by defined as t_uint64 and t_int64. Perhaps these should be changed to uint64 and int64.

pkoning2:
Aren't t_uint64 etc. ancient 64-bit types from long before ANSI C?  I vaguely remember they are MS proprietary names.

It doesn't seem to make a whole lot of sense to deal with stdint.h not being there, because it's been around for a very long time.

markpizz:
What is actually broken in the current simh codebase?

If nothing is broken, then changing this stuff is change for the sake of change rather than to fix something.

deltecent:
There is a problem with `sim_video` caused by the DAZZLER PR I pushed that will cause a link error. PR #397 should fix it.

Rhialto:
What's broken? I guess it is probably more a matter of taste. But in my opinion, the way compatibility is done with older compilers (operating systems, etc) is done the wrong way around.

I am of the school of though which says that (in general) you should write your program to the latest standard that you want to follow. Sometimes you don't want to use the very latest one(s), that's fine, but it's not the issue. I think following at least C99 would be the thing to do by now (25 years later). Later versions are more debatable.

Now if you want to be compatible with older compilers, that is fine. But such allowances should be as localized as possible. Usually that means that if something is not available in the older compiler, you add an implementation of it, written in terms of older facilities, and put it away somewhere in a compat file.

So in the case at hand, that means using `int32_t` everywhere. For older compilers, have some typedefs that create the new names based on what's available.

Because this merge request was created very mechanically it gets that part (in sim_defs.h) slightly wrong. I would propose (for a part of it):
```
diff --git a/sim_defs.h b/sim_defs.h
index 3954ca07..679a4b7c 100644
--- a/sim_defs.h
+++ b/sim_defs.h
@@ -215,23 +215,23 @@ extern "C" {
 /* Handle the special/unusual cases first with everything else leveraging stdints.h */
 #if defined (VMS)
 #include <ints.h>
+typedef int8             int8_t;
+typedef int16            int16_t;
+typedef int32            int32_t;
+typedef uint8            uint8_t;
+typedef uint16           uint16_t;
+typedef uint32           uint32_t;
 #elif defined(_MSC_VER) && (_MSC_VER < 1600)
-typedef __int8           int8;
-typedef __int16          int16;
-typedef __int32          int32;
-typedef unsigned __int8  uint8;
-typedef unsigned __int16 uint16;
-typedef unsigned __int32 uint32;
+typedef __int8           int8_t;
+typedef __int16          int16_t;
+typedef __int32          int32_t;
+typedef unsigned __int8  uint8_t;
+typedef unsigned __int16 uint16_t;
+typedef unsigned __int32 uint32_t;
 #else
 /* All modern/standard compiler environments */
 /* any other environment needa a special case above */
 #include <stdint.h>
-typedef int8_t          int8;
-typedef int16_t         int16;
-typedef int32_t         int32;
-typedef uint8_t         uint8;
-typedef uint16_t        uint16;
-typedef uint32_t        uint32;
 #endif                                                  /* end standard integers */
 
 typedef int             t_stat;                         /* status */
```
From the way how `#if defined (VMS)` is used, I would assume that the names that are currently in use are defined in the VMS-specific file `<ints.h>`. When written, this was probably a good idea, and likely even followed the compatibility pattern that I am advocating here.

For the (u)int64_t types, which may not exist, I try to check the existence of the macro `INT64_MAX` and only if not defined, fall back to more or less the current conditional code, just with the standard names. (I expect that the fallback is rarely needed.)
```
@@ -239,43 +239,45 @@ typedef int             t_bool;                         /* boolean */
 
 /* 64b integers */
 
-#if defined (__GNUC__)                                  /* GCC */
-typedef signed long long        t_int64;
-typedef unsigned long long      t_uint64;
+#if defined (INT64_MAX)
+/* <stdint.h> already defined int64_t. */
+#elif defined (__GNUC__)                                /* GCC */
+typedef signed long long        int64_t;
+typedef unsigned long long      uint64_t;
 #elif defined (_WIN32)                                  /* Windows */
-typedef signed __int64          t_int64;
-typedef unsigned __int64        t_uint64;
+typedef signed __int64          int64_t;
+typedef unsigned __int64        uint64_t;
 #elif (defined (__ALPHA) || defined (__ia64)) && defined (VMS) /* 64b VMS */
-typedef signed __int64          t_int64;
-typedef unsigned __int64        t_uint64;
+typedef signed __int64          int64_t;
+typedef unsigned __int64        uint64_t;
 #elif defined (__ALPHA) && defined (__unix__)           /* Alpha UNIX */
-typedef signed long             t_int64;
-typedef unsigned long           t_uint64;
+typedef signed long             int64_t;
+typedef unsigned long           uint64_t;
 #else                                                   /* default */
-#define t_int64                 signed long long
-#define t_uint64                unsigned long long
+#define int64_t                 signed long long
+#define uint64_t                unsigned long long
 #endif                                                  /* end 64b */
 #ifndef INT64_C
 #define INT64_C(x)      x ## LL
 #endif
...
```
and then there are some bits with `#if defined (USE_INT64)` which just needs renaming of the types.

rcornwell:
In this case I have to agree with Mark, it aint broken, so way make massive changes to the source. SimH seems to be run on a variety of platforms so using the most generic C I think it best idea. I like the convention of uintx rather then uintx_t. I don't have any problem with switching my simulators from t_uintx to uintx. I did this define for my PDP10. The way I would handle this is to declare uint64/int64 while keeping t_uint64 and t_int64. This should only break a few emulators. Once this is done I will move my emulators over. But I would not recommending changing other emulators over. Making major changes like this where there is little functional change is not a good idea.

markpizz:
@Rhialto, rather than looking to make massive changes across all simulators that don't actually fix something broken, you seem to be wanting to contribute usefully since you jumped in with fixing the card stuff in the I650, you might want to look through the open issues in the simh/simh repo. All of those issues existed prior to the open-simh/simh fork, so they are present in the open-simh code base.  (Really long sentence, eh?).

bscottm:
> Actually, the failure isn't vid_xxx missing; it's zlib_xyz missing called from vid_xxx.
> 
> I'm baffled that the build system doesn't have zlib installed, or that it's installed in a place where the makefile can't find it. Probably the latter, actually, given that the cmake case does work.

macOS used to have its own zlib installed, but apparently, that's no longer the case and it has to be manually installed via HomeBrew via `.travis/deps.sh`. See PR #398.

bscottm:
@folkertvanheusden: Be somewhat cautious about changing `Uint32` in the SDL code (`sim_video.c`). That's an actual SDL2/3 type. It should be consistent with what SIMH thinks is a 32-bit unsigned integer.

markpizz:
Fixing the cmake based build in PR#398 now should result in the reversal of the change made PR#393.

bscottm:
@rcornwell: Having some type consistency is likely a good thing, as well as leveraging features of the C99 standard. For all intents and purposes, SIMH's `makefile` forces C99 standard compatibility (although, I tend to agree with you about generic C code, to a point. I happen to like the C99 ability to declare variables where they're needed, instead of all up at the front of the function, as well as structure initialization by member name.)

@Rhialto's suggestion to refactor the code to use `stdint.h` is a pretty good one, IMHO.

pkoning2:
Given the various comments I'd say the consensus is not to proceed with this bulk change.

PR #395, opened 2024-06-18
Types

folkertvanheusden:
Replaced instances of Uint32/uint32 by uint32_t, Uint16/uint16 by uint16_t and Uint8/uint8 by uint8_t

This branch **replaces** the 'Uint8' branch.


rcornwell:
I will repeat, this change effects way to many files, it should be rejects.

folkertvanheusden:
So, back to my initial patch?

pkoning2:
Richard, I see your point.  So does that mean we should not have a general cleanup of type names?  What should we do instead?  Just clean up selectively as we touch files?  Just clean up cases of non-standard names (like Uint)?

rcornwell:
I would recommend as we go. Note all of my simulators are just period clones from my master repository. So making changes to them here might get overwritten if I push an update. The same may be true for several other simulators like 3b2 and SEL32.

pkoning2:
Ok, that sounds like a good plan.

Sorry for the false start, Folkert.

PR #396, opened 2024-06-19
PDP11: Fix to fix disassembly #331

al20878:
As discussed in PR#331, the fix needed more fix with the proper masks array.  This patch completes the fix.


al20878:
Can I ask what's holding this up?

al20878:
@pkoning2

Can this please be pushed?  It's been fixed (like this PR shows) in SIMH-V3.12-5 (current)

markpizz:
> Can this please be pushed? It's been fixed (like this PR shows) in SIMH-V3.12-5 (current)

Actually, SIMH-V3.12-5 (current) is a pending PR in the Supnik-Current branch as described in PR #412 

Both of these (#396 and #412) along with #400, #403, #410 and #414 should all be merged.

PR #397, opened 2024-06-19
VIDEO: Add vid_render_set_logical_size() stub

deltecent:
Fixes oversite of missing sim_video stub function.


psco:
Thank you. This PR fixes the problem that currently AltairZ80 can only be compiled with video support. @pkoning2 this is good for being merged.

PR #402, opened 2024-06-30
Revert "Compile fix: replaced Uint8 by uint8."

markpizz:
The "error" that the original change "fixed" was actually in the cmake build system and not in the sim_ws.c code.

The Uint8 type is specific to SDL2 APIs.

This is now rebased on the cmake fixes so the CI/CD builds should run cleanly.


markpizz:
CI builds ran cleanly.  This can now be merged

PR #404, opened 2024-07-04
Add macports support

rbanffy:
Adds MacPorts as an option in the travis/deps.sh

### DIFF
--- a/.travis/deps.sh
+++ b/.travis/deps.sh
@@ -6,6 +6,11 @@ install_osx() {
         vde cmake gnu-getopt coreutils zlib
 }
 
+install_macports() {
+    sudo port install pkgconfig pcre libpng libedit libsdl2 freetype libsdl2_ttf \
+        vde2 cmake util-linux coreutils zlib
+}
+
 install_arch_linux() {
     sudo pacman -S --noconfirm pkgconf
     sudo pacman -S --noconfirm pcre libpng libedit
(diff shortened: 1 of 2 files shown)


.travis/deps.sh
```diff
@@ -6,6 +6,11 @@ install_osx() {
         vde cmake gnu-getopt coreutils zlib
 }
 
+install_macports() {
+    sudo port install pkgconfig pcre libpng libedit libsdl2 freetype libsdl2_ttf \
```
pkoning2, line 10: I see it's done for various other platforms, but I wonder if it is a good idea to have "sudo" in our scripts.  It seems better to tell people to do the install their way, or tell them to run the script under sudo.  That makes it explicit, rather than surprising people with doing privileged actions under the covers.

rbanffy: I thought it was reasonably harmless, as other architectures, such as Arch Linux and Debian also use sudo in their scripts. On the README.md file the instructions already tell the user that the deps.sh script should be invoked with sudo. If the user fails to invoke it with sudo, the script will call the installer using sudo and will prompt the user for a password.

pkoning2: I guess so.  It feels wrong, and I would view the existing "sudo" occurrences in scripts as defects to be fixed rather than conventions to apply in new work.  Still, that can be treated as a separate task.

rbanffy: We can submit a patch later to fix all of them. I believe it's a good idea.

pkoning2: It's now issue #420 so if you want to take that on, that would be great.

PR #406, opened 2024-07-11
SIMH: Fix spelling errors in comments and strings

psco:
Over the years the SIMH base components files have accumulated some spelling errors. This PR fixes almost 200 typos. Note that only typos in comments and string literals are taken care of.


markpizz:
I'm always a little concerned about changes that affect many files.  Historically, it was appropriate to changes this type of issue in files when other changes were made to that file.  However many of these files never change and have had these minor issues forever, so such a change would seem appropriate to be part of a global cleanup.

In your microscopic review of all of this, you might have found a couple of bugs along the way.  If you did, it seems you kept the focus of these changes to on scope "Fix spelling errors in comments and strings".  Hopefully you will later come along to address any such bugs found.

I'm good with this global cleanup.

Thanks @psco

pkoning2:
Agree with @markpizz -- thanks for this work!

markpizz:
FYI, this change includes minor sim_card changes which should find their way back into @rcornwell's sims repo...

rcornwell:
Getting time I need to resync my SCP anyway.

PR #408, opened 2024-07-14
SCP: Remove unused variable from sim_timer.c

deltecent:


psco:
Thank you! This PR also solves the problem that sometimes Visual Studio projects treat warnings such as unused variables as errors which causes compilation to fail.

PR #410, opened 2024-07-16
AltairZ80: Fix 8080 and Z80 CPU

psco:
This fixes the H flag handling for the 8080 CPU and corrects the parity flag computation for the Z80 CPU for INI, OUTI, IND, OUTD, INIR, OTIR, INDR and OTDR instructions. It is based on Thomas Eberhardt's work.


psco:
I haven't seen any objections and I think it can be merged (@pkoning2). Thank you.

AltairZ80/altairz80_cpu_nommu.c
```diff
@@ -1881,43 +1963,72 @@ t_stat sim_instr_nommu(void) {
...
                 AF = subTable[cbits & 0xff] | cbitsTable[cbits & 0x1ff] | (SET_PVS(cbits));
+                if (chiptype == CHIP_TYPE_8080)
+                    AF ^= 0x10;
                 break;
 
             case 0xa0:      /* AND B */
-                AF = andTable[((AF & BC) >> 8) & 0xff];
+                acu = andTable[((AF & BC) >> 8) & 0xff];
+                if (chiptype == CHIP_TYPE_8080 && ((AF | BC) & 0x0800) == 0)
+                    acu &= ~0x10;
```
pkoning2, line 1980: I'll admit I don't know the instruction set, but this is puzzling.  Some of the changes suggest to me that this flag isn't defined on 8080, which would explain why it's being cleared.  But if so, why is it XORed in many other places (like line 1974 above)?

psco: Thank you for taking a closer look, appreciated!

Actually the H-flag (0x10) is defined for both the Z80 and the 8080 but the computation is different. The code above makes sure that it is properly cleared on an 8080 under certain conditions. For other instructions (such as SBC where line 1974 belongs to) it is the inverse when compared to the Z80 CPU.

There is a test program EX8080.COM running under CP/M 2 that exercises most 8080 instructions and produces a CRC code for various families of instructions. With the change above (and the other similar ones) this program produces now the same CRC values regardless whether the program is run on the AltairZ80 simulator or on a real 8080 CPU.

psco: Are there any further questions / comments from anyone that need to be addressed before this can be merged?

psco:
I think this can now be merged by e.g. @pkoning2. Thank you.

PR #417, opened 2024-09-02
PDP11: RP11: Implement original RP11 controller flavor

al20878:
Studying RSX11M source code I found out that the original RP11 controller (vs. -C and -E later versions) was still supported in the software, and notably UMD (User Mode Diagnostics) was still capable of handing it and dealing with its testing, even in much later OS versions (compared to when the original RP11 was phased out).

The change in this commit basically implements that flavor (which only supported the RP02 drives), and also makes a few minor fixes / cleanups for the device code, here and there.

The default mode for the RP11 controller remains the more widely used -C/-E version, but if so desired, it can be downgraded to support its predecessor with the "SET RR \<type\>" command now.

### DIFF
--- a/PDP11/pdp11_rr.c
+++ b/PDP11/pdp11_rr.c
@@ -1,4 +1,4 @@
-/* pdp11_rr.c: RP11/-C/-E/RP02/RP03 disk pack device
+/* pdp11_rr.c: RP11/-C/-E/RP02/RPR02/RP03 disk pack device
 
    Copyright (c) 2022 Tony Lawrence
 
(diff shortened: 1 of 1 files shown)


pkoning2:
RSTS seems to like this, too, at least INIT.  I don't have an RP03 enabled monitor handy to check.  A quick code check suggests that the original RP11 controller would fail with the "overlapped seek" driver because it expects the RPSUCA register, but the non-overlapped driver doesn't look at that so it should work.

al20878:
If the original RP11 was identified as such (RSX can do, for example, with help from the operator), then the SUCA register is in fact in the upper half in the RPCA register, that's where RSX takes it from, by using a byte move (MOVB), instead of a word move for the (later) -C/-E's SUCA location.  So the overlapped seek is possible if the software is aware of the difference.  But other than that, everything else works exactly the same way, so there should be no problem for the software to work with either version.

PR #425, opened 2024-09-27
NET: Don't unconditionally turn on Ethernet debugging

bscottm:
Make Ethernet debugging (DBG_ETH, DEBUG_ETHER) conditional on the simulator's debugging bits. User must explicitly enable ETH debugging, whereas the default polluted the debug output unconditionaly.


bscottm:
Now part of the libslirp update.

PR #427, opened 2024-09-28
CMake: Regenerate

bscottm:
Periodic CMake update to sync with master branch changes.


pkoning2:
Interesting, there actually are no real changes: two files have white space changed only, and the third has a line that moved.

PR #428, opened 2024-09-28
CMake: Add sanitizer support

bscottm:
Add ASan [address], MSan [memory], TSan [thread] and UBSan [undefined behavior] to the cmake-builder scripts. These are intended for developer usage only to aid in debugging and general code improvement/diagnosis.

Linux, macOS: ASan, MSan, TSan and UBSan.
Microsoft: ASan only.


bscottm:
Part of the libslirp update. Really needed for various debugging purposes.

PR #436, opened 2024-11-26
AltairZ80: Support for disk images of Amstrad CPC 6128 with ULIfAC board

psco:


bscottm:
@pkoning2: macos12 has officially been deprecated by GH. Update is in MR #437 (yeah, I know, but since I was pushing through a MR, might as well kill many birds with the one stone.)

psco:
This is a straightforward and low risk change. @pkoning2 A merge would be appreciated. Thank you.

PR #440, opened 2025-01-14
H316: Convert IMP short leaders to long, and vice versa.

larsbrinkhoff:
This adds the modifiers CONVERT and NOCOVERT to the HI units.  When enabled for a unit, 1822 messages will transparently be converted.  IMP-to-host messages are converted from the old, short (32-bit) format to the new, long (96-bit) format.  Host-to-IMP messages are converted in the other direction.

The motivation for this feature, is that the currently running IMP software is from 1974 and only supports short leaders.  Some operating systems are from a later era, and only support long leaders.

This has been tested using the H316 emulator and the IMP code in the H316/test directory, and ITS running on a PDP-10 KA10 emulator.  I have also written a new NCP implementation that can be used to probe and test the network and hosts; it has successfully ran TELNET sessions in both directions against ITS.

### DIFF
--- a/H316/h316_hi.c
+++ b/H316/h316_hi.c
@@ -105,6 +105,8 @@ void hi_rx_local (uint16 line, uint16 txnext, uint16 txcount);
 t_stat hi_reset (DEVICE *dptr);
 t_stat hi_attach (UNIT *uptr, CONST char *cptr);
 t_stat hi_detach (UNIT *uptr);
+t_stat hi_set_convert (UNIT *uptr, int32 val, CONST char *cptr, void *desc);
+t_stat hi_show_convert (FILE *st, UNIT *uptr, int32 val, CONST void *desc);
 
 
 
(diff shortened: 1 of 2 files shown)


larsbrinkhoff:
CC @SpareTimeGizmos, @charlesUnixPro, @Quantx, @Ninjaneer321

markpizz:
The convert-1822 branch containing the commit in this change in your account has an additional commit with the comment "Debug.".  That extra commit might accidentally be picked up when this change is merged.  I only mention this since the changes in the "Debug." commit explicitly emit output to stderr rather than using the normal simh debug facilities...

pkoning2:
Random question: I noticed the data is given as 16-bit items.  Is this a 16 bit parallel device?  I'm so used to networking devices that are byte oriented, and seeing 16 bit words manipulated without "little endian" or "big endian" consideration was different...

larsbrinkhoff:
@pkoning2, the IMP-host interface (per BBN Report 1822) is bit-serial.  See e.g. https://gunkies.org/wiki/1822_interface

However, the message format has a strong 16-bit flavour since the H316 and 516 are 16-bit minicomputers.  Other than that, the 1822 and NCP messages are fairly byte size agnostic.

pkoning2:
That makes me wonder: if you compile this code on a big-endian host, will it do the right thing?

larsbrinkhoff:
h316_udp.c converts the uint16_t words back and forth with htons and ntohs.

pkoning2:
> h316_udp.c converts the uint16_t words back and forth with htons and ntohs.

Ok, but what about the header transformations you added?  Will the format recognition, and the rewriting, work correctly for both?

larsbrinkhoff:
Yes.

PR #443, opened 2025-01-26
vmnet.framework support on macOS

NattyNarwhal:
macOS has no built-in tun/tap support; third-party extensions to do so require more hoops to jump through as Apple (justifiably) makes it harder to use third-party kexts. The situation for virtual networking is thus clumsy, relying on pcap/VDE, or dealing with the limitations of slirp or the UDP tunneling. This PR adds support for [vmnet](https://developer.apple.com/documentation/vmnet?language=objc), an alternative built into the OS. It provides host-only and shared NAT layer 2 virtual networks, as well as bridging virtual interfaces to physical interfaces.

vmnet requires macOS 10.10 for basic functionality, and 10.15 for bridged networks.

Documentation and SCP help has also been updated.

This has been tested with the following scenarios on macOS 15:

- Netbooting a MicroVAX 3900 in simh from an InfoServer 150 in simh (over MOP); the system was able to install OpenVMS 7.3.
- Installing NetBSD/vax 10.1. NetBSD is able to acquire IPv4 and IPv6 addresses via DHCP, and I can telnet into the VAX via IPv6.

### DIFF
--- a/0readme_ethernet.txt
+++ b/0readme_ethernet.txt
@@ -214,6 +214,79 @@ Note: As mentioned above, NAT networking is specifically capable of providing
 
 
 -------------------------------------------------------------------------------
+Another alternative to pcap and tun/tap on macOS is using the vmnet framework.
+The vmnet framework provides simh a way to provide emulated systems networking,
+without requiring any additional software. Because tun/tap support requires
+third-party kernel extensions (which is heavily discouraged by Apple), this
+provides an equivalent that doesn't require reducing OS security.
+
+vmnet provides Layer 2 networking, so things like DECnet and MOP will work
+properly, even across multiple systems. It can be used to either bridge to a
+physical network, or provide a virtual network for systems, including NAT.
+
+simh must be running as root. (Running without root is possible if simh is a
+notarized application. You're very unlikely to be building simh like this, so
+de facto it needs root.) vmnet.framework requires macOS 10.10; macOS 10.15 is
+required for bridged networking.
+
+When attaching an ethernet device, you have three options available:
+
+  - "vmnet:host": This provides "host-only" networking. It allows the system to
+    talk to the host Mac, as well as other emulated systems on the host-only VM
+    network.
+  - "vmnet:shared": This extends the host-only network to add NAT, using the
+    host as a gateway.
+  - "vmnet:<device>": This bridges vmnet to a physical network. You must
+    specify a valid network device to use.
+
+    You can use SHOW ETHERNET in SCP to see what devices are allowed to be
+    bridged; these are likely your Ethernet and Wi-Fi devices.
+
+    sim> show eth
+    ETH devices:
+     eth0	vmnet:en0
+     eth1	vmnet:en7
+     eth2	udp:sourceport:remotehost:remoteport
+    sim> attach xq vmnet:en0
+
+    You can verify with i.e. "ifconfig" to see if this is the correct interface
+    to bridge to.
+
+When using the host and shared modes, the host system create an IPv4 and v6
+subnet for the emulated systems. macOS will provide DHCP and other discovery
+protocols on the virtual network as well. To see what IP and subnet the host
+has claimed, you can use "ifconfig" to display it; macOS will bind its IP to
+the bridgeN device, with the simh instance having its own vmenetN device joined
+to that bridgeN device:
+
+vmenet0: flags=8963<UP,BROADCAST,SMART,RUNNING,PROMISC,SIMPLEX,MULTICAST> mtu 1500
+	options=60<TSO4,TSO6>
+	ether a6:94:63:7e:f3:04
+	media: autoselect
+	status: active
+bridge100: flags=8a63<UP,BROADCAST,SMART,RUNNING,ALLMULTI,SIMPLEX,MULTICAST> mtu 1500
+	options=3<RXCSUM,TXCSUM>
+	ether b2:f1:d8:65:66:64
+	inet 192.168.67.1 netmask 0xffffff00 broadcast 192.168.67.255
+	inet6 fe80::b0f1:d8ff:fe65:6664%bridge100 prefixlen 64 scopeid 0x23 
+	inet6 fd91:d62b:63cd:ec3d:4ad:c40c:d0b8:5826 prefixlen 64 autoconf secured 
+	Configuration:
+		id 0:0:0:0:0:0 priority 0 hellotime 0 fwddelay 0
+		maxage 0 holdcnt 0 proto stp maxaddr 100 timeout 1200
+		root id 0:0:0:0:0:0 priority 0 ifcost 0 port 0
+		ipfilter disabled flags 0x0
+	member: vmenet0 flags=3<LEARNING,DISCOVER>
+	        ifmaxaddr 0 port 34 priority 0 path cost 0
+	nd6 options=201<PERFORMNUD,DAD>
+	media: autoselect
+	status: active
+
+Currently, settings like the subnet ranges nor port forwarding are controllable
+via simh. For finer grained control over how systems are addressed, you can use
+a bridged network instead of a host-only or shared network. These settings are
+provided by macOS, so a future version of simh may be able to adopt these.
+
+-------------------------------------------------------------------------------
 
 Windows notes:
  1. The Npcap package available from https://nmap.org/npcap is the preferred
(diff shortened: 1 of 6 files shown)


NattyNarwhal:
The ethernet README mentions it, but it's possible to provide additional knobs; vmnet supports them, it's mostly an issue of figuring out how we want to provide the configuration for it.

markpizz:
If what you've proposed here is, in fact, functionally equivalent to existing ethernet functionality (i.e. pcap, tun/tap, nat), then the simh user commands used to configure these things should not be any different than the current configuration steps.
If along with what is mentioned in my first sentence additional capabilities are also available (i.e. bridging) then extended commands should be added for this functionality,

NattyNarwhal:
> If what you've proposed here is, in fact, functionally equivalent to existing ethernet functionality (i.e. pcap, tun/tap, nat), then the simh user commands used to configure these things should not be any different than the current configuration steps.

Correct, like the other alternate ethernet backends, it's just prefixed (i.e. `vmnet:host` or `vmnet:en0`). It is a little different in that bridging and NAT are presented under the same interface (i.e. it can be used as a substitute for slirp or pcap).

> If along with what is mentioned in my first sentence additional capabilities are also available (i.e. bridging) then extended commands should be added for this functionality,

Right, the vmnet API provides some additional configuration options, but I'm wondering how best to present the options (i.e. a larger connection string, or additional `SET` options?).

----

One thing I also forgot to mention is the MAC address behaviour. When creating a vmnet interface, it gets a MAC address assigned, and this can be overridden (`vmnet_interface_id_key` key on the interface description). It doesn't seem you need to have the emulated system use the vmnet interface's MAC address though. I've been using simh with the assigned MAC addresses for VAXen and it seems fine, so I'm not sure if overriding the automatically assigned vmnet MAC address is useful. It does make `SHOW ETHERNET` mildly more useful (since you can xref vmnet interfaces in i.e. ifconfig w/ it).

For the purposes of vmnet in simh, this does mean changing the MAC address handling as to set the MAC address when the interface is created (rather than look it up later).

pkoning2:
Looking at the "vmnet" web documentation you cited I wonder how it applies to SIMH.  It's described as an API for hypervisors to construct virtual interfaces, which have their own IP addresses and other stuff.  How does that fit what SIMH needs, which is an ethernet level send/receive service?
I agree that tun/tap has long been problematic which is why I stopped looking at it years ago.  But pcap works just fine at least on Mac and Linux.

NattyNarwhal:
> Looking at the "vmnet" web documentation you cited I wonder how it applies to SIMH. It's described as an API for hypervisors to construct virtual interfaces, which have their own IP addresses and other stuff. How does that fit what SIMH needs, which is an ethernet level send/receive service?

vmnet provides a virtual ethernet interface, SIMH is sending/receiving L2 frames over it. While the API documentation mentions hypervisors, SIMH needs the exact same thing as them. The L3 IP stuff vmnet does is for the purposes of subnet it uses for shared/host-only networking, where it handles the DHCP pool and NAT.

markpizz:
From my experience, this didn't work on all macOS versions that are in use by users.  Specifically, 10.15.7.

Additionally, the usage requires users to know something different than before.  When I raised this question, his answer was:

> Correct, like the other alternate ethernet backends, it's just prefixed (i.e. vmnet:host or vmnet:en0). It is a little different in that bridging and NAT are presented under the same interface (i.e. it can be used as a substitute for slirp or pcap).

That answer is wrong.  We don't have users saying slirp:xxxxx  They say nat:
Likewise, we don't have users saying: pcap:en0, they say eth0 or en0.

Additionally, the vmnet: NAT functionality is not useful for any of the most common use cases.  This is due to it depending on client systems getting IP addresses via DHCP, while almost no operating systems running on simh simulators have DHCP support built in to the operating systems.  It ALSO DOES NOT provide any way to actually create open inbound ports which are very commonly used by simh users.

pkoning2:
I tried vmnet:en5 (which is a wired Ethernet dongle) and it worked fine.  Yes, pcap (the default) works also.  Personally I tend to choose pcap, but the existing SIMH has long offered a choice of several different mechanisms.  And yes, tun/tap on Mac is either close to or entirely dead.  My interpretation of the PR is that it offers a replacement for tun/tap that works, i.e., an alternate for people who for any reason do not want to use pcap.  I see nothing wrong with that.

On the NAT thing, I assume that the only mode that's actually useful is the bridged mode, because we're emulating an Ethernet NIC, not an IP device.

pmetzger:
Doesn't getting rid of this get rid of the easiest way for a modern Mac user to get networking support? It's very Mac specific. (I see no reason to care about the fact that it won't work on 10 or 15 year old versions of MacOS.)

pkoning2:
It's been a while, but as far as I know installing pcap on a current Mac is a trivial task.

markpizz:
@pmetzger  said:
> Doesn't getting rid of this get rid of the easiest way for a modern Mac user to get networking support? It's very Mac specific. (I see no reason to care about the fact that it won't work on 10 or 15 year old versions of MacOS.)

I've got vmnet working on a mac that's from Late 2012 which is running the latest OS that hardware supports (10.15.7).

@pkoning2  said:
> It's been a while, but as far as I know installing pcap on a current Mac is a trivial task.

Trivial or otherwise, vmnet does a better job of providing network support for a simulator.  Specifically, when using a vmnet bridged connection, the host can talk directly to the simulated system and the simulated system can talk directly to the LAN and beyond.  Clearly the simulated system would need to use an IP address appropriate to the LAN.

My only suggestion is that the name of the device that vmnet uses should merely be the interface name since it functionally replaces pcap which never accepted pcap:en0 but always accepted the raw interface name.  Anyone with a configuration that worked previously will still have something that works.  With that in place, the reason most folks previously had to attempt tun connections really no longer exist.

pmetzger:
Apple has a policy at this point of removing mechanisms that they perceive to be security risks. Third party kernel device drivers are now dead. Tun and tap are dead. pcap is difficult to use unless the user knows enough to change the ownership of various files in /dev and I am unsure whether it will be provided in the future except as a way of listening to packets. The official way on a modern Mac to provide a virtual machine with access to the network is vmnet. There is no argument against allowing a Mac user to have access to that. It doesn't matter that other platforms can't use this mechanism; other platforms can't use the Windows support code but we still have Windows support code.

pkoning2:
@NattyNarwhal sorry for the delay in sorting this out.

I consulted with the other steering group members.  The answer that came back is that I was too critical.  What we want to see is:
The default build (no magic parameters) works on all supported OS and versions, using the standard compiler etc. for that system.  
So what that means here is that things should build on non-Mac systems (no vmnet of course) and on Mac systems both those too old to support vmnet and the newer ones that do support it.  And on Mac that would be with the standard compiler clang.
Given that gcc is no longer standard on Mac OS, build failures with gcc are allowed (though of course avoiding them would be nice, but that is optional).  We do want vmnet support to work cleanly on old Mac OS that don't have vmnet.  All this applies to both build procedures: the makefile and cmake.

I would much prefer vmnet to be a choice for network access just like tap and others are, where you specify a prefix on the device to get that choice.  pcap for historic reasons doesn't have a prefix, but the others do.  So vmnet doesn't replace pcap, it is an additional choice.  The argument is that having to change all the init files when updating Open SIMH is extra work.  Sometimes it is unavoidable but in this case it need not be as far as I can tell.

If all this is agreeable, please submit a new PR with these items addressed.  And again, sorry for the detours.

pkoning2:
BTW, on pcap and devices, I have not seen that.  I have a shiny new Mac OS and it supports pcap without any magic on my part.  I don't remember where it came from; might be the Wireshark kit, or maybe from "homebrew", but in either case, I just did the install and it worked.

quagly:
Does vmnet support decnet?

pmetzger:
For whomever cares: I have restored this functionality on my hardfork of SIMH, which I'm calling Zeerust. It should be up on Github in the next couple of days. I'm sorry, I needed this stuff and I couldn't wait for any longer.

pmetzger:
It looks like the functionality is actually a bit problematic because it requires running as root, so I've backed it out of my own fork temporarily. However, vmnet-helper or vmnet_broker would provide comparable functionality without that requirement. I'll be constructing an implementation that uses those.

gtackett:
> For whomever cares: I have restored this functionality on my hardfork of SIMH, which I'm calling Zeerust. It should be up on Github in the next couple of days. I'm sorry, I needed this stuff and I couldn't wait for any longer.

Is Zeerust just zimh?
Are you *calling* it Zeerust although its *name* is zimh? 😁
Shades of “Through the Looking-Glass”!

PR #452, opened 2025-03-04
ETH_MAC indirection erratum

bscottm:
Pervasive misuse of "ETH_MAC *" (a pointer to an ETH_MAC, aka a 6-element unsigned char array) when a simple "ETH_MAC" is correct.  The best example of this was eth_mac_fmt() in sim_ether.c with the following prototype:

    t_stat eth_mac_fmt (ETH_MAC* const mac, char* strmac)

The first parameter is a pointer to an array of 6 unsigned characters, whereas it really just wants to be a pointer to the first element of the array:

    t_stat eth_mac_scan (const ETH_MAC mac, char* strmac)

The "ETH_MAC *" indirection error also results in subtle memcpy() and memcmp() issues, e.g.:

    void network_func(DEVICE *dev, ETH_MAC *mac)
    {
      ETH_MAC other_mac;

      /* ...code... */

      /* memcpy() bug: */
      memcpy(other_mac, mac, sizeof(ETH_MAC));

      /* or worse: */
      memcpy(mac, other_mac, sizeof(ETH_MAC));
    }

eth_copy_mac() and eth_mac_cmp() replace calls to memcpy() and memcmp() that copy or compare Ethernet MAC addresses. These are type-enforcing functions, i.e., the parameters are ETH_MAC-s, to avoid the subtle memcpy() and memcmp() bugs.

This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)

### DIFF
--- a/3B2/3b2_ni.c
+++ b/3B2/3b2_ni.c
@@ -557,7 +557,7 @@ t_stat ni_setmac(UNIT *uptr, int32 val, CONST char* cptr, void* desc)
     UNUSED(desc);
 
     status = SCPE_OK;
-    status = eth_mac_scan_ex(&ni.macs[NI_NIC_MAC], cptr, uptr);
+    status = eth_mac_scan_ex(ni.macs[NI_NIC_MAC], cptr, uptr);
 
     if (status == SCPE_OK) {
         eth_filter(ni.eth, ni.filter_count, ni.macs, 0, 0);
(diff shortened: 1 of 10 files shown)


bscottm:
@pkoning2 @rcornwell:

- Really need #450 so that the check build succeeds. I may have to resubmit this PR once that happens. The LTO bug is a real pain.
- This problem has been around for a long time and likely the source of Heisenbugs. It solves one that's been plaguing me in the network code for a couple of months.

PR #453, opened 2025-03-05
ETH_MAC indirection erratum

bscottm:
Pervasive misuse of "ETH_MAC *" (a pointer to an ETH_MAC, aka a 6 element unsigned char array) when a simple "ETH_MAC" is correct.  The best example of this was eth_mac_fmt() in sim_ether.c with the following prototype:

    t_stat eth_mac_fmt (ETH_MAC* const mac, char* strmac)

The first parameter is a pointer to an array of 6 unsigned characters, whereas it really just wants to be a pointer to the first element of the array:

    t_stat eth_mac_scan (const ETH_MAC mac, char* strmac)

The "ETH_MAC *" indirection error also results in subtle memcpy() and memcmp() issues, e.g.:

    void network_func(DEVICE *dev, ETH_MAC *mac)
    {
      ETH_MAC other_mac;

      /* ...code... */

      /* memcpy() bug: */
      memcpy(other_mac, mac, sizeof(ETH_MAC));

      /* or worse: */
      memcpy(mac, other_mac, sizeof(ETH_MAC));
    }

eth_copy_mac() and eth_mac_cmp() replace calls to memcpy() and memcmp() that copy or compare Ethernet MAC addresses. These are type-enforcing functions, i.e., the parameters are ETH_MAC-s, to avoid the subtle memcpy() and memcmp() bugs.

This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)

### DIFF
--- a/sim_ether.c
+++ b/sim_ether.c
@@ -386,25 +386,25 @@ static int _eth_get_system_id (char *buf, size_t buf_size);
 static void eth_get_nic_hw_addr(ETH_DEV* dev, const char *devname, int set_on);
 
 static const unsigned char framer_oui[3] = { 0xaa, 0x00, 0x03 };
+       const ETH_MAC eth_mac_any         = {0, 0, 0, 0, 0, 0};
+       const ETH_MAC eth_mac_bcast       = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
 
 /*============================================================================*/
 /*                  OS-independent ethernet routines                          */
 /*============================================================================*/
 
-t_stat eth_mac_scan (ETH_MAC* mac, const char* strmac)
+t_stat eth_mac_scan (ETH_MAC mac, const char* strmac)
 {
 return eth_mac_scan_ex (mac, strmac, NULL);
 }
 
-t_stat eth_mac_scan_ex (ETH_MAC* mac, const char* strmac, UNIT *uptr)
+t_stat eth_mac_scan_ex (ETH_MAC mac, const char* strmac, UNIT *uptr)
 {
   unsigned int a[6], g[6];
   FILE *f;
   char filebuf[64] = "";
   uint32 i;
-  static const ETH_MAC zeros = {0,0,0,0,0,0};
-  static const ETH_MAC ones  = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
-  ETH_MAC newmac;
+  ETH_MAC newmac = {0, 0, 0, 0, 0, 0};
   struct {
       uint32 bits;
       char system_id[64];
@@ -470,8 +470,8 @@ t_stat eth_mac_scan_ex (ETH_MAC* mac, const char* strmac, UNIT *uptr)
       }
 
   /* final check - mac cannot be broadcast or multicast address */
-  if (!memcmp(newmac, zeros, sizeof(ETH_MAC)) ||  /* broadcast */
-      !memcmp(newmac, ones,  sizeof(ETH_MAC)) ||  /* broadcast */
+  if (!eth_mac_cmp(newmac, eth_mac_any) ||        /* broadcast */
+      !eth_mac_cmp(newmac, eth_mac_bcast) ||      /* broadcast */
       (newmac[0] & 0x01)                          /* multicast */
      )
     return sim_messagef (SCPE_ARG, "Can't use Broadcast or MultiCast address as interface MAC address\n");
@@ -482,29 +482,28 @@ t_stat eth_mac_scan_ex (ETH_MAC* mac, const char* strmac, UNIT *uptr)
     f = fopen (state.file, "w");
     if (f == NULL)
       return sim_messagef (SCPE_ARG, "Can't open MAC address configuration file '%s'.\n", state.file);
-    eth_mac_fmt (&newmac, filebuf);
+    eth_mac_fmt (newmac, filebuf);
     fprintf (f, "%s/48\n", filebuf);
     fprintf (f, "system-id: %s\n", state.system_id);
     fprintf (f, "directory: %s\n", state.cwd);
     fprintf (f, "simulator: %s\n", state.sim);
     fprintf (f, "device:    %s\n", state.uname);
     fprintf (f, "file:      %s\n", state.file);
-    eth_mac_fmt (&state.base_mac, filebuf);
+    eth_mac_fmt (state.base_mac, filebuf);
     fprintf (f, "base-mac:  %s\n", filebuf);
     fprintf (f, "specified: %d bits\n", state.bits);
     fprintf (f, "generated: %d bits\n", 48-state.bits);
     fclose (f);
     }
   /* copy into passed mac */
-  memcpy (*mac, newmac, sizeof(ETH_MAC));
+  eth_copy_mac (mac, newmac);
   return SCPE_OK;
 }
 
-void eth_mac_fmt(ETH_MAC* const mac, char* buff)
+void eth_mac_fmt(const ETH_MAC mac, char* buff)
 {
-  const uint8* m = (const uint8*) mac;
+  const uint8* m = (const uint8 *) mac;
   sprintf(buff, "%02X:%02X:%02X:%02X:%02X:%02X", m[0], m[1], m[2], m[3], m[4], m[5]);
-  return;
 }
 
 static const uint32 crcTable[256] = {
@@ -615,8 +614,8 @@ void eth_packet_trace_ex(ETH_DEV* dev, const uint8 *msg, int len, const char* tx
     char dst[20];
     const unsigned short* proto = (const unsigned short*) &msg[12];
     uint32 crc = eth_crc32(0, msg, len);
-    eth_mac_fmt((ETH_MAC*)msg, dst);
-    eth_mac_fmt((ETH_MAC*)(msg+6), src);
+    eth_mac_fmt(msg, dst);
+    eth_mac_fmt(msg + 6, src);
     sim_debug(reason, dev->dptr, "%s  dst: %s  src: %s  proto: 0x%04X  len: %d  crc: %X\n",
           txt, dst, src, ntohs(*proto), len, crc);
     if (detail) {
@@ -625,7 +624,7 @@ void eth_packet_trace_ex(ETH_DEV* dev, const uint8 *msg, int len, const char* tx
       static const char hex[] = "0123456789ABCDEF";
 
       for (i=same=0; i<len; i += 16) {
-        if ((i > 0) && (0 == memcmp(&msg[i], &msg[i-16], 16))) {
+        if ((i > 0) && (0 == eth_mac_cmp(&msg[i], &msg[i-16]))) {
           ++same;
           continue;
         }
@@ -779,6 +778,8 @@ ethq_insert_data(que, type, pack->oversize ? pack->oversize : pack->msg, pack->u
 
 t_stat eth_show_devices (FILE* st, DEVICE *dptr, UNIT* uptr, int32 val, CONST char *desc)
 {
+(void) dptr;
+(void) desc;
 return eth_show (st, uptr, val, NULL);
 }
 
@@ -873,8 +874,11 @@ static char *(*p_pcap_lib_version) (void);
 
 static void _eth_add_to_open_list (ETH_DEV* dev)
 {
-eth_open_devices = (ETH_DEV**)realloc(eth_open_devices, (eth_open_device_count+1)*sizeof(*eth_open_devices));
-eth_open_devices[eth_open_device_count++] = dev;
+ETH_DEV **tmp = (ETH_DEV**)realloc(eth_open_devices, (eth_open_device_count+1)*sizeof(*eth_open_devices));
+if (tmp != NULL) {
+  eth_open_devices = tmp;
+  eth_open_devices[eth_open_device_count++] = dev;
+  }
 }
 
 static void _eth_remove_from_open_list (ETH_DEV* dev)
@@ -895,6 +899,11 @@ t_stat eth_show (FILE* st, UNIT* uptr, int32 val, CONST void* desc)
   ETH_LIST  list[ETH_MAX_DEVICE];
   int number;
 
+  /* Squelch unused parameters. But do we really need them? */
+  (void) uptr;
+  (void) val;
+  (void) desc;
+
   number = eth_devices(ETH_MAX_DEVICE, list, FALSE);
   fprintf(st, "ETH devices:\n");
   if (number == -1)
@@ -912,11 +921,11 @@ t_stat eth_show (FILE* st, UNIT* uptr, int32 val, CONST void* desc)
     }
   if (eth_open_device_count) {
     int i;
-    char desc[ETH_DEV_DESC_MAX], *d;
+    char devdesc[ETH_DEV_DESC_MAX], *d;
 
     fprintf(st,"Open ETH Devices:\n");
     for (i=0; i<eth_open_device_count; i++) {
-      d = eth_getdesc_byname(eth_open_devices[i]->name, desc);
+      d = eth_getdesc_byname(eth_open_devices[i]->name, devdesc);
       if (d)
         fprintf(st, " %-7s%s (%s)\n", eth_open_devices[i]->dptr->name, eth_open_devices[i]->dptr->units[0].filename, d);
       else
@@ -945,8 +954,7 @@ t_stat eth_attach_help(FILE *st, DEVICE *dptr, UNIT *uptr, int32 flag, const cha
   fprintf (st, "This simulator was not built with ethernet device support\n");
   return SCPE_OK;
   }
-t_stat eth_check_address_conflict (ETH_DEV* dev,
-                                   ETH_MAC* const mac)
+t_stat eth_check_address_conflict (ETH_DEV* dev, const ETH_MAC mac)
   {return SCPE_NOFNC;}
 t_stat eth_set_throttle (ETH_DEV* dev, uint32 time, uint32 burst, uint32 delay)
   {return SCPE_NOFNC;}
@@ -958,13 +966,13 @@ t_stat eth_write (ETH_DEV* dev, ETH_PACK* packet, ETH_PCALLBACK routine)
   {return SCPE_NOFNC;}
 int eth_read (ETH_DEV* dev, ETH_PACK* packet, ETH_PCALLBACK routine)
   {return SCPE_NOFNC;}
-t_stat eth_filter (ETH_DEV* dev, int addr_count, ETH_MAC* const addresses,
+t_stat eth_filter (ETH_DEV* dev, int addr_count, const ETH_MAC addresses[],
                    ETH_BOOL all_multicast, ETH_BOOL promiscuous)
   {return SCPE_NOFNC;}
-t_stat eth_filter_hash (ETH_DEV* dev, int addr_count, ETH_MAC* const addresses,
+t_stat eth_filter_hash (ETH_DEV* dev, int addr_count, const ETH_MAC addresses[],
                    ETH_BOOL all_multicast, ETH_BOOL promiscuous, ETH_MULTIHASH* const hash)
   {return SCPE_NOFNC;}
-t_stat eth_filter_hash_ex(ETH_DEV* dev, int addr_count, ETH_MAC* const addresses,
+t_stat eth_filter_hash_ex(ETH_DEV* dev, int addr_count, const ETH_MAC addresses[],
                           ETH_BOOL all_multicast, ETH_BOOL promiscuous,
                           ETH_BOOL match_broadcast, ETH_MULTIHASH* const hash)
   {return SCPE_NOFNC;}
@@ -2762,7 +2770,7 @@ return (rand() & 0xFF);
 }
 
 t_stat eth_check_address_conflict_ex (ETH_DEV* dev,
-                                      ETH_MAC* const mac,
+                                      const ETH_MAC mac,
                                       int *reflections,
                                       t_bool silent)
 {
@@ -2779,9 +2787,9 @@ eth_mac_fmt(mac, mac_string);
 sim_debug(dev->dbit, dev->dptr, "Determining Address Conflict for MAC address: %s\n", mac_string);
 
 /* 00:00:00:00:00:00 or any address with a multi-cast address is invalid */
-if ((((*mac)[0] == 0) && ((*mac)[1] == 0) && ((*mac)[2] == 0) &&
-     ((*mac)[3] == 0) && ((*mac)[4] == 0) && ((*mac)[5] == 0)) ||
-     ((*mac)[0] & 1)) {
+if (((mac[0] == 0) && (mac[1] == 0) && (mac[2] == 0) &&
+     (mac[3] == 0) && (mac[4] == 0) && (mac[5] == 0)) ||
+     (mac[0] & 1)) {
   return sim_messagef (SCPE_ARG, "%s: Invalid NIC MAC Address: %s\n", sim_dname(dev->dptr), mac_string);
   }
 
@@ -2850,15 +2858,15 @@ memset (&send, 0, sizeof(ETH_PACK));
 send.len = ETH_MIN_PACKET;                              /* minimum packet size */
 for (i=0; i<send.len; i++)
   send.msg[i] = _eth_rand_byte();
-memcpy(&send.msg[0], mac, sizeof(ETH_MAC));             /* target address */
-memcpy(&send.msg[6], mac, sizeof(ETH_MAC));             /* source address */
+eth_copy_mac(&send.msg[0], mac);                        /* target address */
+eth_copy_mac(&send.msg[6], mac);                        /* source address */
 send.msg[12] = 0x90;                                    /* loopback packet type */
 send.msg[13] = 0;
 send.msg[14] = 0;                                       /* Offset */
 send.msg[15] = 0;
 send.msg[16] = 2;                                       /* Forward */
 send.msg[17] = 0;
-memcpy(&send.msg[18], mac, sizeof(ETH_MAC));            /* Forward Destination */
+eth_copy_mac(&send.msg[18], mac);                       /* Forward Destination */
 send.msg[24] = 1;                                       /* Reply */
 send.msg[25] = 0;
 
@@ -2905,13 +2913,12 @@ if (reflections)
 return SCPE_OK;
 }
 
-t_stat eth_check_address_conflict (ETH_DEV* dev,
-                                   ETH_MAC* const mac)
+t_stat eth_check_address_conflict (ETH_DEV* dev, const ETH_MAC mac)
 {
 char mac_string[32];
 
 eth_mac_fmt(mac, mac_string);
-if (0 == memcmp (mac, dev->host_nic_phy_hw_addr, sizeof *mac))
+if (0 == eth_mac_cmp (mac, dev->host_nic_phy_hw_addr))
     return sim_messagef (SCPE_OK, "Sharing the host NIC MAC address %s may cause unexpected behavior\n", mac_string);
 return eth_check_address_conflict_ex (dev, mac, NULL, FALSE);
 }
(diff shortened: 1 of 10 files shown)


bscottm:
@pkoning2:

- Really need #450 so that the check build succeeds. I may have to resubmit this PR once that happens. The LTO bug is a real pain.
- This problem has been around for a long time and likely the source of Heisenbugs. It solves one that's been plaguing me in the network code for a couple of months.

pkoning2:
I wonder about the correct vs. incorrect thing.  The problem is that C has a broken type system, and its way of treating an array as essentially synonymous with a pointer is what lies at the bottom of this.

bscottm:
> > I wonder about the correct vs. incorrect thing. The problem is that C has a broken type system, and its way of treating an array as essentially synonymous with a pointer is what lies at the bottom of this.
> 
> Quite true. However, passing around a pointer to an array vs. the pointer to the array's first element is the issue here. At least with a type-enforcing inline function, code is more "correct" (since the parameters are pointers to the array's first element.)

You'd have been surprised how many places code was `memcpy`-ing using the first 4 or 8 bytes of the ETH_MAC array as the source pointer (the effect of "ETH_MAC *"). Or awkward `(ETH_MAC *)` casts to `eth_mac_fmt()` (that should have signaled an obvious error to someone reviewing the code.)

bscottm:
> > I wonder about the correct vs. incorrect thing. The problem is that C has a broken type system, and its way of treating an array as essentially synonymous with a pointer is what lies at the bottom of this.
> 
> Quite true. However, passing around a pointer to an array vs. the pointer to the array's first element is the issue here. At least with a type-enforcing inline function, code is more "correct" (since the parameters are pointers to the array's first element.)

To put it another way, `ETH_MAC *` is `unsigned char **` when passed as a parameter. `ETH_MAC` is effectively `unsigned char *`. I'd argue that `ETH_MAC *` is incorrect and not what was originally intended.

markpizz:
So lets see.  You need to change 100's of lines in 10 files to change the view of a ETH_MAC which is a 6 byte object into an array of 6 bytes.

You claim that there were places that "you'd have been surprised how many places code was memcpy-ing using the first 4 or 8 bytes of ".  Given the 100's of lines of code affected in your proposed "fix", you don't specifically identify any  of these 4 or 8 byte error cases.

Looking through ALL of your changes, I can't find any memcpy (or memcmp) that you "fixed" with your massive changes that actually did what you claim (4 or 8 byte).

markpizz:
You claim: "This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)", but there isn't actually a function named _eth_close().  There is a function named _eth_close_port() which makes no reference to ETH_MAC objects.  Similarly there is a function named eth_close() which also makes no reference to any ETH_MAC objects.

bscottm:
> Looking through ALL of your changes, I can't find any memcpy (or memcmp) that you "fixed" with your massive changes that actually did what you claim (4 or 8 byte).

`ETH_MAC *` is equivalent to `unsigned char **`, so if you have code that looks like the example I wrote above, there's a subtle bug when `ETH_MAC` and `ETH_MAC *` are both coerced to `void *` in a `memcpy`. Might not be obvious to you at first glance, but really obvious when replaced with a type enforcing inline.

Just to make it really plain, I'll further annotate the example:

```
void network_func(DEVICE *dev, ETH_MAC *mac)
{
  ETH_MAC other_mac;

  /* ...code... */

  /* memcpy() bug: other_mac's type is equivalent to 'unsigned char *' (pointer to the first
   * array element), mac's type is* equivalent to 'unsigned char **' (pointer to a pointer to
   * the array's first element), both of which are coerced to 'void *': */
  memcpy(other_mac, mac, sizeof(ETH_MAC));

  /* or worse: */
  memcpy(mac, other_mac, sizeof(ETH_MAC));
}
```

bscottm:
The reason for replacing all of the `memcpy` and `memcmp` calls involving an `ETH_MAC` is consistency -- instead of examining every single `memcpy` or `memcmp` to catch the `&mac2` case below, it's just easier to replace every MAC address `memcpy` and `memcmp` with `eth_copy_mac` and `eth_mac_cmp` to catch all cases.

Example program where the extra indirection is an error (i.e., MAC address comes from a `uint8_t` array, aka a packet):

```
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>

typedef uint8_t ETH_MAC[6];

void foo(const char *label, ETH_MAC *mac)
{
    printf("%-12s: mac = %0.16p [*mac = %0.16p, **mac = %02x]\n", label, mac, *mac, **mac);
}

int main(void)
{
    ETH_MAC mac = { 0x01, 0x23, 0x45, 0x67, 0x89, 0xab };

    puts("mac and &mac arrays:");
    foo("mac", (ETH_MAC *) mac);
    foo("&mac", &mac);

    uint8_t *mac2 = malloc(64 * sizeof(uint8_t));
    for (size_t i = 0; i < 63; ++i)
      mac2[i] = (uint8_t) (i + 127);

    puts("\nuint8_t masquerading as ETH_MAC *:");
    foo("mac2", (ETH_MAC *) mac2);
    foo("&mac2", (ETH_MAC *) &mac2);

    return 0;
}
```

Output:

```
mac and &mac arrays:
mac         : mac = 0x0000007fd02bc5a0 [*mac = 0x0000007fd02bc5a0, **mac = 01]
&mac        : mac = 0x0000007fd02bc5a0 [*mac = 0x0000007fd02bc5a0, **mac = 01]

uint8_t masquerading as ETH_MAC *:
mac2        : mac = 0x00000055afebb6b0 [*mac = 0x00000055afebb6b0, **mac = 7f]
&mac2       : mac = 0x0000007fd02bc598 [*mac = 0x0000007fd02bc598, **mac = b0]
```

bscottm:
Remove the extra `ETH_MAC` indirection, and you get:

```
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>

typedef uint8_t ETH_MAC[6];

void foo(const char *label, ETH_MAC mac)
{
    printf("%-12s: mac = %0.16p [*mac = %02x]\n", label, mac, *mac);
}

int main(void)
{
    ETH_MAC mac = { 0x01, 0x23, 0x45, 0x67, 0x89, 0xab };

    puts("mac array:");
    foo("mac", mac);

    uint8_t *mac2 = malloc(64 * sizeof(uint8_t));
    for (size_t i = 0; i < 63; ++i)
      mac2[i] = (uint8_t) (i + 127);

    puts("\nuint8_t masquerading as ETH_MAC:");
    foo("mac2", mac2);

    return 0;
}
```

Output:
```
mac array:
mac         : mac = 0x0000007ff2cfa7d8 [*mac = 01]

uint8_t masquerading as ETH_MAC:
mac2        : mac = 0x000000558320f6b0 [*mac = 7f]
```

You get a consistent result for when the `ETH_MAC` is actually an underlying packet.

bscottm:
> You claim: "This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)", but there isn't actually a function named _eth_close(). There is a function named _eth_close_port() which makes no reference to ETH_MAC objects. Similarly there is a function named eth_close() which also makes no reference to any ETH_MAC objects.

Oh. Gee. I had a typo. BFD.

bscottm:
> You claim: "This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)", but there isn't actually a function named _eth_close(). There is a function named _eth_close_port() which makes no reference to ETH_MAC objects. Similarly there is a function named eth_close() which also makes no reference to any ETH_MAC objects.

@markpizz There is an `eth_close()` function -- ok, so I typo-ed the name with an extra underscore. The Heisenbug pops up at line 2686 while `free()`-ing the `write_requests` list (MS's heap validator detects corruption on the `pdp10-kl` simulator, primarily.) Going through all of the `ETH_MAC` `memcpy()`-s to track that down would have been very time consuming and error prone.

Much easier and more reliable to replace the `memcpy()`-s with a type enforcing function to simply "get it right."

markpizz:
> In other words, ETH_MAC * and ETH_MAC are distinct and different types.

I don't dispute this statement.  However, there may (or may not) be one or at most a couple of places where this is mixed up.  Fixing those few places would be an appropriate fix rather than the 100's of lines of code changes you've proposed.  You carefully avoid pointing out any specific case (or cases).

> > You claim: "This fix solves at least one Heisenbug in _eth_close() while free()-ing write request buffers (and possibly other Heisenbugs.)", but there isn't actually a function named _eth_close(). There is a function named _eth_close_port() which makes no reference to ETH_MAC objects. Similarly there is a function named eth_close() which also makes no reference to any ETH_MAC objects.
> >
> Oh. Gee. I had a typo. BFD.

Surely you can have a typo.

> @markpizz There is an eth_close() function -- ok, so I typo-ed the name with an extra underscore. The Heisenbug pops up at line 2686 while free()-ing the write_requests list (MS's heap validator detects corruption on the pdp10-kl simulator, primarily, since that's what I'm testing and working with at the moment.) Going through all of the ETH_MAC memcpy()-s to track that down would have been very time consuming and error prone.
> 
> Much easier and more reliable to replace the memcpy()-s with a type enforcing function to simply "get it right."

Independent of the typo, you claim that there is a problem there which, in fact you don't actually fix since you didn't actually find ANY case where the issue you claim has happened (wrong pointer to memcpy) is actually at fault.  Your changes no longer demonstrate the memory corruption problem which occurs in the cleanup being done of a linked list WITHOUT ANY ETH_MAC objects being referenced by eth_close().  Your fix changes SO MANY things that you've merely hidden the potential heisenbug problem.

You claim: "Going through all of the ETH_MAC memcpy()-s to track that down would have been very time consuming and error prone."  Actually, there are maybe 3 places in essentially any running simulator which write ETH_MAC objects.  Those should be somewhat easy to find, but there are other ways to track down a memory corruption problem which would be useful to track down the problem.  Since you are already using MS's heap validator, there should be an easy way to force a heap validation check with a specific call in the very small number of places that actually copy ETH_MAC objects.  There could be bugs in other code (outside of sim_ether) that corrupt the heap.  I recently sent a message to the author of the pdp10-kl simulator about such a bug in kx10_imp.c that could possibly write outside of the contents of a packet.

Maybe you want to provide specific details about how you reproduce the failing case.

markpizz:
> > Maybe you want to provide specific details about how you reproduce the failing case.
> 
> And if you're unable to reproduce it? What would that prove? All it would prove is that you cannot reproduce the problem while I still get to suffer with it. I've solved a particular issue for which I've submitted a PR and from which the open-simh could benefit.

Your PR DOES NOT specifically identify an actual problem and a fix for it.

If you were solving the whole network problem from scratch, you could code the various pieces however you want, but the existing WORKING code happens to be written with a view of things that you personally wouldn't have chosen from scratch.  

Your PR does not prove that the problem has been solved, only that it seems to have not been detected in the same way as before.

Interesting that you see this problem with only one simulator but your change requires changes to many simulators (dozens of simulators involving 10 files and hundreds of lines of change).  

Once again, describe the problem's reproduction and I will either demonstrate it and propose a very minimal fix, or prove that it has nothing to do with memcpy of ETH_MAC objects.

pkoning2:
I think you're mistaking C for a strongly typed language.  It isn't.

bscottm:
> I think you're mistaking C for a strongly typed language. It isn't.

Haskell, it ain't for sure. My contention is that it's a good idea to use the type system, such as it is, to catch subtle errors. This is one case where the type system helps, not hurts.

markpizz:
> > Your PR does not prove that the problem has been solved, only that it seems to have not been detected in the same way as before.
> 
> So, you're admitting there's a bug while claiming that I haven't actually found one?

There is some sort of bug which corrupts heap.  The failure on exit proves that, but you haven't actually identified the specific bug.

The failure on exit is NOT the bug.  The bug happened some time before that and the failure on exit is the demonstration that the bug had been encountered during this particular run.

Meanwhile, You've provided good instructions about how to see the failure.  It starts with: "1. Create an [ITS](https://github.com/PDP-10/its) image. I've been using the pdp10-kl simulator."  Since you've already got an ITS image that fails, will you make that available somewhere (Google Drive, FTP, github, etc.) so that hill doesn't need to be climbed to see the failure?

Thanks.

bscottm:
> The failure on exit is NOT the bug. The bug happened some time before that and the failure on exit is the demonstration that the bug had been encountered during this particular run.

<sarcasm>You don't say. I'd have never guessed. You mean that can happen with `malloc()`?</sarcasm>

> Meanwhile, You've provided good instructions about how to see the failure. It starts with: "1. Create an [ITS](https://github.com/PDP-10/its) image. I've been using the pdp10-kl simulator." Since you've already got an ITS image that fails, will you make that available somewhere (Google Drive, FTP, github, etc.) so that hill doesn't need to be climbed to see the failure?

It's easy enough to build it, if @larsbrinkhoff or @rcornwell haven't uploaded it already somewhere. Standard, stock ITS using the generated SIMH configuration file with aforementioned tweaks to the IMP network. If you build it from the PDP-10/its repo on GH, you also get the SIMH configuration file in the `out/<sim>/run` file, e.g., `out/pdp10-kl/run`.

markpizz:
> > Meanwhile, You've provided good instructions about how to see the failure. It starts with: "1. Create an [ITS](https://github.com/PDP-10/its) image. I've been using the pdp10-kl simulator." Since you've already got an ITS image that fails, will you make that available somewhere (Google Drive, FTP, github, etc.) so that hill doesn't need to be climbed to see the failure?

> 
> It's easy enough to build it, if @larsbrinkhoff or @rcornwell haven't uploaded it already somewhere. Standard, stock ITS using the generated SIMH configuration file with aforementioned tweaks to the IMP network. If you build it from the PDP-10/its repo on GH, you also get the SIMH configuration file in the out/<sim>/run file, e.g., out/pdp10-kl/run.

Well, technically it is easy enough to build, BUT a build seems to take some 2 hours on my working environment AND the build doesn't attempt to build the ENVIRONMENT needed pieces until the very end.  These may fail depending on whether all the dependencies are installed.  One which wasn't mentioned in the repo was openssl  for cbridge.  There were a couple of others so it took some 6 hours to get through everything, since I started from scratch each time.  Even if I didn't start from scratch, it still built all the contents again which seemed to be 99% of the elapsed time.  I've got the results now.

bscottm:
@markpizz: While I appreciate your minimalist approach to rectifying places where the Ethernet MAC address is embedded in a packet buffer (an `unsigned char *` allocated space), you will still have residual risk because the solution is incomplete. There can still be `memcpy()` calls where an `unsigned char *` pointer has been cast to `ETH_MAC *` and hasn't been dereferenced (which is what you'll need to do.)

That's one reason why I developed a more comprehensive solution. Ultimately, it makes the code safer because `unsigned char *` and `ETH_MAC` are interchangeable when passed as a parameter. Passing the address of a pointer, which is required to hoist an `unsigned char *` to `ETH_MAC *` becomes a _latent_ bug.

Let me underscore my point with an expanded version of the example code above:

```
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>
#include <string.h>

typedef uint8_t ETH_MAC[6];

void foo(const char *label, ETH_MAC *mac, ETH_MAC *compare)
{
    printf("%-12s: mac = %0.16p *mac = %0.16p\n            **mac = %02x:%02x:%02x:%02x:%02x:%02x\n",
           label, mac, *mac, (*mac)[0], (*mac)[1], (*mac)[2], (*mac)[3], (*mac)[4], (*mac)[5]);
    printf("              cmp = %02x:%02x:%02x:%02x:%02x:%02x\n",
           (*compare)[0], (*compare)[1], (*compare)[2], (*compare)[3], (*compare)[4], (*compare)[5]);
    printf("              memcmp(mac, compare, sizeof(ETH_MAC)) = %d\n",
           memcmp(*mac, *compare, sizeof(ETH_MAC)));
}

int main(void)
{
    ETH_MAC mac = { 0x01, 0x23, 0x45, 0x67, 0x89, 0xab };

    /* mac == &mac: The address of the first element is the
     * same as the address of the array. */
    foo("mac", (ETH_MAC *) mac, &mac);
    /* &mac == &mac */
    foo("&mac", &mac, &mac);

    uint8_t *mac2 = malloc(64 * sizeof(uint8_t));
    for (size_t i = 0; i < 63; ++i)
      mac2[i] = (uint8_t) (i + 127);

    puts("\nuint8_t pointer masquerading as ETH_MAC *:");
    foo("mac2", (ETH_MAC *) mac2, (ETH_MAC *) mac2);
    /* Fails. */
    foo("mac2.1", (ETH_MAC *) mac2, (ETH_MAC *) &mac2);
    /* Fails */
    foo("&mac2", (ETH_MAC *) &mac2, (ETH_MAC *) mac2);

    return 0;
}
```

Here's the output:

```
mac         : mac = 0x0000007fca658cc0 *mac = 0x0000007fca658cc0
            **mac = 01:23:45:67:89:ab
              cmp = 01:23:45:67:89:ab
              memcmp(mac, compare, sizeof(ETH_MAC)) = 0
&mac        : mac = 0x0000007fca658cc0 *mac = 0x0000007fca658cc0
            **mac = 01:23:45:67:89:ab
              cmp = 01:23:45:67:89:ab
              memcmp(mac, compare, sizeof(ETH_MAC)) = 0

uint8_t pointer masquerading as ETH_MAC *:
mac2        : mac = 0x00000055989806b0 *mac = 0x00000055989806b0
            **mac = 7f:80:81:82:83:84
              cmp = 7f:80:81:82:83:84
              memcmp(mac, compare, sizeof(ETH_MAC)) = 0
mac2.1      : mac = 0x00000055989806b0 *mac = 0x00000055989806b0
            **mac = 7f:80:81:82:83:84
              cmp = b0:06:98:98:55:00
              memcmp(mac, compare, sizeof(ETH_MAC)) = -1
&mac2       : mac = 0x0000007fca658cb8 *mac = 0x0000007fca658cb8
            **mac = b0:06:98:98:55:00
              cmp = 7f:80:81:82:83:84
              memcmp(mac, compare, sizeof(ETH_MAC)) = 1
```

In the first two `mac` cases, the address of the array (`&mac`) is the same as the array itself (`mac`), and it also works for the `mac2` case with hoisting the `unsigned char *`/`uint8_t` as an `ETH_MAC *`. The `mac2.1` and `&mac2` cases illustrate the latent bug where the pointer's address is being used as the MAC address.

I'm not as confident as you that this latent bug only exists in three places. Hence the more comprehensive solution that ultimately makes the code safer, where `unsigned char *` can be used safely and interchangeably with `ETH_MAC`.

bscottm:
Let's ask Copilot this question: Assume I have a C typedef, "typedef unsigned char ETH_MAC[6]". Why is it a bug to pass the array as a pointer, i.e., "ETH_MAC *", to functions?

Answer:
In C, using a typedef to define an array type, such as typedef unsigned char ETH_MAC[6];, creates a specific array type with a fixed size. When you pass an array to a function, it decays to a pointer to its first element. Therefore, passing an array as a pointer to a function can lead to potential issues and bugs. Here are some reasons why it is a bug to pass ETH_MAC * to functions:

Loss of Array Size Information: When you pass an array as a pointer, the size information of the array is lost. The function receiving the pointer has no way of knowing the actual size of the array. This can lead to out-of-bounds access and buffer overflows if the function assumes a different array size.

Pointer Arithmetic: When you pass ETH_MAC * to a function, it is treated as a pointer to an array of 6 unsigned char elements. However, pointer arithmetic on this type can be confusing and error-prone, as it may not behave as expected. For example, incrementing the pointer will move it by 6 bytes, not 1 byte.

Unexpected Behavior: If the function modifies the contents of the array through the pointer, it may lead to unexpected behavior if the caller is not aware of the modifications. This can result in hard-to-debug issues, especially if the array is used elsewhere in the program.

Type Mismatch: Passing ETH_MAC * to functions expecting a different pointer type can lead to type mismatches and potential undefined behavior. This can cause compilation errors or runtime issues.

To avoid these issues, it is better to pass the array directly or use a pointer with explicit size information.

markpizz:
As I said before, your hundreds of lines of code changes don't actually identify a problem anywhere, just something you wouldn't have done the way it was originally written.   You provide examples that cover your theory, but don't actually reflect anything that was done in the original code.  

Meanwhile, I can reliably reproduce the failure you observe when closing the IMP device ethernet connection both with the https://github/simh/simh and https://github.com/bscottm/open-simh code in your PR.  I'll provide an appropriate PR to fix this problem after I fully test the minimal change it involves.

bscottm:
> As I said before, your hundreds of lines of code changes don't actually identify a problem anywhere, just something you wouldn't have done the way it was originally written. You provide examples that cover your theory, but don't actually reflect anything that was done in the original code.

On the contrary, I identified a latent bug source that doesn't result in a catastrophic termination of SIMH. Latent bugs don't have to cause catastrophic termination; they are known to cause corruption and incorrect behavior.  Instead of identifying individual instances of `memcpy()` where corruption or incorrect behavior occurs, I took the opportunity to eliminate the latent bug source.

Coincidentally, it eliminated a tiresome heap corruption message each time I exited the PDP10-KL simulator.

This isn't a coding style issue. It's a correctness issue -- `ETH_MAC *` is incorrect if you're going to mix arrays, pointers and pointers to pointers cast to `ETH_MAC *` pointers as happens in the code. You will have to examine each call site to eliminate the residual risk that you will have found all of the `memcpy()` or `memcmp()` instances that are this particular latent bug source.

bscottm:
Ultimately, @pkoning2 and the open-simh steering group decide whether to accept this PR.

markpizz:
> > As I said before, your hundreds of lines of code changes don't actually identify a problem anywhere, just something you wouldn't have done the way it was originally written. You provide examples that cover your theory, but don't actually reflect anything that was done in the original code.
> 
> On the contrary, I identified a latent bug source that doesn't result in a catastrophic termination of SIMH. Latent bugs don't have to cause catastrophic termination; they are known to cause corruption and incorrect behavior. Instead of identifying individual instances of memcpy() where corruption or incorrect behavior occurs, I took the opportunity to eliminate the latent bug source.
> 
> Coincidentally, it eliminated a tiresome heap corruption message each time I exited the PDP10-KL simulator.

Since you didn't identify ANY place where your theoretic bug source ever was happening, your code change merely changed the timing, so that on your limited testing cases, the heap problem (like many timing related issues) might not happen as often.

You didn't change the fact that it can happen.  The test case I'm using fails almost every time after 10 or 15 seconds of simulator execution WHEN RUNNING using your code in this PR.  Note: I'm not using your test cases to observe this.

The fix to this heap corruption problem is a 1 line change to one module.  This fix applied to either the original codebase or the codebase in your PR correctly fixes the heap corruption problem COMPLETELY independent of the timing.

Your deep expertise with this code, along with Copilot's help, should easily identify the problem and the fix should then be obvious.

> This isn't a coding style issue. It's a correctness issue -- ETH_MAC * is incorrect if you're going to mix arrays, pointers and pointers to pointers cast to ETH_MAC * pointers as happens in the code. You will have to examine each call site to eliminate the residual risk that you will have found all of the memcpy() or memcmp() instances that are this particular latent bug source.

Somehow all of the simulators that have used sim_ether in simh have actually managed to correctly avoid the problem spelled out in your your theoretical memcmp/memcpy problem.  These mostly have been around for one or maybe two decades with the last 2 additions going back 5 or 6 years.  If new ethernet network devices were being added today, maybe someone might run into your theoretical issue and thus possibly benefit from your personal view that makes changing hundreds of lines of code in 10 files a good idea.  I haven't heard about anyone adding new simulators with ethernet support.

Although you admit that while changing the hundreds of lines of code in each of the 10 source modules, you didn't check/test each use case of ETH_MAC in memcmp and memcpy, the original authors of EACH of the source modules actually did test each case while they did their development.  Therefore, had your theoretical horrible problems would have caused each of those memcmp/memcpy cases to not work as they intended, so their explicit intents were actually correct.

> Ultimately, @pkoning2 and the open-simh steering group decide whether to accept this PR.

Absolutely!  It is their jobs, not mine!

If accepted for merge, the latent heap corruption bug will remain to be found and fixed.  Good Luck.

bscottm:
> > Ultimately, @pkoning2 and the open-simh steering group decide whether to accept this PR.
> 
> Absolutely! It is their jobs, not mine!
> 
> If accepted for merge, the latent heap corruption bug will remain to be found and fixed. Good Luck.

There's no graceful way to put this: `ETH_MAC *`, while legal, is a software defect that needs to be corrected. CoPilot and I are making the same exact point -- simply use `ETH_MAC` as the parameter and not `ETH_MAC *`. Simplify the code. Sure, 256 lines counts as "hundreds", but a lot of that is removing the extra level of indirection, so "mountains out of molehills."

Since you're stooping to the level of ad hominem, I've been coding professionally since 1987, playing with computers since 1979 with my high school's PDP-8. I have a PhD from UCLA Computer Science (2004), taught and mentored undergraduates in operating systems and software development practices and been involved in the software acquisitions of several space systems that you use or affect your life every day.

I don't disrespect your professional background or coding skills, so I'd really appreciate it if you'd lay off with the ad hominem.

rcornwell:
Sorry I have been out of development loop for a bit while I get my kite company reopened. But the comments on this are absurd. I see nothing wrong with the patch. Fixing potential bugs is a good thing, code cleanup is a good thing. The bickering about a simple code change is uncalled for. I see no problem with the patch. When committed I will eventually merge it into my sims repo. But might be a while.

markpizz:
> I don't disrespect your professional background or coding skills, so I'd really appreciate it if you'd lay off with the ad hominem.

I'm not sure what you're specifically referring to here.  Maybe when I said: "Your deep expertise with this code, along with Copilot's help, should easily identify the problem and the fix should then be obvious.".  When I said this, I was very specifically referring to what I had just said in the previous two paragraphs:

> You didn't change the fact that it can happen. The test case I'm using fails almost every time after 10 or 15 seconds of simulator execution WHEN RUNNING using your code in this PR. Note: I'm not using your test cases to observe this.
> 
> The fix to this heap corruption problem is a 1 line change to one module. This fix applied to either the original codebase or the codebase in your PR correctly fixes the heap corruption problem COMPLETELY independent of the timing.

I narrowed the scope of where you (and CoPilot) should look to solve the heap corruption problem.  

If you're really interested in fixing the heap corruption problem, you might want to focus on that rather than changing things which aren't broken.  If so, I might help point you in the right direction. beyond what has already been said above.

bscottm:
@markpizz: If you're looking for a challenge, after applying this patch, there's at least one place where the debug code is attempting to print data from expired stack frames. You were correct: there are other Heisenbugs. (Yes, I figured it out and crafting another commit to add to this one, but I wouldn't want to deprive you of the pleasure.)

bscottm:
@markpizz: And whilst you're at it, you might also want to clean up the SDL warning. SDL2 is at 2.32.4 and the warning message is still emitted to upgrade to 2.0.14. Alternatively, I could just land my rewrite of the sim_video.c source.

sim_ether.c
```diff
@@ -470,8 +470,8 @@ t_stat eth_mac_scan_ex (ETH_MAC* mac, const char* strmac, UNIT *uptr)
       }
 
   /* final check - mac cannot be broadcast or multicast address */
-  if (!memcmp(newmac, zeros, sizeof(ETH_MAC)) ||  /* broadcast */
-      !memcmp(newmac, ones,  sizeof(ETH_MAC)) ||  /* broadcast */
+  if (!eth_mac_cmp(newmac, eth_mac_any) ||        /* broadcast */
```
pkoning2, line 473: I agree that the all-zeroes address is not a reasonable value to set, but it isn't a broadcast address, obviously.

markpizz:
These changes DON'T fix anything.  He didn't actually find a bug.  The bug he experienced after changing MANY things merely changed the timing enough so that his difficult to reproduce failure had sufficiently different timing that the failure didn't occur with his test case.

The test case I used failed in seconds every time USING HIS CHANGES.  The failing case was fixed with a 1 line change.

pkoning2:
Well, if you know of a bug you're invited to submit a fix.

markpizz:
Accepting this change in the categories of:
1) No actual bug found or fixed
2) "gee if I wrote this from scratch, I'd do it differently"

suggests you would significantly reduce your workload if  you merely let @bscottm commit his own changes.

markpizz:
Although this change actually fixed nothing that was actually broken, the results, which like other proposed changes from this 'contributor' raised the C language  codebase.  
The current code  such that the simulators with Ethernet devices (all the ones affected by this PR) no longer build with the Visual Studio Projects.

PR #456, opened 2025-03-20
Separate debug character from signal

bscottm:
Add the "DBGSIGNAL", "DBGSIG" options to enable sending the interrupt to the user's debugger, which is presumably gdb.  Separates setting the debugging interrupt character from the action so that it is possible to turn debugger interrupts on and off:

     set console dbgint=1f dbgsignal
     set console nodbgsignal

Previously, this feature was only available if the compiler did not define __OPTIMIZE__, which precludes optimized debugging code, i.e., compiling with "-O2 -g".

### DIFF
--- a/sim_console.c
+++ b/sim_console.c
@@ -161,7 +161,9 @@
 /* Forward declarations of platform specific routines */
 
 static t_stat sim_os_poll_kbd (void);
+#if defined(SIM_ASYNCH_IO) && defined(SIM_ASYNCH_MUX)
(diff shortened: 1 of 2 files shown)


markpizz:
Once again: Either the steering committee has changed the project commit message goals, or you forgot to format the commit message identifying the affected component (or simulator). The original commit message rules were specifically recommended by one of the steering committee members.

Otherwise this looks good.

bscottm:
PJFKM.

PR #460, opened 2025-04-05
Avoid name space collision for the global variables PC, SP and BC whe…

### LINKED ISSUE #458: Crash in simulators using variable named "SP" when using ncurses
This issue was found on Ubuntu 22.04 but is likely a problem on all platforms using ncurses. 
It turns out that ncurses library libtinfo exports two symbols unfortunately named "SP" and "PC". Any simulator using non-static variables with those names will clash with libtinfo causing crashes. The only reason those simulators using name "SP" don't crash immediately is that they set to initial SP value to 0, which is the same value libtinfo apparently expects.

This can be reproduced reliably, I confirmed using the NOVA emulator:

NOVA simulator Open SIMH V4.1-0 Current        git commit id: b036821a
sim> reset <PRESS ARROW UP>
sim> reset
sim> dep SP 1234
sim> <PRESS ARROW UP> Segmentation fault (core dumped)

As soon as SP is set to a value other than 0, bad things happen.

I will send a pull request for the NOVA simulator that makes SP a static, fixing the issue.

I have confirmed the same issue exists in the Altair simulator, and likely others as well.

pkooiman:
…n readline is dynamically loaded. 
This fixes #458 for all affected simulators.

### DIFF
--- a/ALTAIR/altair_defs.h
+++ b/ALTAIR/altair_defs.h
@@ -40,3 +40,6 @@
 #define STOP_IBKPT      3                                /* breakpoint */
 #define STOP_OPCODE     4
 
+/* Rename of global BC and SP variables to avoid namespace conflicts on some platforms */
+#define BC BC_Global
+#define SP SP_Global
--- a/Intel-Systems/Intel-MDS/system_defs.h
+++ b/Intel-Systems/Intel-MDS/system_defs.h
@@ -31,6 +31,11 @@
 #include <string.h>
 #include "sim_defs.h"                   /* simulator defns */
 
+/* Rename of global PC, BC and SP variables to avoid namespace conflicts on some platforms */
+#define PC PC_Global
+#define BC BC_Global
+#define SP SP_Global
+
 #define SET_XACK(VAL)   (xack = VAL)
 
 #define I3214_NUM       0
--- a/LGP/lgp_defs.h
+++ b/LGP/lgp_defs.h
@@ -31,6 +31,10 @@
 
 #include "sim_defs.h"                                   /* simulator defns */
 
+/* Rename of global PC variable to avoid namespace conflicts on some platforms */
+#define PC PC_Global
+
+
 #if defined(USE_INT64) || defined(USE_ADDR64)
 #error "LGP-30 does not support 64b values!"
 #endif
--- a/NOVA/nova_defs.h
+++ b/NOVA/nova_defs.h
@@ -51,6 +51,9 @@
 
 #include "sim_defs.h"                                   /* simulator defns */
 
+/* Rename of global SP variable to avoid namespace conflicts on some platforms */
+#define SP SP_Global
+
 #if (defined(USE_INT64) && !defined(ECLIPSE)) || defined(USE_ADDR64)
 #error "Nova does not support 64b values!"
 #endif
--- a/PDP10/kx10_defs.h
+++ b/PDP10/kx10_defs.h
@@ -30,6 +30,9 @@
 
 #include "sim_defs.h"                                   /* simulator defns */
 
+/* Rename of global PC variable to avoid namespace conflicts on some platforms */
+#define PC PC_Global
+ 
 #if defined(USE_ADDR64)
 #error "PDP-10 does not support 64b addresses!"
 #endif
--- a/sigma/sigma_defs.h
+++ b/sigma/sigma_defs.h
@@ -34,6 +34,10 @@
 
 #include "sim_defs.h"                                   /* simulator defns */
 
+/* Rename of global PC variable to avoid namespace conflicts on some platforms */
+#define PC PC_Global
+
+
 #if defined(USE_INT64) || defined(USE_ADDR64)
 #error "Sigma 32b does not support 64b values!"
 #endif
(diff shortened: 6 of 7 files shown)


pkooiman:
Since Mark said the exact opposite yesterday, I will wait for his comments before making any changes.

psco:
My point is that the `static` approach works today but we do not know whether a future change requires access to one of these variables. If this should happen, the renaming via a macro is needed anyway. I do not see a downside to the macro approach.

markpizz:
> My point is that the static approach works today but we do not know whether a future change requires access to one of these variables. If this should happen, the renaming via a macro is needed anyway. I do not see a downside to the macro approach.

That is true, but in each of the "static:" cases, there is one source module which completely implements the processor so any theoretical future addition of devices would not be messing with the processor's internal registers.  However, there really isn't a downside to the macro approach.  The simname_defs.h is where definitions and globals shared between different source modules belongs.

FYI, you can "git push -f" to change an existing pull request if changes are needed after review.

rcornwell:
Only problem I can see with global renaming of something like PC is it might confuse people trying to debug a simulator as it would be called PC_Global instead of PC like in the source.

markpizz:
> Only problem I can see with global renaming of something like PC is it might confuse people trying to debug a simulator as it would be called PC_Global instead of PC like in the source.

True, but it depends on the level of debugging they are doing.  The name PC_Global is what the OS debugger sees, but the simh REG variable is still whatever the simulator author named it and is visible from the sim> prompt by that name.  

These changes are only for much older simulators which OS level debugging is quite unlikely.  New simulators being developed now will naturally have locally scoped variable names (static) rather than the default of global the author(s) of the older simulators had.

The original changes that addressed the PC_Global case (c2cef3c) have been in the codebase for more than 10 years.

pkooiman:
If someone in the future needs to do major work on one of the affected simulators and is annoyed enough at having to use  "PC_Global" instead of PC in the debugger, I suppose they have the option of manually renaming PC to something else. This seems to have happened with the hp2100 simulator according to [hp2100_bugfixes.txt](https://github.com/open-simh/simh/blob/b036821a68a91bf6a797e8cd94a6293c3997aa9d/HP2100/hp2100_bugfixes.txt#L7109C43-L7111C68). 

I am happy to amend the pull request to use the #define everywhere, let me know if that is the preferred option.

psco:
I would prefer the renaming option via #define.

markpizz:
> I would prefer the renaming option via #define.

Agreed.

psco:
Thank you - this looks good. I checked and the issue was present even in Ubuntu 24.10 (in my previous attempt I had not installed libedit-dev and without editline functionality there is no crash).

pkoning2:
Libraries that use names like this clearly have a design issue; it's annoying to have to work around errors like that.

rcornwell:
I do not believe that PDP10 simulators include ncurses. I have deposited values into PC without issue. Perhaps only those simulators the use ncurses be fixed.

markpizz:
You could easily correctly argue that the problem is within ncurses or readline or whatever logic has the shared global variables, BUT getting that changed everywhere would be a much bigger hill to climb than the proposed fix to the simh simulators.

rcornwell:
Why is ncurses exporting these variable? I checked the ncurses manual and it does not list these variables.

pkooiman:
That’s due to an unfortunate design choice made by ncurses. See [here](https://github.com/mirror/ncurses/blob/87c2c84cbd2332d6d94b12a1dcaf12ad1a51a938/ncurses/tinfo/lib_data.c#L111) for the definition of the global exported SP symbol, PC and BC are exported similarly. Imagine how much “fun” it is chasing down the source of crashes due to clashes with those exported symbols, which as you rightly say are not documented.

markpizz:
> Why is ncurses exporting these variable? I checked the ncurses manual and it does not list these variables.

I have no idea.  The original issue / commit (c2cef3c) specifically adding the "#define PC PC_Global" was added to some simulators a bit  more than 10 years ago.   I don't recall the particular detail that caused that one to be tracked down.  There were some 10 simulators changed at that time.  I don't remember when your simulators were added to the simh/simh repo, but probably after that.

pkoning2:
I'm wondering if it might not be simpler/cleaner to do something like:
```
#define SP _dont_use_SP
#include <ncurses.h> /* or whatever it's called */
#undef SP
```
where the offending ncurses include file is referenced.  That way the simulators will just use the natural names and the offending globals from ncurses become inaccessible.

pkooiman:
That won't work unfortunately. The clash happens at runtime when the dynamic linker loads the ncurses libraries. It sees that both the simulator and the library export a symbol called "SP" and figures it must be the same symbol. In other words, because the ncurses dynamic libraries export a symbol called "SP" there is no way any program using the libraries can have a global (non-static) symbol also called "SP" without a clash occurring.

pkoning2:
Which OS?  I thought that the dynamic linking rule says that the main program can override a global from the library, and what's in the main program then takes effect.
But I suppose avoiding the wrongheaded symbol is a simpler answer.  I'm just a little bothered by it being a per-simulator workaround.  Could the #define go into sim_defs.h so it applies to everyone?

pkooiman:
The main program overriding the global from the library is the problem, because the library is trying to write to a symbol SP that should be internal to the library but it is in fact overwriting the main program's SP variable. 
The per-simulator workaround is definitely not ideal but I don't know how you'd make sure that a wholesale #define in sim_defs.h does not have unforeseen effects on any of the simulators.

markpizz:
It can't go in sim_defs.h since other simulators already have a #define SP which will then clash.

PR #464, opened 2025-04-25
Fix PDP-11 CIS documentation comments

thaliaarchi:
Only the string instructions document the registers used by the register-form instructions. Also document the BCD register-form instructions. Although, the operands have already been loaded into the special-purpose instructions before the op switch, I think this documentation is useful.

Also, fix the names of `L2DR` and `L3DR` in comments. These instructions are referred to as `L2Dr` and `L3Dr` in the [PDP-11/24 System Technical Manual](https://www.vt100.net/manx/details/1,23) and their [opcode strings](https://github.com/open-simh/simh/blob/2e0d51e9e97b2c3a28c55261108550d1f533916b/PDP11/pdp11_sys.c#L407-L413) call them `L2DR` and `L3DR`. These comments seem to be a simple typo.

For my education, where are good sources for information on the CIS? The manual linked above has good high-level information, but I have been unable to find more specific information on it besides SIMH's own implementation.


pkoning2:
I have not done anything to speak of with these instructions, but my main reference for PDP-11 details is EB-23657-18 PDP-11 Architecture Handbook.  Chapter 7 documents the "Commercial Instruction Set" in detail.

thaliaarchi:
Thanks! That's indeed a much more comprehensive resource.

Then, if I may, do you know where I could find information on the PDP-11 microcode? I've been toying with the idea of making a timing-accurate microcode emulator, but have found only resources sporadically mentioning the microcode.

pkoning2:
Re microcode, I believe that's a processor model specific question, and much of it is likely to be unavailable.  The PDP-11/60 had an option for user microprogramming, but I don't remember any others that did.  I don't think there is such a a thing as "PDP-11" microcode; instead, there is PDP-11/05 microcode, 11/44 microcode, and so on.  Oh yes, there is also the 11/20 which is a straight logic design, without the use of microcode.
You might ask on the list and see if others can tell you more; Bob Supnik knows far more than I do about this subject.

PR #469, opened 2025-06-29
PDP8: Add support for the Type 34 display

larsbrinkhoff:
This is a point-plotting display in the same vein as the PDP-1 Type 30.  You can set X and Y in the range 0-1023, and optionally "intensify" i.e. plot a point there.

The PDP-5 "ICBM" game by Len Berger is supplied as a test.

The IOT instructions for the Type 34 display are documented here:
https://archive.computerhistory.org/resources/text/DEC/pdp-5/dec.pdp-5.pdp-5_Manual.1964.102630361.pdf#page=46


larsbrinkhoff:
https://pico-systems.com/stories/1969.html

The PDP-5 computer was at Washington University: "_We had a cast-off DEC PDP-5 computer (basically an early version of the PDP-8, all discrete transistors) in a pretty much defunct lab called the Hybrid Computer lab_"

larsbrinkhoff:
@bscottm, good catch!  I have fixed that.

PR #471, opened 2025-07-04
3B2: Multiple bugfixes and improvements

sethm:
- Allow 3 MB RAM configuration (previously only 1MB, 2MB, and 4MB configurations were allowed)
- Allow SCSI CIO card to be used under 3B2/400 emulation (previously it could only be used under 3B2/700 emulation)
- Improved CTC, PORTS, and SCSI diagnostic checks
- Fixed a bug in IDISK device that allowed impossible disk configurations

The last update is a breaking change that disables the HD161 disk type by default, since real 3B2 hardware does not support it. The disk type will still allowed in backward compatibility mode through use of the "SET IDISK LARGE" command.

### DIFF
--- a/3B2/3b2_cpu.c
+++ b/3B2/3b2_cpu.c
@@ -355,11 +355,15 @@ static const char *att3b2_clock_precalibrate_commands[] = {
 
 MTAB cpu_mod[] = {
 #if defined(REV2)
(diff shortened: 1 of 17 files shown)


bscottm:
@sethm: Check line 423 in 3B2/3b2_scsi.c: `rqe` is never initialized.

Issue #434, opened 2024-11-11
Adding support for PDP- Major States within the CPU.


poetnerd:
- #### Context
- Oscar Vermeulen has produced kits that offer a 3/4 scale model of traditional DEC hardware with software emulated on a Raspberry Pi: PiDP-10, PiDP-11, and the first, and still very popular PiDP-8/i.  The joy of these kits is two-fold: experiencing old computer hardware and software architectures, and playing with the lights and switches of very-well emulated front panels. See: [PiDP-8/I: RECREATING THE PDP-8/I
](https://obsolescence.wixsite.com/obsolescence/pidp-8)
- The PiDP-8/i looks like the old PDP-8/i with the major registers and major states on display.
- SIMH doesn't emulate down to the Major States. It's an instruction-level emulation.  Indeed, the first foundational principle reported at [About the Open SIMH project(https://opensimh.org/about/) says, **_but not cycle-accurate, simulation._**
- [The PiDP-8/i software project](https://tangentsoft.com/pidp8i/wiki?name=Home) has developed, tested and benchmarked a fork of the OPEN SIMH PDP-8 emulation that supports the major states.  This is of value to the PiDP-8/i project because it means that the "SING STEP" button on the emulated PDP-8/i front panel actually works and steps through the Fetch/Defer/Execute Major States.
- Benchmarking says that performance with emulation of Major States is half the performance of the instruction-level emulation.
- We would really like to get out of the business of maintaining a fork of SIMH.  But we recognize that there's a performance penalty, and the proposed change is a step down the road of "cycle-accurate simulation."
- **QUESTION**: Might the PDP8 SIMH maintainers be willing to accept a pull request to add Major State support?

- #### the output of "sim> SHOW VERSION" while running the simulator which is having the issue

PDP-8 simulator Open SIMH V4.1-0 Current
Simulator Framework Capabilities:
32b data
32b addresses
no Ethernet
Idle/Throttling support is available
Virtual Hard Disk (VHD) support
RAW disk and CD/DVD ROM support
Asynchronous I/O support (Lock free asynchronous event queue)
Asynchronous Clock support
FrontPanel API Version 12
Host Platform:
Compiler: GCC Apple LLVM 15.0.0 (clang-1500.1.0.2.5)
Simulator Compiled as C (Release Build) on May 19 2024 at 13:01:35
Build Tool: autosetup+gmake
Memory Access: Little Endian
Memory Pointer Size: 64 bits
Large File (>2GB) support
[...]

PR #472, opened 2025-07-11
Update pdp8_cpu.c Rewrite of instruction decoder, implement Major States

poetnerd:
…tates.

* Update pdp8_cpu.c

Rewrite of instruction decoder to implement Major States.

* Update pdp8_cpu.c

Resolve issues with some C compilers

### DIFF
--- a/PDP8/pdp8_cpu.c
+++ b/PDP8/pdp8_cpu.c
@@ -406,7 +406,7 @@ reason = 0;
 // so it's no longer "uninitialized" with respect to the simulated execution. See
 // the TBD FIX in the code for the Start switch in main.c.in. When that is fixed,
 // this can be removed.
-if (IB = -1)
+if (IB == -1)
     IB = IF;
 ////////////////////////////////////////////////////////////////////////////////////
 
@@ -1497,3 +1497,4 @@ for (k = 0; k < lnt; k++) {                             /* print specified */
     }                                                   /* end for */
 return SCPE_OK;
 }
+
--- a/PDP8/pdp8_cpu.c
+++ b/PDP8/pdp8_cpu.c
@@ -23,6 +23,38 @@
    used in advertising or otherwise to promote the sale, use or other dealings
    in this Software without prior written authorization from Robert M Supnik.
 
+   ----------------------------------------------------------------------------
+
+   Portions copyright © 2015 by Oscar Vermeulen
+                      © 2016-2018 by Warren Young
+                      © 2021 by HB Eggenstein
+                      © 2021 by Steve Tockey
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+   THE AUTHORS LISTED ABOVE BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+   DEALINGS IN THE SOFTWARE.
+
+   Except as contained in this notice, the names of the authors above shall
+   not be used in advertising or otherwise to promote the sale, use or other
+   dealings in this Software without prior written authorization from those
+   authors.
+
+   ----------------------------------------------------------------------------
+
    cpu          central processor
 
    21-Oct-21    RMS     Fixed bug in reporting device conflicts (Hans-Bernd Eggenstein)
@@ -64,6 +96,9 @@
    MQ<0:11>             multiplier-quotient
    L                    link flag
    PC<0:11>             program counter
+   MA<0:11>             memory address
+   MB<0:11>             memory buffer
+   Major_State<0:1>     major state register
    IF<0:2>              instruction field
    IB<0:2>              instruction buffer
    DF<0:2>              data field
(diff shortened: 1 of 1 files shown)


bscottm:
@poetnerd: Check line 409 in pdp8_cpu.c. You're probably missing an `=`.

markpizz:
Actually, I reviewed your original issue about this and I have a few questions:

1) The original pdp8 code simulated instructions and was never intended to be cycle accurate.  The presumption now is that the simulator is cycle accurate.  
a) Is that true?
b) if so, must this cycle accurate mode be explicitly enabled.  If so, how?  If not, and thus the simulator is now just simulating cycles, then the updated PDP8 simulator should change some SCP global variables.  Specifically:
```
extern const char *sim_vm_interval_units;               /* Simulator can change this - default "instructions" */sim_
extern const char *sim_vm_step_unit;                    /* Simulator can change this - default "instruction" */
``` 

2) The SCP STEP command steps a particular number of cycles, right?  Then sim_vm_step_unit should point to "cycle".

Are these reasonable questions?

markpizz:
All of that makes sense and is fine.  Setting the mentioned global variables to "cycles" and "cycle" merely affects the output of some SCP core commands you can enter at the sim> prompt.  Most are related to some aspect of timing or throttling.

It is fine that  your group carefully considered the simulation performance impact of various parts of the changes you were making, but as I mentioned in the original where you described this project, as a practical matter, as long as simulators running on modern processors can execute instructions (or cycles) many times faster than the original hardware  there everyone will be fine with the results.

One thing that the change from instructions to cycles will actually affect is how the simulation framework (SCP) schedules simulated device activities.  Whenever a simulated device wants the framework to come back later to finish an I/O operation it will be calling sim_activate (uptr, NNNN), where NNN is the number of instructions (now cycles).  The changes you've made from instructions to cycles will affect when these device callbacks (aka activations) happen.  The result will be that fewer instructions will now get to execute before the respective callbacks.  The consequences will likely not be very significant, but there may be cases where the NNN values should probably be adjusted to somewhat larger values.

pkoning2:
As I said in the discussion for issue #434:

The question to me is what problem needs to be solved. Open SIMH is, fundamentally, an instruction emulator. It emulates instructions accurately enough that essentially all existing (and new, for that matter) code works correctly. But it does not aim to expose internal hardware details of specific implementations that are not visible from the programmer's point of view.

It seems to me that major state emulation is an example of a hardware detail not visible to the programmer. Here I am discounting the use of the "step single cycle" console panel switch, which is a service technician tool, not something a programmer would use. And in fact, I don't think I have ever seen an FE use it, either -- they tend to run diagnostics and, on older machines, poke around on suspect boards with their handy oscilloscope.

So the answer would be, yes it can be done, at a cost in performance, but why would we do it?

pmetzger:
@vrs42:
> Just because @markpizz hasn't always done things in the way you and I felt was most helpful, does not make it sensible to refuse his help when he is actually helpful, which is a fair bit of the time. More bad feelings just wastes all our time.

I don't particularly love the fact that he's free to take whatever changes we make here to improve SIMH and put them in his closed-source fork, and I don't like that he encourages people submitting bug reports to use his fork instead of the main SIMH.

pkoning2:
Perry, those are the rules of BSD-style licensed open source software.  The answer would be different if Open SIMH used GPL, but we don't and won't.  With our license, as for any BSD style license, anyone is free to make a closed fork of our work.
But let's not discuss that further.  The subject here should be what goes into Open SIMH.

poetnerd:
I've taken a stab at answering the question @pkoning2  asked in the discussion for issue #434.

QUESTION: Do we want to have that conversation there in that issue or here in this issue?

bscottm:
@poetnerd: "Cycle accurate" can mean a few different things. @markpizz interprets it as accurate in instructions/second or instruction cycles/second. There's the sense in which you're using it, namely, machine cycle accurate, i.e., simulating a complete machine cycle in the way that more deeply matches the way the original hardware worked.

I don't think you need to delve into making the PDP-8 simulator instructions/second accurate or make that the default setting.

rcornwell:
The problem I have with this PR, is that it is not really cycle accurate. It is missing three cycle break. And devices have not been modified to send data one word at a time. Also EAE has no cycles. The only point I can see of this change is to drive the PiDP8 LED's closer to actual hardware.

poetnerd:
Hello @rcornwell ,

You raise a good point.  Notice tht the PR is for "Adding Major States", and that's partly to stay away from the paradigm of "cycle-accurate" in keeping with the published principle of OpenSIMH, "instruction-accurate in preference to cycle-accurate".

TBH, I had not dug into how the EAE and the emulated device support had made use of the Major States.

I think it's an interesting question, "Is there value to updating the EAE and IOT code to make fuller use of the Major States?"

My selfish motivation is, "If Major States as offered in this PR is accepted, the PiDP-8/i project has a smaller footprint of our fork.

A Polyanna interpretation of the PR might be, "With Major States available, follow-on work could provide richer device emulation."

I know that someone is also working on a "Family of 8" update to SIMH that would allow run-time setting of "Straight-8" versus "8/i" versus "8/e" versus "VT278" versus "PDP-12" operation.  The 8/i and PDP-12 EAE don't have Mode B, and that change might benefit from utilization of Major States as a foundation.

My question to you, Richard, is: Do you think the PR should be rejected as incomplete, or that it could be accepted in the Polyanna interpretation, as serving as a basis for follow-on improvements?

rcornwell:
I am not sure whether this PR should be rejected or not. However here I way I would reject it.

1) This is not a completely cycle accurate PDP8 emulation.
   a) Lacks support for EAE.
   b) Lacks cycle support for devices.
   c) I don't see any way to enable single cycle operation.

2) The code is overly complex.

3) It does not serve much of a purpose in terms of emulation.

4) The commit message does not fit SimH convention.

On an emulator there is really no need to single cycle stepping (this was for debugging hardware). It also appears to be used specifically for PiDP8 users. When I did support for the PiDP10 there was little changes. Check out patch: https://github.com/rcornwell/sims/commit/e42f503b19fc37f72d1ea33f9962bc3c88aa4d97 to see changes for support of the PiDP10 panel.

I have a electrically accurate model of a PDP8e, that will run OS/8 and TSS/8. However there are differences between various PDP8 models that effects the cycles and results. I don't see these differences taken into account here.

pkoning2:
I posed the question to the Steering Group, and the answer from Bob Supnik (as I understand it) was that this change is not appropriate for the primary emulator.  But we allow multiple emulators for a given machine (you can see this in the pdp10 series of emulators, for example) so if you want to submit this as an alternate emulator (perhaps "pdp8-ms") that would be an acceptable contribution.

pkoning2:
Given the decision from Bob Supnik I quoted in the previous comment I'm going to close this PR.  As he said, an alternate simulator (but not the primary pdp8 simulator) along the lines you proposed would be an acceptable thing, if you would like to create a PR for that.

pmetzger:
@poetnerd Feel free to open something like this on the [ZIMH repository](https://github.com/pmetzger/zimh) although we would have to discuss how it would be incorporated.

PR #474, opened 2025-07-24
SIMH: Fix sim_printf_fmts.h for Visual Stduio 2013 and older

cmgauger:
Per Microsoft documentation versions Visual Studio 2013 and below do not support the hh, j, z, and t length prefixes for printf format-type specifiers. Further, the preprocessor selection logic was modified so that 64-bit Visual Studio versions are actually recognized.

### DIFF
--- a/sim_printf_fmts.h
+++ b/sim_printf_fmts.h
@@ -5,6 +5,7 @@
  * this header so that these formats are available to more than SCP.
  *
  * Author: B. Scott Michel
+ *         C. Gauger-Cosgrove
  *
  * "scooter me fecit"
  *~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~=~*/
@@ -15,7 +16,9 @@
 /* cross-platform printf() format specifiers:
  *
  * Note: MS apparently does recognize "ll" as "l" in its printf() routines, but "I64" is
- * preferred for 64-bit types.
+ * preferred for 64-bit types.  _MSC_VER and _WIN32 are always defined for Visual Studio
+ * builds. To check if the build is using 64-bit Visual Studio, you must check if the
+ * _WIN64 macro is defined.
  *
  * MinGW note: __MINGW64__ and __MINGW32__ are both defined by 64-bit gcc. Check
  * for __MINGW64__ before __MINGW32__.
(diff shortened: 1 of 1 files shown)


bscottm:
@cmgauger: Why the ancient compiler?

bscottm:
I ended up with some research time this afternoon, queried Perplexity.ai, which turns out to be quite useful and didn't turn out "botshit".

Here's what I suggest, pending your concurrence:

```
#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
#  if _MSC_VER >= 1900 || defined(__USE_MINGW_ANSI_STDIO)
/* VS 2015 or later, MinGW with ANSI stdio: Use the standard. Don't assume that
 * MS or MinGW will support the "I" qualifier indefinitely. */
#    define SIZE_T_FMT   "z"
#    define T_UINT64_FMT "j"
#    define T_INT64_FMT  "j"
#  elf (_MSC_VER < 1900 && defined(_WIN64)) || defined(__MINGW64__)
/* VS, 64-bit: use Microsoft's "I" qualifier.  */
#    define SIZE_T_FMT   "I64"
#    define T_UINT64_FMT "I64"
#    define T_INT64_FMT  "I64"
#  elf (_MSC_VER < 1900 && defined(_WIN32)) || defined(__MINGW32__)
#    define SIZE_T_FMT   "I32"
#    define T_UINT64_FMT "I32"
#    define T_INT64_FMT  "I32"
#endif
```

cmgauger:
I can agree with just about all of that change, except that even on 32-bit platforms the `T_UINT64_FMT` and `T_INT64_FMT` should be `I64`.

Because `int64_t`/`uint64_t` are 64-bit.

So altogether the new conditional block would be:

```C
#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
#  if _MSC_VER >= 1900 || defined(__USE_MINGW_ANSI_STDIO)
/* VS 2015 or later, MinGW with ANSI stdio: Use the standard. Don't assume that
 * MS or MinGW will support the "I" qualifier indefinitely. */
#    define SIZE_T_FMT   "z"
#    define T_UINT64_FMT "j"
#    define T_INT64_FMT  "j"
#  elif (_MSC_VER < 1900 && defined(_WIN64)) || defined(__MINGW64__)
/* VS, 64-bit: use Microsoft's "I" qualifier.  */
#    define SIZE_T_FMT   "I64"
#    define T_UINT64_FMT "I64"
#    define T_INT64_FMT  "I64"
#  elif (_MSC_VER < 1900 && defined(_WIN32)) || defined(__MINGW32__)
#    define SIZE_T_FMT   "I32"
#    define T_UINT64_FMT "I64"
#    define T_INT64_FMT  "I64"
#endif
```

As with regards to the C standard VS 2010 (and older) support, Herb Sutter mentioned this on his blog (in 2012):

> VC++ 2010 already fully supports the C subset of C++98, including things like <stdint.h> and declarations in the middle of a block.[*] The C subset of C++98 is approximately C95 (with very few incompatibilities with C95; i.e., there are very few cases where legal C95 code has a different meaning or is invalid in C++98) plus a few C99 features like declaring variables in the middle of blocks).

My own experience is that a lot of C99 code just works. But there's the occasional bit that blows up (oh hi there `stdbool.h`).

Unrelated to Visual Studio, but as you mentioned bringing in C99 libraries: That'll break the VAX build, because the HP C V6.4-005 compiler is not C99.

From the vim v9.1 Reference Manual (which is from all of five days ago as of writing this -- July 22, 2025):

> C COMPILER					assumptions-C-compiler
> 						ANSI-C C89 C90 C95 C99
> 
> Vim strives for maximum portability (see design-multi-platform and must still build with Compaq C V6.4-005 on OpenVMS VAX V7.3.
> 
> Therefore, the latest ISO C standard we follow is:
> 
> 	C95 (ISO/IEC 9899:1990/AMD1:1995)
> 
> In addition, the following two C99 features are explicitly allowed:
> 	– // comments, as required by style-comments;
> 	– the _Bool type.
> 
> Platform-specific code may use any newer compiler features supported on that platform.

But that's an entirely different topic, unrelated to this PR.

bscottm:
> I can agree with just about all of that change, except that even on 32-bit platforms the `T_UINT64_FMT` and `T_INT64_FMT` should be `I64`.

My bad. Even I get things wrong from time to time. :-) :-) :-)

> Unrelated to Visual Studio, but as you mentioned bringing in C99 libraries: That'll break the VAX build, because the HP C V6.4-005 compiler is not C99.

Not much I can do that. And I definitely can't ask the `libslirp` developer/maintainer to regress. Their code "is what it is."

cmgauger:
I've gone ahead and squashed the commits on this down into one commit, to clean things up some.

PR #475, opened 2025-07-26
CMake: Support RelWithDebInfo configuration

bscottm:
Support optimized compiles with debugging info (e.g., "-O2 -g" for GCC/Clang.) Useful with 'valgrind' to produce intelligible output.

### DIFF
--- a/README-CMake.md
+++ b/README-CMake.md
@@ -398,9 +398,11 @@ simh                      # Top-level SIMH source directory
 +-- BIN                   # Simulator executables (note 1)
 |   +-- Debug
 |   +-- Release
+|   +-- RelWithDebInfo
 |   +-- Win32
 |       +-- Debug
 |       +-- Release
+|       +-- RelWithDebInfo
 +-- cmake                 # CMake modules and build subdirectories
 |   +-- build-vs2022      # Build directory for VS-2022 (note 2)
 |   +-- build-vs2019      # Build directory for VS-2019 (note 2)
@@ -434,7 +436,7 @@ Notes:
      will appear directly underneath the `BIN` directory.
    - Multi-configuration builders (`msbuild`, `xcodebuild`): The simulator
      executables will appear underneath individual configuration subdirectories
-     ("Debug" and "Release").
+     ("Debug", "Release" and "RelWithDebInfo").
    - The Windows platform has its `Win32` subdirectory.
 2. The `cmake-builder.ps1` and `cmake-builder.sh` scripts create the
    `cmake/build-*` subdirectories as needed.
--- a/cmake/cmake-builder.ps1
+++ b/cmake/cmake-builder.ps1
@@ -87,7 +87,8 @@ param (
     [Parameter(Mandatory=$false)]
     [string] $flavor         = "vs2022",
 
-    ## The target build configuration. Valid values: "Release" and "Debug"
+    ## The target build configuration. Valid values: "Release", "Debug" and
+    ## "RelWithDebInfo"
     [Parameter(Mandatory=$false)]
     [string] $config         = "Release",
 
@@ -297,7 +298,7 @@ if (!$testonly)
 }
 
 ## Validate the requested configuration.
-if (!@("Release", "Debug").Contains($config))
+if (!@("Release", "Debug", "RelWithDebInfo").Contains($config))
 {
     @"
 ${scriptName}: Invalid configuration: "${config}".
--- a/cmake/cmake-builder.sh
+++ b/cmake/cmake-builder.sh
@@ -12,7 +12,7 @@ Configure and build simh simulators on Linux and *nix-like platforms.
 --clean (-x)      Remove the build subdirectory
 --generate (-g)   Generate the build environment, don't compile/build
 --cache           '--generate' and show CMake's variable cache
---parallel (-p)   Enable build parallelism (parallel builds)
+--parallel (-p)   Enable parallel builds
 --notest          Do not execute 'ctest' test cases
 --noinstall       Do not install SIMH simulators.
 --testonly        Do not build, execute the 'ctest' test cases
@@ -27,7 +27,8 @@ Configure and build simh simulators on Linux and *nix-like platforms.
                     msys2
                     mingw
                     ucrt
---config (-c)     Specifies the build configuration: 'Release' or 'Debug'
+--config (-c)     Specifies the build configuration: 'Release', 'Debug' or
+                  'RelWithDebInfo'
 
 --target          Build a specific simulator or simulators. Separate multiple
                   targets with a comma, e.g. "--target pdp8,pdp11,vax750,altairz80,3b2"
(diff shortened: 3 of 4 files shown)


README-CMake.md
```diff
@@ -434,7 +436,7 @@ Notes:
      will appear directly underneath the `BIN` directory.
    - Multi-configuration builders (`msbuild`, `xcodebuild`): The simulator
      executables will appear underneath individual configuration subdirectories
-     ("Debug" and "Release").
+     ("Debug", "Release" and "RelWithDebInfo").
```
pmetzger, line 439: An Oxford comma would be good here, I think. It gets confusing otherwise.

cmake/cmake-builder.ps1
```diff
@@ -87,7 +87,8 @@ param (
     [Parameter(Mandatory=$false)]
     [string] $flavor         = "vs2022",
 
-    ## The target build configuration. Valid values: "Release" and "Debug"
+    ## The target build configuration. Valid values: "Release", "Debug" and
```
pmetzger, line 90: Oxford comma is less needed here but I'd still recommend it.

cmake/cmake-builder.sh
```diff
@@ -27,7 +27,8 @@ Configure and build simh simulators on Linux and *nix-like platforms.
                     msys2
                     mingw
                     ucrt
---config (-c)     Specifies the build configuration: 'Release' or 'Debug'
+--config (-c)     Specifies the build configuration: 'Release', 'Debug' or
```
pmetzger, line 30: I recommend an Oxford comma again.

cmake/cmake-builder.ps1
bscottm: @pmetzger: My writing habits were pretty ingrained by my PhD advisor's admin, who proofed and edited my dissertation. She completely disabused me of writing with Oxford commas. Happy to have a five gin martini debate elsewhere.

pmetzger: She was wrong. Without an oxford comma, technical documentation becomes ambiguous. https://medium.com/english-language-style-usage/the-oxford-comma-15e3fb65ce21 and many other links will explain the issue.

pmetzger:
FYI, I've applied a version of this to the ZIMH repository on a branch; see https://github.com/pmetzger/zimh/commit/3b0346c9

LegalizeAdulthood:
What's amazing to me is how much all of this isn't needed if you just use [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html), which have been available since November, 2020.

pmetzger:
I'm always happy to get advice from CMake experts. I had no idea that CMake presets existed, but I am not a CMake expert; CMake is a big, complicated system and I don't know enough about using it. Thank you for pointing out their existence.

LegalizeAdulthood:
For the benefit of readers, I'll summarize.  The details are all covered in the [CMake-Presets documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).

The reason that people write helper scripts, like the powershell script in this PR, is because typing command lines like this more than once becomes tedious:
```
# configure
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=OFF -B ../my-preferred-build-dir -S . -G "Ninja Multi-Config"
# build
cmake --build ../my-preferred-build-dir --config RelWithDebInfo
# we skipped the tests, so don't run ctest
# package
cd ../my-preferred-build-dir
cpack --config CPackConfig.cmake -G ZIP -C RelWithDebInfo
```
A preset is a JSON file that contains settings that you would normally by typing as command-line arguments above.  For instance, the entirety of the above can be accomplished with a workflow preset:
```
cmake --workflow easy-peasy
```
The workflow preset defines the steps of configure, build and package:
```
"workflowPresets": [ {
  "name": "easy-peasy",
  "steps": [
    { "type": "configure", "name": "easy-peasy" },
    { "type": "build", "name": "easy-peasy" },
    { "type": "package", "name": "easy-peasy" }
    ]
} ]
```
With the corresponding configure preset:
```
"configurePresets":[{
  "name": "easy-peasy",
  "generator": "Ninja Multi-Config",
  "binaryDir": "../my-preferred-build-dir",
  "cacheVariables": {
    "CMAKE_BUILD_TYPE": "RelWithDebInfo",
    "BUILD_TESTING": false
  }
}]
```
and build preset:
```
"buildPresets": [{
  "name": "easy-peasy",
  "configurePreset": "easy-peasy",
  "configuration": "RelWithDebInfo"
}]
```
and package preset:
```
"packagePresets": [{
  "name": "easy-peasy",
  "configurePreset": "easy-peasy",
  "generators": [ "ZIP" ],
  "configurations": [ "RelWithDebInfo" ]
}]
```

For a small project, this probably suffices, but if you want to get fancier, you can define presets as building blocks and use an inheritance chain to compose presets together.  There's also a limited form of macro expansion, access to environment variables and conditional usage of presets (for instance, disabling a preset that is macOS specific on a host environment that isn't macOS).

This is why I say that these helper scripts designed to supply repeated arguments to CMake, CTest, and CPack are superfluous.  Presets already handle all that.  This repository doesn't (yet) have a `CMakePresets.json` file that would define presets for everyone to use, but you can always create your own presets in the file `CMakeUserPresets.json` (which shouldn't be added to source control as it's not designed to be shared; use `CMakePresets.json` for shared presets).

In addition to making your life at the command-line easier, being a well-defined JSON file with a schema to validate the contents, various IDEs and other tools understand how to parse preset files and can supply you with options in the IDE to say "use this preset".  Bespoke helper scripts will never be understood by IDEs and other tooling.

bscottm:
@LegalizeAdulthood: The presets tend to explode when dealing with multiple platforms. I did submit a PR at one point or another with an initial set of presets (or maybe I solicited your feedback privately, it's been a while.)

LegalizeAdulthood:
Yes, at the moment, there are some scenarios where presets become a combinatorial expansion.  I think that will be addressed as presets evolve because it's a common pain point.  Personally, I think the answer to this is for users to compose their own combinations in their own personal `CMakeUserPresets.json` file from building blocks provided by the `CMakePresets.json` file.  Specific combinations used for CI builds should also be present in the `CMakePresets.json` file in order to simplify the workflows for CI builds.  In my fork, I have presets and packages building for macOS, linux and Windows for both arm64 and x64 and I didn't need a large number of presets.

bscottm:
@LegalizeAdulthood So open a PR with an initial set of presets, already! It's not as if they can't be iterated upon and extended.

LegalizeAdulthood:
@bscottm I have other open PRs right now and I'm waiting for those to be accepted first.  Acceptance of PRs is very slow on this project, so I don't want to send low priority PRs in before higher priority ones are accepted.  I have my own fork where presets are working just fine.

LegalizeAdulthood (approved):
The cmake changes are fine.

bscottm:
@LegalizeAdulthood: My point was to encourage you to share the obvious improvement, since I'm very, very unlikely to return to open-simh.

pkoning2:
Re GCC/Clang with -g -O2: I learned long ago that this combination works acceptably well with gcc (though -O1 makes it easier because it limits reordering).  But I never had any success debugging Clang output with anything better than -O0.  And that is with lldb -- clang with gdb is even worse.  But perhaps others have had better luck.

PR #487, opened 2025-08-27
SCP: video improvements

bscottm:
Re-engineered sim_video support. Submitted as two commits so that it's easier to cherry-pick the second commit.

The first commit is the re-engineered video that removes the `SDL_main()` asymmetry between macOS, Linux and Windows -- they all use `SDL_main()` if video is needed. Video is also automagically rescaled if it can't fit on the physical screen; SDL2 has been able to handle this for a long time. Audio is also asynchronous and won't stall the simulator while the tone plays out. There's a new "onto" API that allows the simulator to draw pixels directly into the SDL texture, reducing the amount of video frame buffering. Lastly, users can enable native hardware acceleration, for what it's worth; simulator video is not limited to the SDL software renderer.

The second commit is sim_timer.c cleanups, notably using actual constants for various conversions, e.g., `MSEC_PER_SEC_l` for millisecond to second (long) conversions. It also corrects the long-running issue on Windows where `Sleep(1)` can sleep for 1 + 15 + 5 milliseconds (requested delay + timer interrupt periodicity + wakeup latency) by using waitable timers.


bscottm:
Retracting temporarily -- strange refresh beat frequency on Linux X server.

PR #488, opened 2025-08-27
Not finding vmnet shouldn't be a fatal error

NattyNarwhal:
Allows the build to continue if vmnet isn't supported, either because the deployment target is too old (before 10.10) or the compiler lacks support for blocks (GCC).

See GH-486. I don't have an older Mac handy to test this with.


markpizz:
You should fix the makefile with appropriate tests as well.

PR #492, opened 2025-09-01
SCP: Add a -F option to SET <dev> DISABLE command to force detach and cancel

markpizz:
Historically some devices may not have been able to be DISABLED once they've been ENABLED.

This change allows such a device to be disabled when SET {-F} <device> DISABLE command is invoked with the -F flag specified.

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -1649,7 +1649,9 @@ static const char simh_help2[] =
       " specified device from the configuration.  A DISABLED device is invisible\n"
       " to running programs.  The device can still be RESET, but it cannot be\n"
       " ATTAChed, DETACHed, or BOOTed.  SET <device> ENABLED restores a disabled\n"
-      " device to a configuration.\n\n"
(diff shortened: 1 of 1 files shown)


hharte:
Hello @markpizz 

Instead of requiring -F, why not disable the timers unconditionally if the device does not have any attached units?

Thanks,
Howard

markpizz:
> Instead of requiring -F, why not disable the timers unconditionally if the device does not have any attached units?

Patrick had units doing timing activity.  Units could be doing anything.  

In any case, the point here is that a functional change to how a command behaves conceptually requires different input.  For whatever reason, someone might want the prior behavior, so it remains as the default.

The -F forces disabling of devices with attached units or units in the event queue.

pmetzger:
Committed to my ZIMH fork, with some modifications, in https://github.com/pmetzger/zimh/commit/2d4c0f0a7cda15f88c28c — note that I also added documentation patches, unit tests, and changed one error handling path a bit.

Issue #499, opened 2025-09-23
PDP11: EXAMINE of register arrays may display incorrect data


shattered:
Transcript (start the simulator with no configuration file):
```
sim> ex rx sbuf[0/4]
SBUF[0]:        000
SBUF[1]-SBUF[3]: same as above
sim> dep rx sbuf[2] 1
sim> ex rx sbuf[0/4]
SBUF[0]:        000
SBUF[1]:        001
SBUF[2]:        001
SBUF[3]:        000
sim> ex sbuf[1]
SBUF[1]:        000
sim> ex sbuf[2]
SBUF[2]:        001
```

Seems to be an off-by-one somewhere, but I don't quite understand the code to find it.

markpizz:
If or when #502 is merged, the reported problem will be fixed.

pmetzger:
Fixed on the ZIMH fork in https://github.com/pmetzger/zimh/commit/d106b1e65f167a0216472f5a068f6c0c4a007ca3 which also includes tests.

shattered:
Please consider merging #492 and #500 into ZIMH, too

pmetzger:
I will look at them in a few hours!

pmetzger:
@shattered I have committed 492 and 500 into ZIMH, as requested.

PR #502, opened 2025-10-01
SCP: Fix output produced when EXAMINE an array register

markpizz:
This problem was a latent bug introduced when summarized array register output was added back in 2013 or 2018.

This fixes the problem reported in https://github.com/open-simh/simh/issues/499


pmetzger:
Fixed on the ZIMH fork in https://github.com/pmetzger/zimh/commit/d106b1e65f167a0216472f5a068f6c0c4a007ca3 which also includes tests.

PR #504, opened 2025-10-02
SCP: Fixed typo in sim_card.h

rcornwell:


bsdphk:
This patch still leaves "Top four bits..." line which totally wrong.

That line is also wrong in sim_card.c, where it is equally wrong.

bsdphk:
looks better to me now :-)  Thanks!

pmetzger:
Fixed in the ZIMH fork in https://github.com/pmetzger/zimh/commit/6a47e48f3932a2961755d2817477e9877e0d555d

PR #505, opened 2025-10-11
LINC: New emulator for the classic LINC computer.

larsbrinkhoff:
This emulates the classic LINC from 1965.  Out of scope: earlier models e.g. with only 1024 words of memory, or newer models such as micro-LINC 300, LINC-8, or PDP-12.  The emulator passes diagnostics provided by the System Source Museum.  It can run Mary Allen Wilke's LAP6 "LINC assembly program" which provides text editing, assembling, and tape filing.

~~I didn't dare mess with the Visual Studio projects.  Someone, help?~~  Fixed, Mark P sent me the files.

The first commit adds support for the "left backslash" key between left shift and Z.  It's useful because the LINC keyboard had a key there.  It's also mapped from the regular left bracket key.
image

Typing is a bit quirky.  I haven't decided on the kep mapping yet.  For now, host keys are mapped to the closest corresponding LINC key.  But some LINC keys require pressing CASE first to get the upper case symbol.  E.g. to get comma, press *and release* CASE (mapped from shift), and type comma.  Plain comma generates minus, since that is the unshifted (or un-CASE'ed) key.  Typing minus also generates minus.  Zero is mapped from both 0 and backtick, because the LINC keyboard had 0 to the left of 1.  LAP6 uses CASE-0 and CASE-Q to move text by the page, and CASE-1 and CASE-W to move by the line; this arrangement is more apparent with the original key placement.  F1 is mapped to the p/u key; I can think of no other suitable key.  The LINC has three lower case letters: i, p, and u, but they are used as special assembly language symbols, not text.


LegalizeAdulthood:
> I didn't dare mess with the Visual Studio projects. Someone, help?

We'll also want to add it to the CMake build.

larsbrinkhoff:
Oh right.  I added the CMake stuff.

larsbrinkhoff:
@markpizz sent me the Visual Studio files and I added them here.

Visual Studio Projects/linc.vcxproj.user
LegalizeAdulthood, line 1: Visual Studio `.user` project files shouldn't be added to source control

markpizz:
@LegalizeAdulthood said: 
> Visual Studio .user project files shouldn't be added to source control

Well, the .vcxproj, .vcxproj.filters and .vcxproj.user files for all the simulators were added back in February of 2023 by @tlhackque.  None of them actually needs to be in source control, since they will be autogenerated when a newer Visual Studio version opens the simh.sln file.  This takes several minutes, so it seems reasonable as to why they were added,  The .user files have nothing special or concerning in them except for 4 lines of XML.

The new linc.vcxproj file has been provided the same way all the other existing .vcxproj files were including the .vcxproj.user files.

LegalizeAdulthood:
They aren't intended to be shared among developers; they are per-user settings.  They don't belong in source control.  It doesn't matter if they were added previously because a previous mistake doesn't make for a correct action now.

larsbrinkhoff:
I'm happy to remove the .user file, and will probably do so given the information above.  I'll raise the more general issue on the mailing list.

larsbrinkhoff:
This documentation should make it easier to use the emulator:

https://github.com/larsbrinkhoff/open-simh/blob/lars/linc/linc/README.md

Visual Studio Projects/linc.vcxproj.user
larsbrinkhoff: Removed.

pkoning2:
I'm still seeing a stream of commits.  Could you indicate when you think it's finished?  It seems more logical to merge it at that point rather than in pieces.

A comment on the keyboard: typically the best mapping is one that matches the markings on the keys rather than the keypresses.  So while you indicate that typing comma on the original keyboard means pressing shift then the minus/comma key, it would make sense for the modern keycode comma to produce that code, or sequence of codes.  
Sometimes it is necessary or helpful to have a mapping that represents key positions.  That could be a mode, not the default but perhaps a "raw" or "key-positions" mode.  I did something analogous in an emulation for a PLATO terminal, where the shifted digits are marked differently than on modern keyboards (closer to ASR33) and it would do that as an option.  But not the default because most of the time it isn't what people prefer.

pkoning2:
Silly question perhaps: does this emulation include PDP-12 emulation, or is it only the base LINC?

vrs42:
On 10/14/2025 9:57 AM, Paul Koning wrote:
> *pkoning2* left a comment (open-simh/simh#505) <https://github.com/open- 
> simh/simh/pull/505#issuecomment-3402814485>
> 
> Silly question perhaps: does this emulation include PDP-12 emulation, or 
> is it only the base LINC?

My fork has some progress toward a PDP-12 simulation.  (It can also 
emulate various models of the PDP-8, not just the 8/E.)  The main thing 
missing is the LINCtape drive emulation (which is of course a big deal).
https://github.com/vrs42/simh

	Vince

larsbrinkhoff:
@pkoning2, it's "finished" (if any software ever is) in the sense that it's ready to go in at any time.  I think passing all the diagnostics was the milestone I considered it ready.  So please press that merge button any time you like.

As long as this pull request has not been merged, I have been making some minor updates.  I received some feedback which I wanted to address.  I may have some larger updates in the future, but that's definitely for a new pull request.

There are no trace amounts of PDP; this only emulates the classic LINC from 1965.  Out of scope: earlier models e.g. with only 1024 words of memory, or newer models such as micro-LINC 300, LINC-8, or PDP-12.

larsbrinkhoff:
@vrs42 wrote:
> The main thing missing is the LINCtape drive emulation

The LINC diagnostics are quite particular about the relative timing between the processor and tape motion.  It will start the tape and expect to see it at roughly the right place some instruction loops later.  It will check to see the narrow interblock zones between data blocks at the right time.  I have the tape motion simulated to great detail.  The tape is moving at a rate of 160 microseconds per word.  The svc routine is called for each word to update the simulation, so I know at all times whether the head is positioned at data,  a block number, a checksum, or an interblock zone.  The CPU accounts for the number of memory cycles at 8 microseconds each.

larsbrinkhoff:
I continue tweaking as long as the pull request is open.  The implementations of DSC and MUL are now closer to the schematics and will side-effect the internal states accordingly.  Added F11 to toggle full screen.

larsbrinkhoff:
FYI @spacehobo

linc/linc_cpu.c
```diff
@@ -0,0 +1,988 @@
...
+
+static void
+cpu_misc(void)
+{
+  switch (C) {
+  case 00000: //HLT
+    stop_reason = STOP_HALT;
+    break;
+  case 00002: //PDP
+    sim_debug(DBG_CPU, &cpu_dev, "This is not a DEC machine.\n");
```
spacehobo: Extremely petty nit to pick, but the LINC machines *were* largely built by DEC.  It's not a PDP, but most assembly was done at the Maynard mill!

larsbrinkhoff: *The best way to get correct information online, is to post slightly wrong information online.* - Myself.

@spacehobo, thank you for this.  Can you expand a bit, please?  How many were made at MIT (around 50 I heard), and how many by DEC?

spacehobo: It gets murky!  So first of all they were all made with DEC system modules (the precursor to flipchips), so there's always a connection there.  I assume the prototype "alpha" LINC was mainly assembled at MIT, but then the next 12 were kind of a mix: DEC sent over incomplete machines, and Charlie Molnar decided it would be good to give the biomedical researchers a chance to practice repair skills by finishing the assembly.

So there's a cool anecdote (with photos to back it up!) that these researchers sat through the lectures on how to lace the wires together.  The MIT folks taught a standard lineman's technique.  The researchers smiled, stood up, and then laced the cables using a surgeon's suture technique they'd all learned in medical school!

Anyway, after that the group moved to a space in Kendall Square (because their funding came from the National Institutes of Health rather than Lincoln Labs) and sold some more.  I expect most of those were more completely assembled by DEC.  By 1964 they'd moved to St. Louis to make the LINC III with 2k of core.  I'm not sure what the relationship with DEC was by then, but [the backplane wiring plans were still being updated within Digital that year](https://www.computerhistory.org/collections/catalog/600000502/record/102734402/).

And then you get stuff like the LINC-8 in the late 60s, leading to the PDP-12.  Those were clearly 100% DEC built, but only the 12 had that `0002` instruction: the LINC-8 required hardware knobs to change mode.

spacehobo: image
Also, here's a brochure from 1964 where DEC were advertising the LINC as "made and sold by Digital Equipment Corporation".  I'd imagine this sort of marketing took a back seat once the PDP-8 was out.

larsbrinkhoff: Found some numbers in *Early and Small*.  At MIT, around 1963, there was "around 12".  16 is mentioned in some context, as is 6.  DEC was to have made 50-60.

The [Wikipedia page](https://en.wikipedia.org/wiki/LINC) say "Units shipped 50", and "Units sold 21 (by DEC)".

I'll run these numbers by my LINC guys.

spacehobo: Yeah, Mary Allen Wilkes said in her oral history interviews that the "original 12" NIH grant recipients were supplemented by a few extras (so 16 seems reasonable for the LINC II units).  She went around to all these installations to do user testing when working on LAP5 and LAP6, as is described in her CACM article as well.  I believe one of the NIH recipients had their LINC repossessed for not making effective use of the equipment (a rather infamous dolphin researcher, if I remember right).

I suspect the bulk of the LINCs in that 50-60 figure were LINC IIIs with the expanded memory, like the one at MIT that was retired in 1993.  Those were the ones that could run LAP6 without modification.  I suspect people largely ordered them from the group at Washington University, but they were still assembled by DEC.

larsbrinkhoff: Can you expand on the LINC II and III (and I guess I) monikers?  I recognize LINC III from https://bitsavers.org/pdf/washingtonUniversity/linc/LINC_III_Order_Code_Jun63.pdf.  I would have expected memory expansion to come around 1965.

larsbrinkhoff:
This has still not been merged, so I added a TTY device.  This is a teletype connected to relay output 0.  LAP6 will bit bang this output to send text for printing.  (So the teletype can't be used for input.)  The emulator samples the output to find the start bit edge, and then decodes bits at 110 baud (relative to CPU speed).  Tested with PM (print manuscript) and PX (print index) from LAP6.

linc/linc_cpu.c
spacehobo: The models I've seen are the "Alpha LINC" (which was a prototype), and then "Classic LINC" which seems to have somewhat fuzzy boundaries but I always interpreted to mean "any production 1k LINC from 1962-1964", and then "LINC III" are the 2k LINCs.  I don't know if there was a LINC I, or if anyone was even considering this naming scheme before they made the III.

I wouldn't put it past Wesley Clark to decide that they needed a name to make the technology seem solid and mature after they moved to St. Louis.

spacehobo: The second-to-last page of [Programming the LINC](https://bitsavers.org/pdf/washingtonUniversity/linc/Programming_the_LINC_Second_Edition_Jan69.pdf) from 1969 seems to use "Classic LINC" to mean "any LINC that came out of Wes Clark's group, wherever they were".  It's meant to contrast against the more powerful ones from Spear and Digital's PDP-8 hybrids.

I wonder if "LINC III" just means "Wilkes's Third Edition of the LINC ISA" in that case.  It's possible different areas of the LINC group were using different terminology.

spacehobo: The meaning of "classic LINC" I originally used is based on Charlie Molnar's 1990 ACM presentation: http://dx.doi.org/10.1145/89482

vrs42:
On 10/18/2025 5:20 AM, Abolish ICE wrote:
> I believe one of the NIH recipients had their LINC repossessed for not 
> making effective use of the equipment (a rather infamous dolphin 
> researcher, if I remember right).
 From what I read, the LINC was totally unsuited to the corrosive 
environment at sea, and thus basically unusable by Dr. Lilly. 
(Especially the LINCtape drives.)

larsbrinkhoff:
Updated again to add a modified DO command.  Without arguments, it will act like the DO button which executes a command from the switches.  (With an argument, run a script like usual.)

BOOT TAPE now accepts RCG=, RDC=, and START= to say how to boot.  It works like toggling in a tape read command, pushing DO, and then starting at the specified address.  The default is RCG 7300 and starting from 20; this is the conventional way to start LAP6.

linc/linc_cpu.c
spacehobo: On further review, it seems I was way off the mark with my use of "LINC III": the ISA manual is from 1963, long before the 2k LINCs were out.  It's clear that this was just r3 of the LINC ISA.  I think I was mostly confused by the fact that the tape instructions could address 2k of core, which may have been part of an advocacy effort on her part (she describes the expanded memory as "an apple in my eye, or a bee in my bonnet" in the oral histories).

larsbrinkhoff:
Added symbolic input.

larsbrinkhoff:
There's a companion repository with a LAP6 compatible assembler and filing tool.

https://github.com/PDP-12/lap6-tools

larsbrinkhoff:
Steering committee @pkoning2 et al, feel free to merge this at any time.  It's ready to go.  I may push minor tweaks in the meanwhile, but that doesn't mean the emulator is unstable.

larsbrinkhoff:
Future directions: I might add support for the Spear, Inc micro-LINC 300 if software and documentation becomes available.  There are LINCtapes with tests for the 300, but I would also like to see some more useful software.

I'm unlikely to make additions to support the LINC-8 or PDP-12.  Please refer to @vrs42's https://github.com/vrs42/simh for PDP-12 emulation.

pkoning2:
Thanks Lars.  I would hope that other work will get merged here at some point.  Having Open SIMH forks is perfectly valid of course, but it's more desirable to have new emulators merged.  Otherwise things can diverge painfully far, as we can see with the unfortunate case of XHomer.

pkoning2:
Using your current code, "make linc" works fine and the self-test passes quickly.  But "cmake/cmake-builder.sh --target linc" does the build without problem but the selftest hangs and eventually times out.

pkoning2:
By the way: I try to avoid force-push and amended commits after the original push, but if you must, in the future at least please include `--date=now` in the commit command.  It's strange to see new commits with old dates.

larsbrinkhoff:
I would recommend you do revise commits in an open pull request.  If well done, it makes pull requests easier to review, and the history more useful in the future if it's necessary to go back and check what happened.  I can very well use --date=now.

I tested cmake and the tests passed for me.  Is this something that happens consistently?

pkoning2:
It was consistently broken before your last commit, and worked after that, which is why I did the merge.

PR #509, opened 2025-10-24
Avoid redefining `bool` in GCC 15+

rbanffy:
On c23 and onwards, `null` is a keyword and can't be redefined when compiling with GCC 15. This change checks for GCC and the GCC version and skips the line if that's the case.


rbanffy:
Never mind. There is more in this PR than it should.

PR #513, opened 2025-10-27
Fix bool typedef slirp glue

OpenPosixDev:
This PR fixes a bug in `slirp_glue/config-host.h` where the `bool` type was not correctly defined in environments using C versions prior to C23 (e.g., C99 or C11). The `typedef` for `bool` is only defined as `int` when the C version is lower than C23. Starting with C23, `bool` is automatically supported, so this typedef is only provided for older C versions.

### Changes:
- Added: Conditional definition of `bool` as `int` for C versions lower than C23.
- No functional changes for C23 or newer versions.
- These changes ensure compatibility with older C versions and avoid conflicts with `bool` in C23 and later.

### Testing:
- This change has been tested on systems using C99 and C11 to ensure that the `bool` type is correctly defined.
- No changes in behavior when using C23 or higher.

### Additional Notes:
- This change is backward-compatible and should not interfere with existing implementations.

### DIFF
--- a/slirp_glue/config-host.h
+++ b/slirp_glue/config-host.h
@@ -12,7 +12,7 @@ typedef int SOCKET;
 #endif
 
 #ifndef  __cplusplus
-#if __STDC_VERSION__ >= 202300L
+#if __STDC_VERSION__ < 202300L
 typedef int bool;
 #endif
 #endif
--- a/slirp_glue/config-host.h
+++ b/slirp_glue/config-host.h
@@ -12,7 +12,9 @@ typedef int SOCKET;
 #endif
 
 #ifndef  __cplusplus
+#if __STDC_VERSION__ >= 202300L
 typedef int bool;
+#endif
 #endif
 #ifdef _MSC_VER
 #include <win32/stdint.h>


rbanffy:
Did the issue appear to you with the `make` build process or only with `cmake`? The former is passing the --std=c99 option to the compiler. The issue seems to be that the `cmake` workflow isn't passing it.

OpenPosixDev:
Thanks for the fast reply! The problem only occurs with CMake, as it's not passing the --std=c99 flag to the compiler. The fix works fine for me, but the CMake workflow could definitely be improved if needed.

rbanffy:
I believe that's the right way to solve this issue, but my knowledge of cmake is not the best.

OpenPosixDev:
Same for me.

bscottm:
@OpenPosixDev: While passing the "-std=c99" flag might fix the problem in the short term, but take a deeper look at the slirp code base in open-simh. It's a very hacky adaptation from the QEMU code base from around 8 years ago. All of the types are fixed in place where a `configure` would otherwise have provided the proper platform definition (among other things.)

If you could, try the commit in PR #485 where some of your proposed fixes might already be addressed. Might not specifically address the slirp code base (been a couple of weeks since I looked at it.) Might be a better place to merge fixes.

pmetzger:
@bscottm FWIW, ZIMH now uses externally maintained SLiRP packages. It works very well and took very little work to accomplish. No point in vendoring code if you are just going to run it unchanged anyway.

PR #517, opened 2025-11-16
PDP-11: RP-11: Fixing device register read / write consistency

al20878:
1. Making sure that only expected values appear on the bus when registers are read by CPU;
2. Fix order of I/O cancellation in case of disk detachment with pending I/O;
3. Fix WLOA bug and make protection consistent with documentation;
4. Added more comments.


pmetzger:
Also committed, along with some tests, in ZIMH's https://github.com/pmetzger/zimh/commit/042a358e

PR #519, opened 2025-12-04
[I650] fix: wrong capitalization in soap and is .ini scripts

ch1pkav:
Scripts `I650/sw/Build_is_from_decks.ini` and `I650/sw/Build_soap_from_source.ini` are incorrectly named for `I650/650_demo_all.ini` to function correctly. `File not found` errors are thrown immediately after running the demo. Moreover, their capitalization does not match any of the other .ini scripts (all start with lowercase letters). This change renames those files with the correct capitalization.


pmetzger:
I've fixed this in my ZIMH fork in commit https://github.com/pmetzger/zimh/commit/65e09038606d94210e8cca08374ee966edbaf184

PR #523, opened 2026-01-08
Fix build failure of asm1130 on recent Gentoo (gcc version 14.2.1)

romforth:
I can build asm1130 using my ancient Ubuntu version where gcc -v shows:
gcc version 5.3.1 20160413 (Ubuntu 5.3.1-14ubuntu2)

But using a recent version of Gentoo where gcc -v shows:
gcc version 14.2.1 20241221 (Gentoo 14.2.1_p20241221 p7)
the build (`make asm1130`, within the Ibm1130/utils subdirectory) fails with:

---snip---
asm1130.c: In function ‘save_symbols’
asm1130.c:2132:9: error: implicit declaration of function ‘unlink’; did you mean ‘x_link’? [-Wimplicit-function-declaration]
 2132 |         unlink(SYSTEM_TABLE);
      |         ^~~~~~
---snip---

`man -s2 unlink` shows that `#include <unistd.h>` is needed but that is
currently ifdef'ed under VMS. So the fix is to use it unconditionally.

I've tested that this works on both the old and newer gcc versions.

BTW, I was using `asm1130` as the cross-assembler for my port of
[romforth](https://github.com/romforth/romforth) which is dialect of Forth
to the IBM 1130 (only under emulation, using SIMH/ibm1130 as the emulator)

### DIFF
--- a/Ibm1130/utils/asm1130.c
+++ b/Ibm1130/utils/asm1130.c
@@ -157,9 +157,7 @@
  * Comment out this define to make @ and ' different in symbol names, keep to make equivalent
  */
 
-#if defined(VMS)
     #  include <unistd.h>                   /* to pick up 'unlink' */
-#endif
 
 #define BETWEEN(v,a,b) (((v) >= (a)) && ((v) <= (b)))
 #define MIN(a,b)       (((a) <= (b)) ? (a) : (b))
--- a/Ibm1130/utils/asm1130.c
+++ b/Ibm1130/utils/asm1130.c
@@ -157,7 +157,9 @@
  * Comment out this define to make @ and ' different in symbol names, keep to make equivalent
  */
 
+#if defined(__linux) || defined(VMS)
     #  include <unistd.h>                   /* to pick up 'unlink' */
+#endif
 
 #define BETWEEN(v,a,b) (((v) >= (a)) && ((v) <= (b)))
 #define MIN(a,b)       (((a) <= (b)) ? (a) : (b))


pkoning2:
Have you tested it on Windows?

pkoning2:
It doesn't look like the Github test scripts build this subdirectory.  I wonder about an include of "unistd" on Windoze.

markpizz:
@pkoning2 said: It doesn't look like the Github test scripts build this subdirectory. I wonder about an include of "unistd" on Windoze.

FYI, there has been no effort to avoid excess warning messages when building things in this directory.  In fact, I've never known anyone who actually tried to run any of the makefiles in this directory.  On windows, the builds have many warnings, but all but checkdisk.mak produce binaries.  Looking closely at that build failure and the details in the checkdisk.mak seem to refer to include files from way before Visual Studio 2008 which is the oldest I've got.  Removing the irrelevant references in checkdisk.mak and a binary is built.  All the windows binaries run and spit out a usage message.

The simplest change to the asm1130.c source would be to change:

    #if defined(VMS)
    to 
    #if !defined(_WIN32)

pkoning2:
Thanks Mark.  @romforth  could you make that update?

markpizz:
As I mentioned, there was all sorts of warning messages produced when building all those programs on windows.  When building asm1130 on Ubuntu, many warnings are also produced, including about unlink.  Meanwhile, a binary is produced and no link errors were generated, so I suspect the result works just fine.

markpizz:
My point is that removing a single warning among many, if the resulting binary works fine, probably doesn't warrant a dedicated change.

markpizz:
On Ubuntu I see:
```
$ make asm1130 2>&1|grep unlink
asm1130.c:2132:9: warning: implicit declaration of function ‘unlink’; did you mean ‘x_link’? [-Wimplicit-function-declaration]
 2132 |         unlink(SYSTEM_TABLE);
```
Your original complaint has the same message, but in your case we see "error:" rather than "warning:".  Is something about your build causing warnings to return as errors?

pkoning2:
It's probably caused by gcc moving to a newer edition of the C standard for its default dialect, one in which this issue is an error.

romforth:
Hi Paul and Mark,

I didn't have a Windows system to test - so the only testing I did was on Ubuntu 16.04 (which worked with the original code) and on a recent build of Gentoo (where I saw the failure that this fix attempts to solve).

Sorry about causing the Github CI build failures - the fix that I pushed to my repo just now should fix it, I think.
As before I've tested it on the Linux distros that I have (where the fix works as expected) and I'll monitor the CI
for any new build failures on the Windows build.

Thanks for the code  review.
Charles

romforth:
> @pkoning2 said: It doesn't look like the Github test scripts build this subdirectory. I wonder about an include of "unistd" on Windoze.
> 
> FYI, there has been no effort to avoid excess warning messages when building things in this directory. In fact, I've never known anyone who actually tried to run any of the makefiles in this directory. On windows, the builds have many warnings, but all but checkdisk.mak produce binaries. Looking closely at that build failure and the details in the checkdisk.mak seem to refer to include files from way before Visual Studio 2008 which is the oldest I've got. Removing the irrelevant references in checkdisk.mak and a binary is built. All the windows binaries run and spit out a usage message.
> 
> The simplest change to the asm1130.c source would be to change:
> 
> ```
> #if defined(VMS)
> to 
> #if !defined(_WIN32)
> ```

Rather than use this change (which would affect anything other than Windows), I've chosen to do the more conservative
`#if defined(__linux) || defined(VMS)`
which is applicable only on Linux and VMS so nothing else should be affected.

pmetzger:
@romforth I've fixed this in my ZIMH hard fork of the code. The fix is obviously correct and very simple; unistd.h is on every modern POSIX platform. See: https://github.com/pmetzger/zimh/commit/52be0eb5ceec41aaa358f770d5ae9b5cc23be5c2

LegalizeAdulthood:
Just be aware that `<unistd.h>` doesn't exist on Windows, so it must be guarded for that.

pmetzger:
Yes, if you clicked on my diff, you would see the guard.

LegalizeAdulthood:
I figured that `:)` just adding that to the discussion for the casual reader.

pkoning2:
Checking for "VMS or Linux" is not right.  "unistd.h" as the place where "unlink" is defined isn't specific to Linux but rather POSIX standard.   For example, Mac OS is not Linux but it also mentions unistd.h as required.

The earlier recommendation to make the guard be "not Windows" is the right one.  Can you make that change?

PR #524, opened 2026-01-16
AltairZ80: Corrects bug in ICOM floppy device

deltecent:
A hobbyist found an original FDOS for 8" drives (FD360, FD3712) that has turned up a bug in both the FDC+ and AltairZ80 implementation of the FD3712.

In the FDC+, the data pointer into the 128 byte write "shift register" is reset upon receipt of a CLEAR command. The real hardware never resets the shift register, so software should not adjust the pointer either. AltairZ80 may need the same adjustment.

### DIFF
--- a/AltairZ80/s100_icom.c
+++ b/AltairZ80/s100_icom.c
@@ -1434,8 +1434,6 @@ static uint8 ICOM_Command(UNIT *uptr, ICOM_REG *pICOM, int32 Data)
             pICOM->status &= ~ICOM_STAT_DRVFAIL;
             pICOM->status &= ~ICOM_STAT_CRC;
             pICOM->status &= ~ICOM_STAT_DDM;
-            pICOM->rDataBuf = 0;
-            pICOM->wDataBuf = 0;
             break;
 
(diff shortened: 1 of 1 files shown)


pmetzger:
Howdy! I'm committing this to [ZIMH](https://github.com/pmetzger/zimh), but in the process of doing that, I've noticed there are no Altair integration tests in the repository at all, no attempts to boot and run anything as a smoke test even. Do you have suggestions for an integration test? Like booting an operating system disk I can commit or some such? I don't really know the old S100 machines well.

pmetzger:
I've committed this on my fork at https://github.com/pmetzger/zimh/commit/5a16b98cb4 and as I said, I would appreciate help setting up an AltairZ80 integration test for the repository.

deltecent:
@pmetzger There are test programs for AltairZ80 located here: https://github.com/hharte/altairz80-tests

Packages are here: https://github.com/hharte/altairz80-packages

pmetzger:
@deltecent Thank you so much! I'm importing those into ZIMH now!

PR #527, opened 2026-02-11
Update makefile for musl libc

MrSVCD:
This is a small change to the makefile script to get simh to build on alpine linux which uses musl libc.

By defining `-D_LARGEFILE64_SOURCE` when os type is linux it sets a define macro in musl that defines fopen64 etc as a alias for fopen etc.

This works since fopen etc is already 64-bit in musl, there is no 32-bit fopen etc.


pkoning2:
This changes all Linux builds to work around a problem in just the Alpine version.  Why is that correct?

MrSVCD:
It does not fix x86_64 alpine, only x86 alpine.

pkoning2:
You might take this up with the Alpine folks, to have them match the standards of other Linuxes.

PR #534, opened 2026-04-10
Rename variable to avoid conflict with Winsock type

LegalizeAdulthood:


markpizz:
The change you propose would seemingly only be relevant for Windows builds since you mention "Winsock".

The legacy windows builds currently build cleanly without any change to any source modules.

Maybe the real problem is in the build toolchain and build instructions that you're using.

LegalizeAdulthood:
No, you're wrong.

C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\um\WinSock2.h

line 907 of WinSock2.h
```
/*
 * WinSock 2 extension -- data type and manifest constants for socket groups
 */
typedef unsigned int             GROUP;
```

markpizz:
I never said that under some combination of tool chain build activities you might not encounter the problem you are reporting.  

However, like I said, if you (or I) build with the legacy build facilities (using Visual Studio 2022 or Visual Studio 2008), absolutely no build complaints arise.

If it is failing for you, then you are showing a proof of a problem with the tool chain build process.

LegalizeAdulthood:
The build error is manifest on the windows CI builds as you can see from my fork:

https://github.com/LegalizeAdulthood/simh/actions/runs/24319064062/job/71001783892

> If it is failing for you, then you are showing a proof of a problem with the tool chain build process.

I don't know how many times I have to say it.  You're wrong.

markpizz:
Oops.  I wasn't trying to build the linc simulator which had the conflicting definition.

The simh/simh repo doesn't have this problem (and thus need any change to the linc source code) since other SCP changes implicitly removed that potential conflict some 3 years ago.

The best solution for the open-simh/simh repo is the proposed PR.

LegalizeAdulthood:
Please just go back to playing in your own self-imposed licensed sandbox and stop replying to my bug reports and PRs and wasting my time on the fork that was created explicitly to get away from you and your decisions.

rcornwell:
This looks like more of a Winsock bug. This file probably shouldn't be included where it is to define this variable. GROUP is a name they should not have used.

LegalizeAdulthood:
You can argue that it's poor naming on the part of the winsock header and I might agree with you, but it is what it is.

We can't change winsock headers.

You might as well just argue that it's bad form for POSIX to declare a function called open, preventing us from having a function called open.

pmetzger:
This is a real problem, as I established in the course of fixing the ZIMH build for Windows. My fix for this is here: https://github.com/pmetzger/zimh/commit/6224745f

larsbrinkhoff:
As the original author of this file, I'm happy for this fix to be merged.

PR #537, opened 2026-04-12
CMAKE: Remove CPACK_BUILD_CONFIG from package name

LegalizeAdulthood:
This isn't a predefined CMake or CPack variabe and isn't set anywhere in the CMake scripts.

### DIFF
--- a/cmake/cpack-setup.cmake
+++ b/cmake/cpack-setup.cmake
@@ -39,7 +39,11 @@ else ()
             list(APPEND buildSuffix "win32")
         endif ()
 
-        list(APPEND buildSuffix "\${CPACK_BUILD_CONFIG}")
+        # Append -C argument to CPack, if not empty
+        if (CPACK_BUILD_CONFIG)
+            list(APPEND buildSuffix "${CPACK_BUILD_CONFIG}")
+        endif ()
+
         ## If using Visual Studio, append the compiler and toolkit:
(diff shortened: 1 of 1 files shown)


pmetzger:
Committed on my ZIMH hard fork in https://github.com/pmetzger/zimh/commit/fd0fc004

LegalizeAdulthood:
Apparently this is undocumented as [mentioned here](https://cmake.org/Bug/view.php?id=10940), but if you specify `-C` to inform CPack of a configuration, then this variable will be set to that.  So theoretically, there could be a Debug package and a Release package.

So perhaps a "better" change would be to only append the value of this variable to the list if it is non-empty.

When I was building packages, because I didn't run CPack manually with `-C`, the variable was empty and I got a package name with an extra `-` in it.

LegalizeAdulthood:
I'll update my PR to append the variable only if non-empty.

pmetzger:
I dunno. I think it's not a great idea to rely on undocumented CPack behavior? Makes me pretty uncomfortable.

LegalizeAdulthood:
Since I didn't contribute the entirety of the CMake build scripts, I'm trying to make changes consistent with the original intent.

LegalizeAdulthood:
@pkoning2 What's holding up the merge of this?

LegalizeAdulthood:
Thanks

PR #538, opened 2026-04-13
Handle macOS target triplet

LegalizeAdulthood:
...and when the triplet couldn't be determined, include some diagnostic info to help you figure out why it failed.


cmake/vcpkg-setup.cmake
```diff
@@ -42,7 +42,16 @@ if (NOT DEFINED VCPKG_TARGET_TRIPLET)
             endif ()
 
             set (SIMH_VCPKG_PLATFORM "linux")
+        elseif (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+            set(SIMH_VCPKG_PLATFORM "osx")
```
pmetzger, line 46: Shouldn't this now be macos or some such?

pmetzger:
Honest question (not being negative): what does adding this improve in the current build process?

PR #540, opened 2026-04-16
Correct count in sim_fwrite

### LINKED ISSUE #539: sim_fio.c: sim_fwrite() advances the source pointer incorrectly across multiple flip buffers
This was written by Codex but so far as I can tell the bug report is genuine. I've been rototilling the entire source in a hard fork Codex-generated automated tests found this.

## Summary

`sim_fwrite()` in `sim_fio.c` has a bug in the
big-endian/swapped write path. When a write spans more than one internal
flip buffer, the function advances its source pointer by
`size * count` on every chunk instead of by `size * c`, where `c` is the
number of elements actually written in that chunk.

This causes later chunks to read from the wrong part of the caller's
buffer, which corrupts the file contents.

## Where the bug is

In `sim_fio.c`, inside `sim_fwrite()`:

```c
for (i = (int32)nbuf; i > 0; i--) {                     /* loop on buffers */
    c = (i == 1)? lcnt: nelem;
    sim_buf_copy_swapped (sim_flip, sptr, size, c);
    sptr = sptr + size * count;
    c = fwrite (sim_flip, size, c, fptr);
```

The source pointer update should use `c`, not `count`.

## Why this is wrong

`sim_fwrite()` may break the caller's write into multiple pieces:

- `nelem` is the number of elements that fit in the flip buffer
- `nbuf` is the number of chunks to write
- `c` is the number of elements in the current chunk

After writing one chunk, the source pointer must advance by the size of
that chunk:

```c
sptr = sptr + size * c;
```

But the current code advances by the size of the entire original request
every time:

```c
sptr = sptr + size * count;
```

That is only correct if there is exactly one chunk.

## Minimal patch

```diff
diff --git a/src/runtime/sim_fio.c b/src/runtime/sim_fio.c
@@
     c = (i == 1)? lcnt: nelem;
     sim_buf_copy_swapped (sim_flip, sptr, size, c);
-    sptr = sptr + size * count;
+    sptr = sptr + size * c;
     c = fwrite (sim_flip, size, c, fptr);
```

## How this was found

This was exposed while creating host-side unit coverage for
`sim_fio.c`.

A new test forces `sim_fwrite()` to use more than one flip buffer by
writing more than:

```c
FLIP_SIZE / sizeof(uint16_t)
```

elements with `sim_end = FALSE`, then reads the data back with
`sim_fread()` and checks that the logical values round-trip unchanged.

Before the patch, the test failed once the second chunk was reached.

## Reproducer shape

In outline:

1. Force the swapped path:
   - `sim_end = FALSE`
2. Write more than one flip buffer of `uint16_t` values
3. Read the file back with `sim_fread()`
4. Compare the result against the original buffer

Before the fix, values in the later portion of the file came from the
wrong source region.

## Impact

This affects endian-swapped writes that are large enough to require more
than one internal flip-buffer chunk.

Small writes do not expose the problem, which is why it can remain hidden
unless specifically tested.

LegalizeAdulthood:
Fixes #539

### DIFF
--- a/sim_fio.c
+++ b/sim_fio.c
@@ -271,7 +271,7 @@ sptr = (const unsigned char *) bptr;                    /* init input ptr */
 for (i = (int32)nbuf; i > 0; i--) {                     /* loop on buffers */
     c = (i == 1)? lcnt: nelem;
     sim_buf_copy_swapped (sim_flip, sptr, size, c);
-    sptr = sptr + size * count;
+    sptr = sptr + size * c;
     c = fwrite (sim_flip, size, c, fptr);
     if (c == 0) {
         free(sim_flip);


markpizz:
@LegalizeAdulthood , Since you've never actually had commits merged to the simh repo, you might not have realized that there is a convention in the simh repo which was established based on the great suggestion of Timothe Litt (@tlhackque) some 13 years ago.  Specifically, the commit comments should be of the form:

AFFECTED-FACILITY: One line description max 80 characters
\<blank-line-if-additional-comments\>
Optional additional comments as needed for as many lines as would be useful < 80 characters 

AFFECTED-FACILITY would be things like FIO (in this case), SCP, ETHER, TMXR etc., or (one or more simulator names).

Also, the additional comments are best when they include the description and justification for the change since that info gets carried anywhere the repo goes, while pointers to PR comments or Issue comments are stuck on github.

So, the commit for this PR should look at least like:

   FIO: Fix Big Endian sim_fwrite() error

   maybe with a summary of what was discussed in the Issue message.

Additionally, although you are the one committing the change in your repo, if I were doing this, I would specifically set Perry Metzger \<perry@piermont.com\> as the Author.

LegalizeAdulthood:
Go away

PR #547, opened 2026-04-29
PDP18b: Add Type 34 display.

larsbrinkhoff:
Tested with the DECUS 7-40 DUAL game, in both Type 34 and Type 340 mode.

### DIFF
--- a/PDP18B/pdp18b_dpy.c
+++ b/PDP18B/pdp18b_dpy.c
@@ -1,6 +1,6 @@
-/* pdp18b_dpy.c: PDP-7 Type 340 interface
+/* pdp18b_dpy.c: PDP-7 Type 34 and Type 340 interface
 
-   Copyright (c) 2019, Lars Brinkhoff
+   Copyright (c) 2019, 2026, Lars Brinkhoff
 
    Permission is hereby granted, free of charge, to any person obtaining a
    copy of this software and associated documentation files (the "Software"),
@@ -38,6 +38,11 @@
 #define DBG_IOT         001         /* IOT instructions. */
 #define DBG_IRQ         002         /* Interrupts. */
 #define DBG_INS         004         /* 340 instructions. */
+#define DBG_PLT         010         /* Points plotted. */
+
+#define UNIT_V_TYPE34   (UNIT_V_UF + 0)                 /* Type 34 */
+#define UNIT_TYPE34     (1 << UNIT_V_TYPE34)
+#define TYPE34          (dpy_unit[0].flags & UNIT_TYPE34)
 
 /*
  * Number of microseconds between svc calls.  Used to age display and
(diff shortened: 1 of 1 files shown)


LegalizeAdulthood:
By the way, the VS 2022 jobs that are failing are due to:

> The build failed due to a naming conflict with the symbol GROUP in linc/linc_tape.c.

This is fixed by PR #534

pmetzger:
Added to ZIMH in: https://github.com/pmetzger/zimh/commit/90eb517e71bf0555beaffd01e9a35b8ba93c0b5b

Normally I try to add tests to everything I touch, but in this case the way to do that wasn't entirely obvious. It would be nice to have a precise automated test procedure for this; ideas are solicited.

larsbrinkhoff:
@pmetzger, regaring testing.  I imagine there could be a small test that draws a few points.  The result could be verified using the output from `SET DPY DEBUG=PLT`

larsbrinkhoff:
@Frodevan used this Type 34 device to successfully run the PDP-7 game Napoleon: https://forum.vcfed.org/index.php?threads/university-of-oslo-pdp7-paper-tapes-vs.1257172/page-2#post-1500671

Frodevan:
We do have the PDP-7 Display type 34 maintenance test-tape in the Oslo dumps as well, which I started disassembling yesterday to figure out what the different tests does. The first one at address 101 at least sets X or Y based on the switches (msb selects the axis to use, the remaining bits selects value, with the other axis at 0). It's too close to the edge to actually see it very well visually in the emulated display, but I would guess this could easily be used for an automated test and the mentioned debug flag.

pmetzger:
> @pmetzger, regaring testing. I imagine there could be a small test that draws a few points. The result could be verified using the output from `SET DPY DEBUG=PLT`

So for ZIMH, I need automated tests. What would an automated test do here? Set up a virtual screen and draw points expected to be in particular places? Are there other interesting behaviors that the thing has? Also, is there something I can add to the documentation about the device? (The ZIMH docs are in Markdown, not Word.)

pmetzger:
> We do have the PDP-7 Display type 34 maintenance test-tape in the Oslo dumps as well, which I started disassembling yesterday to figure out what the different tests does. The first one at address 101 at least sets X or Y based on the switches (msb selects the axis to use, the remaining bits selects value, with the other axis at 0). It's too close to the edge to actually see it very well visually in the emulated display, but I would guess this could easily be used for an automated test and the mentioned debug flag.

That would be exceptionally cool! I'd be happy to work with you guys on getting this (and documentation) in.

Frodevan:
> > We do have the PDP-7 Display type 34 maintenance test-tape in the Oslo dumps as well [...]
> 
> That would be exceptionally cool! I'd be happy to work with you guys on getting this (and documentation) in.

Everything is available here
https://archive.org/details/oslo-pdp-7

The loader to use can be found under the DIGITAL 7-13-I folder, the mentioned test tape is in DIGITAL 7-60-M. In accordance with the PDP-7 user-handbook, I tend to load the loader tape with LOAD -R [file] 17763, and RUN 17770 after attaching the final tape to PTR. Technically, loading the RIM-loader this way might be a bit of an overkill, but it ensures everything remains in memory and I like to do it this way just in case a tape to load has a patched FF-loader or something like that.

In the case of this particular test-tape, the header instructs the loader to halt when complete, for the user to manually run from a particular test-entry address. I unfortunately don't have the original manuals, and the one for this tape is not on Bitsavers, which is why I am disassembling the binary.

pmetzger:
I'm not going to get to this for a few days but I have put it in my queue. I'm doubtless going to have a lot of questions once I do get to it.

PR #549, opened 2026-05-03
AltairZ80: Improved choice of boot ROM

psco:
An appropriate boot ROM is chosen based on the attached disk image.

Replaces the PR with the same name and the same functionality but hopefully without the commit clutter.


pmetzger:
Also in [ZIMH as this](https://github.com/pmetzger/zimh/commit/ef57021b).

PR #554, opened 2026-05-19
Fix self lockout on VDE socket

Janrupf:
SIMH seems to somewhat quietly lock itself out of its own VDE socket because it sets the permission bits to all zero. I found some resources saying SIMH needs to run as root to use VDE, which seems to also stem from this problem. Now it works without running as root

### DIFF
--- a/sim_ether.c
+++ b/sim_ether.c
@@ -2358,6 +2358,8 @@ else { /* !tap: */
     const char *devname = savname + 4;
 
     memset(&voa, 0, sizeof(voa));
+    voa.mode = 0600;
+
     if (!strcmp(savname, "vde:vdedevice"))
       return sim_messagef (SCPE_OPENERR, "Eth: Must specify actual vde device name (i.e. vde:/tmp/switch)\n");
     while (isspace(*devname))


pmetzger:
Weird because this effect seems to contradict the VDE documentation but testing shows the patch is indeed correct. I applied it to [ZIMH in this commit](https://github.com/pmetzger/zimh/commit/238bdeae5395683f610c6e615ea8e98e179a38a6); I have no control over the SIMH repository. I used mode 0600 as it appears there was no need to give the socket an execute bit.

Janrupf:
Ah whoops, you are right on the permission bits. I think I just had the `7` there for testing and then forgot to change it. Fixed that now

Janrupf:
Regarding the VDE documentation, @pmetzger where would you say it contradicts it? I think its worded a bit weirdly with "[...] has been left for backwards compatibility", but it only explicitly mentions that the open args can be set to `NULL` (in which case VDE would apply mode `0700`, interestingly enough also including the executable bit...), but not that they can be zeroed or are ignored.

I'm considering making a PR at VDE to maybe ignore the mode if its set to zero since I don't see how that would ever be useful, but arguably VDE shouldn't work around bugs in applications.

pmetzger:
The documentation seemed to indicate to me that zero meant defaults would be applied.

Anyway, it's rare that I've seen such a simple fix completely correct a problem. Thank you for bringing it to everyone's attention. I hope SIMH also applies it.

PR #583, opened 2026-09-22
LINC: various updates

larsbrinkhoff:
- Fix for MUL.
- Game controllers for Spacewar.
- Fixes for disassembler.
- Fix for loading octal.
- Escape maps to CASE.
- Map up, down, page up, page down to LINC movement.
- Fix recording of P for CPU instruction history.
- Fix line endings to CRLF.

CC @dgesswein


larsbrinkhoff:
The check "cmake-builds / LEGACY XP executables" fails, but it does on the master branch too.

markpizz:
While you doing this series of updates, please make sure the line endings are CRLF for all source files to align with the historic simh standard.

larsbrinkhoff:
Good point; I keep forgetting that.  I have fixed my line endings.  Maybe an automated test would be in order.

tlhackque:
> Good point; I keep forgetting that. I have fixed my line endings. Maybe an automated test would be in order.

This keeps coming up.  Git can handle this issue transparently.

Short form: set `core.autocrlf` to `true` on windows, and `input` for others.  `core.eol` to your working directory preference.  Or put `*.c text eol=crlf` (and `*.h`) in `.gitattributes` if you insist.  (To force unix line endings -- e.g. for shell scripts -- `eol=lf`; for binary files, use the `binary` macro.)

See:
GitHub's doc: https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings (TLDR)
Git's doc: https://git-scm.com/docs/git-config, https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration, https://git-scm.com/docs/gitattributes/2.6.7, and https://git-scm.com/docs/gitfaq/2.46.0#Documentation/gitfaq.txt-WhatstherecommendedwaytostorefilesinGit

The last reference summarizes git's handling and recommendations.

markpizz:
> This keeps coming up. Git can handle this issue transparently.

This has been discussed several times.  

Git can do what you suggest when you start from the beginning.

However, to get there now you have to be willing to have a massive commit which changes ALL of the existing source files in the entire repo to the internal standard.  

The project has repeatedly decided against this.

